PR #132, opened 2013-10-31
Fix Python 3 compatibility on Windows

SAPikachu:
In Python 3, `pathname2url` accepts `str` (which is aliased as `unicode` in `compat.py`), so `path` should not be encoded.

### DIFF
--- a/weasyprint/urls.py
+++ b/weasyprint/urls.py
@@ -77,7 +77,7 @@ def path2url(path):
         # Make sure directory names have a trailing slash.
         # Otherwise relative URIs are resolved from the parent directory.
         path += os.path.sep
-    if isinstance(path, unicode):
+    if isinstance(path, unicode) and sys.version_info[0] < 3:
         path = path.encode(FILESYSTEM_ENCODING)
     path = pathname2url(path)
     if path.startswith('///'):


rhvonlehe:
I'd like to see this compatibility cleaned up as well in a future release.  See stackoverflow issues [1] and [2]:
[1] http://stackoverflow.com/questions/37994733/weasyprint-usage-with-python-3-x-on-windows/38082024#38082024
[2] http://stackoverflow.com/questions/38081511/need-help-understanding-bytes-like-object-vs-str-in-python-3-5

liZe:
I'm trying to make WP work on Windows with Python 2.x and 3.x (ie. fixing #331 #330 #132 #329).

As said by @rhvonlehe, the Windows version of `pathname2url` requires unicode and not bytes (it's not really documented and works on Linux, well…). But I'd really prefer to keep returning bytes in `path2url`, as it's probably safer.

So, instead of testing the Python version, I'd prefer to call `path = pathname2url(path)` before `path = path.encode(FILESYSTEM_ENCODING)`. @SimonSapin what do you think of that?

SimonSapin:
So… this probably requires more time to research than I have right now. But I vaguely remember thinking that we’d be better of implementing the `file:` URL scheme ourselves rather than leave it to `urllib`. Stay in Unicode all the way, and give a Unicode string to `open()`.

liZe:
> Stay in Unicode all the way, and give a Unicode string to open().

@SimonSapin It's better for Python 3, but we probably need to keep bytes for Python 2 (I suppose that the `path.encode` had been added for a good reason). So: we merge the PR as is?

> I'm trying to make WP work on Windows with Python 2.x and 3.x

@rhvonlehe @ChameleonRed I've installed WeasyPrint on Windows 10 64 bit following these steps:
- Install [Python 2.7.x 64 bit](https://www.python.org/downloads/windows/) with `python` in `PATH`.
- Install [GTK-3 64 bit](https://github.com/tschoonj/GTK-for-Windows-Runtime-Environment-Installer).
- Reboot (true story).
- Install [Visual C++ compiler for Python 2.7](http://aka.ms/vcpython27).
- Install WeasyPrint with `python -m pip install weasyprint`.
- Test with `python -m weasyprint http://weasyprint.org weasyprint.pdf`.

I've met the problem in #330 during my different tests. What seems to be important is:
- To compile some packages (which ones?) on your version of Windows. That's what `pip` does, and that's why it relies on Visual C++.
- To use 64 bit packages everywhere.

Following these two rules on other versions of Windows (including 32 bit with 32 bit packages) should be OK. With this PR, it should even work on Windows with Python 3.5.

If it works for @rhvonlehe and @ChameleonRed, I can change the documentation.

SimonSapin:
> (I suppose that the path.encode had been added for a good reason)

Haha, that’s a strong supposition. Python 2 supports Unicode in `open()` fine, doesn’t it?

SimonSapin:
To clarify, my suggestion to stay in Unicode all the way is if we change `default_url_fetcher` to have a special case for `file:` URLs (like there’s one for `data:`) where we handle them ourselves instead of giving them to `urlopen()`: split to extract the path component, reject if host is not empty or `localhost`, percent-decode, then call `open()`.

ghedin:
Hello folks,

First of all I would like to congratulate @SimonSapin and all the project's collaborators. This is a great piece of software :+1: 

I also was unable to use the library with Python3 on Windows 7. After applying the pull request suggested by @SAPikachu it just works. I have tested it with a few simple HTML files and could successfully convert them to PDF.

I'm not with much spare time at the moment, but I'd like to contribute by applying @SimonSapin suggestion (stay Unicode all the way). Since I did not look at the code yet, I can't imagine how big that change would be, but I will take a look at it asap.

rhvonlehe:
I understand you'd like me to install using 64-bit versions of everything
to see if that helps with Python 3.5 compatibility on Windows.  I can do
that, but it might be this weekend before I get to it.

liZe:
> I understand you'd like me to install using 64-bit versions of everything
> to see if that helps with Python 3.5 compatibility on Windows.

:) not exactly! Here's what I propose:
1. you can try follow the installation guide I provided to fix #329 and #330,
2. I write the new version of this PR to fix #132 and #331.

With these two points solved, we can consider the Windows support as quite good!

rhvonlehe:
@liZe Ah, I see.  I think I can safely say I don't experience issue #329, using either 32-bit Python 2.7 or 32-bit Python 3.5, and 32-bit GTK (on 64-bit Windows 10).  I think your installation guide clarifications will definitely help, though.  I had the best luck following instructions here [1], rather than the current weasyprint instructions.  I am willing to follow your instructions explicitly on my home machine (current no weasy installed), though, if that helps.  Looking forward to your new PR for #132, thanks!

[1] https://gist.github.com/doobeh/3188318

dom96:
Don't have much to contribute to the discussion except that I ran into the same problem and applying this PR fixed it for me. (Windows 7 x86_64, Python 3.4 (Installed via Anaconda), let me know if you need more info).

Hope this helps!

liZe:
@SAPikachu @rhvonlehe @dom96 I've written and tested #345. Could you please try that it works for you?

liZe:
Fixed now that #345 is fixed. Thanks for your help!

TheEasyProgrammer:
I am getting cool to work in ubuntu box but I am struggling in this issue in windows box. Please help me on this I have installed gtk and have path in environment variables. But no luck. Please help me on this.

libcairo.so.2 os error 0x7e in python

PR #149, opened 2013-12-27
box-shadow and text-shadow support

### LINKED ISSUE #13: CSS box-shadow and text-shadow
WeasyPrint currently does not support the CSS [`box-shadow`](http://www.w3.org/TR/css3-background/#the-box-shadow) or [`text-shadow`](http://www.w3.org/TR/css3-text/#text-shadow) properties.

The main limitation is that these shadow use a Gaussian blur, but cairo (by itself) does not do such blurs. Since the blur can not be expressed as vectors in a PDF file, they need to be rendered to pixel-based images. It would be best to make images at the [surface’s fallback resolution](http://www.cairographics.org/documentation/pycairo/2/reference/surfaces.html#cairo.Surface.get_fallback_resolution). Scipy can do the [gaussian filter](http://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.filters.gaussian_filter.html) on a numpy array based on the [image’s data](http://www.cairographics.org/documentation/pycairo/2/reference/surfaces.html#cairo.ImageSurface.get_data).

liZe:
Fix #13.

### DIFF
--- a/weasyprint/css/computed_values.py
+++ b/weasyprint/css/computed_values.py
@@ -360,6 +360,17 @@ def box_shadow(computer, name, values):
     return shadows
 
 
+@register_computer('text-shadow')
+def text_shadow(computer, name, values):
+    """Compute the ``text-shadow`` property."""
+    shadows = []
+    for value in values:
+        shadows.append([
+            length(computer, name, i, pixels_only=True)
+            for i in value[:3]] + [value[3]])
+    return shadows
+
+
 @register_computer('content')
 def content(computer, name, values):
     """Compute the ``content`` property."""
--- a/weasyprint/css/computed_values.py
+++ b/weasyprint/css/computed_values.py
@@ -354,9 +354,9 @@ def box_shadow(computer, name, values):
     """Compute the ``box-shadow`` property."""
     shadows = []
     for value in values:
-        print(value)
-        shadows.append(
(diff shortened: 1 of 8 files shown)


liZe:
I close this PR as it won't be merged. Depending on the PIL and using raster blurs are an awful solution.

Afoucaul:
Hi, is there any plan to support `box-shadow` one day? Is there any work in progress other than this PR?

liZe:
> Hi, is there any plan to support `box-shadow` one day? Is there any work in progress other than this PR? I'd be glad to help, but I'm a bit lost in the code, and I have troubles getting the test to run from this PR. If you updated your branch, that would help me start working on this feature.

This pull request generates a raster image included in the generated PDF, that's an ugly solution. Unless there's a clean way to include blurs in PDF files, and unless this feature is included in Cairo, I don't see how we could include blurs in a clean way.

hlandau:
Worth noting that box shadow can be used without blurs. So I guess it would be possible to support a subset of box-shadow where blur isn't used.

liZe:
> Worth noting that box shadow can be used without blurs. So I guess it would be possible to support a subset of box-shadow where blur isn't used.

That’s right, and that would be useful in some cases. We can continue to discuss on #13 and see if some people are interested in providing and/or using this feature.

PR #150, opened 2013-12-31
Add support for empty-cell and visibility: collapse

liZe:
And remove column, row and group borders with border-collapse = 'separate'.


liZe:
The `cell-visibility: collapse` is not implemented correctly for tables here. For other features, it's implemented in #291.

PR #197, opened 2014-05-23
Add missing hyphen character at line break activated by a soft hyphen.

jdus:
This code adds the missing hyphen for line breaks activated by soft hyphens. It will also check whether the hyphen character still fits or not. I send this request as a basis for discussion. The code definitely can be improved and needs testing.


liZe:
It's been reviewed and merged, thank you!

PR #239, opened 2015-02-16
Add more apt get packages for Ubuntu 14.04

simkimsia:
Had the error of   ** make sure the development packages of libxml2 and libxslt are installed **
when attempt to install WeasyPrint on Ubuntu 14.04 via virtualenv

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -107,6 +107,14 @@ If your favorite system is not listed here but you know the package names,
 Debian / Ubuntu
 ~~~~~~~~~~~~~~~
 
+Ubuntu 14.04 Trusty:
+
+.. code-block:: sh
+
+    sudo apt-get install libxml2-dev libxslt-dev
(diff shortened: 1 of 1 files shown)


SimonSapin:
Aren’t they dependencies of `python-lxml`?

mbertheau:
@SimonSapin not the `-dev` packages: http://packages.ubuntu.com/trusty/python-lxml

SimonSapin:
Ok. What I think is going on: if you don’t use a virtualenv or a virtualenv with `--system-site-packages`, the current python-lxml dependency is enough. If you do use virtualenv with `--no-site-packages` (now the default), pip will try to build another copy lxml and then the -dev dependencies will be needed.

However, none of this is specific to Ubuntu 14.04, I believe the same happens on Debian and older Ubuntu. Therefore, please add these dependencies in the existing Debian / Ubuntu section rather than adding a new section. (Also, I’d prefer keeping it to one `apt-get` line, even if it’s long.)

simkimsia:
I have created another PR. Will close this one.

PR #243, opened 2015-03-02
Ensure libxml2-dev libxslt-dev are added

simkimsia:
for Ubuntu 14.04 or newer


SimonSapin:
Please see my comment on #239:

> However, none of this is specific to Ubuntu 14.04, I believe the same happens on Debian and older Ubuntu. Therefore, please add these dependencies in the existing Debian / Ubuntu section rather than adding a new section.

Also, can modify an existing PR without creating a new one by pushing more commits to the same branch, or, if you prefer, using `git commit --amend` to modify an existing commit and then `git push --force`.

Issue #174, opened 2014-03-15
Basic support for Named strings


zxcat:
It's quite simple to add. And very useful, e.g.:
```
h1 { string-set: title content() }
@page { @top-right { content: string(title) }}
```
Details: http://www.w3.org/TR/css3-gcpm/#running-headers-and-footers

Basic `content()` and `string(name)` support without any other parameters and forms will be great. _(Looks like other variants like `env()` are not mature enough to implement.)_

I think implementation is quite easy:
1. `string-set:` is much like `counter-reset:name value`. The only difference: `value` there is more complex, like `content:`;
2. `string(name)` is exactly like `counter(name)`, but with different dictionary.

I've tried to implement this myself, but I don't understand Python source.

Any chances *Named strings* will be in the next version of WeasyPrint? : )

SirNeuman:
Is there any way around this for the time being? I, for example, need a string from my html page (a unique ID number) to be displayed in @bottom-left{}. So I wanted to do in the css:
 #ID-Number {string-set: the-number content();} 
@page{@bottom-left{content: string(the-number)}}
but the area at the bottom left of the pdf just shows up blank. I didn't realize I couldn't do this until I found this post, but is there some way of getting that ID-Number in the bottom left of the pdf some other way? If i try and put it in the html it appears above where i want it to appear, which is not in line with the rest of my footer (I have other stuff in bottom-center and bottom-right)

SimonSapin:
@SirNeuman Your `#ID-Number` selector seems to imply that you have only one such element. If you want some content from the document in page footers *that is the same on all pages*, you can use `position: fixed` (which repeats on all pages) and some margin or positioning trick (like `top: 100%` maybe) to move it into the footer.

However if that number varies within a document (e.g. a chapter number), then the only way is named strings or some other feature very much like it.

SirNeuman:
ahh ok. Thanks. Knowing that fixed position puts it on every page helps.

E-M-P-I-R-E:
I am interested in working on this. If you are able to provide some guidance, @SimonSapin, that would be fantastic.

SimonSapin:
@E-M-P-I-R-E Good!

Unfortunately, the GCPM spec suffers from being largely under-defined and having received very little review. For example, it’s not clear to me whether the `string()` function of the `content` property is valid in contexts other than on page-margin boxes (e.g. `::before` or `::after` pseudo-elements) and what the second argument means in this case.

[...]

PR #246, opened 2015-03-20
Support for named strings in margin-boxes

E-M-P-I-R-E:
Named strings using the `-weasy-string-set` style and the CSS `string` function passed to `content()`.
http://dev.w3.org/csswg/css-gcpm/#named-strings
named_strings

More discussion in feature request - issue https://github.com/Kozea/WeasyPrint/issues/174

Below is a simple example (more in tests/test_boxes.py).

``` html
<html>
<head>
<style>
body{
     counter-reset: header_counter; 
}
@page:first{
    @top-center { content: string(header)}
}
@page{
    @top-center { content: string(header, first-except)}
}
.page{
      page-break-before: always;
}

h1{
    counter-increment: header_counter;
	string-set: header content(after);
	-weasy-string-set: header content(text);
}
h1:before{
	content: "Chapter " counter(header_counter) " ";
}
</style>
</head>
<body>
    <div class="page"><h1>First</h1>This demonstrates named strings using the `string-set` style and the CSS `string` function passed to `content()`.<br>There are two assignments "First" and "Second" plus the psuedo ::before element. The numbers are produced using CSS `counter`.</div>
    <div class="page">CONTENT</div>
    <div class="page" ><h1 >Second</h1>Pages after the first have `first-except` set, meaning the page on which the assignment is made is excluded. Proceeding pages will have the value set here ("Second").</div>
    <div class="page">CONTENT</div>
</body>
</html>
```

### DIFF
--- a/weasyprint/css/properties.py
+++ b/weasyprint/css/properties.py
@@ -97,8 +97,8 @@ INITIAL_VALUES = {
     'quotes': list('“”‘’'),  # depends on user agent
     'position': 'static',
     'right': 'auto',
-    #default not defined in spec http://dev.w3.org/csswg/css-gcpm/#named-strings
-    'string_set': '',
+    # http://dev.w3.org/csswg/css-gcpm/#named-strings
+    'string_set': 'none',
     'table_layout': 'auto',
     # Taken from CSS3 Text.
     # The only other supported value form CSS3 is -weasy-end.
--- a/weasyprint/css/properties.py
+++ b/weasyprint/css/properties.py
@@ -97,6 +97,8 @@ INITIAL_VALUES = {
     'quotes': list('“”‘’'),  # depends on user agent
     'position': 'static',
     'right': 'auto',
+    #default not defined in spec http://dev.w3.org/csswg/css-gcpm/#named-strings
+    'string_set': '',
     'table_layout': 'auto',
     # Taken from CSS3 Text.
     # The only other supported value form CSS3 is -weasy-end.
--- a/weasyprint/css/validation.py
+++ b/weasyprint/css/validation.py
@@ -18,6 +18,7 @@ import math
 
 from tinycss.color3 import parse_color
 from tinycss.parsing import split_on_comma, remove_whitespace
+from tinycss.token_data import FunctionToken
 
 from ..logger import LOGGER
 from ..formatting_structure import counters
@@ -686,6 +687,10 @@ def validate_content_token(base_url, token):
             if style in ('none', 'decimal') or style in counters.STYLES:
                 return (name, args)
         elif prototype in (('string', ['IDENT']),('string', ['IDENT', 'IDENT'])):
+            if len(args) > 1:
+                args[1] = args[1].lower()
+                if args[1] not in ('first', 'start', 'last', 'first-except'):
+                    raise InvalidValues()
             return (name, args)
 
 
@@ -1272,17 +1277,23 @@ def bookmark_level(token):
 @validator(prefixed=True)  # CSS3 GCPM, experimental
 def string_set(tokens):
     """Validation for ``string-set``."""
-    if len(tokens) != 2:
-        raise InvalidValues("Only one value is supported. Received %s" %len(tokens)-1)
-    function = parse_function(tokens[1])
-    if function is None:
-        keyword = 'text'
-    else:
-        name, args = function
-        keyword = args[0].value
-    var_name = get_keyword(tokens[0])
-    if tokens[0].type == "IDENT" and keyword in ('text', 'before', 'after'):
-        return ('keyword', keyword, 'name', var_name)
+    if len(tokens) == 2:
+        var_name = get_keyword(tokens[0])
+        function = parse_function(tokens[1])
+        if function is None:
+            if not isinstance(tokens[1], FunctionToken):
+                raise InvalidValues
+            keyword = 'text'
+            name = tokens[1].function_name
+        else:
+            name, args = function
+            keyword = args[0].value.lower()
+        if name != 'content':
+            raise InvalidValues
+        if tokens[0].type == "IDENT" and keyword in ('none', 'text'):
+            return (var_name, keyword)
+    elif len(tokens) and tokens[0].value == 'none':
+        return 'none'
     raise InvalidValues
 
 
--- a/weasyprint/css/validation.py
+++ b/weasyprint/css/validation.py
@@ -1291,8 +1291,8 @@ def string_set(tokens):
             keyword = args[0].value.lower()
         if name != 'content':
             raise InvalidValues
-        if tokens[0].type == "IDENT"
-                and keyword in ('none', 'text', 'after', 'before'):
+        if (tokens[0].type == "IDENT"
+                and keyword in ('none', 'text', 'after', 'before')):
             return (var_name, keyword)
     elif len(tokens) and tokens[0].value == 'none':
         return 'none'
--- a/weasyprint/css/validation.py
+++ b/weasyprint/css/validation.py
@@ -685,6 +685,8 @@ def validate_content_token(base_url, token):
             style = args[-1]
             if style in ('none', 'decimal') or style in counters.STYLES:
                 return (name, args)
+        elif prototype in (('string', ['IDENT']),('string', ['IDENT', 'IDENT'])):
+            return (name, args)
 
 
 def parse_function(function_token):
@@ -1265,6 +1267,16 @@ def bookmark_level(token):
             return value
     elif get_keyword(token) == 'none':
         return 'none'
+
+
+@validator(prefixed=True)  # CSS3 GCPM, experimental
+def string_set(tokens):
+    """Validation for ``string-set``."""
+    function = parse_function(tokens[1])
+    keyword = function[1][0].value
+    name = get_keyword(tokens[0])
+    if len(tokens) == 2 and tokens[0].type == "IDENT" and keyword in ('text', 'before', 'after'):
+        return ('keyword', keyword, 'name', name)
 
 
 @validator(unprefixed=True)
--- a/weasyprint/css/validation.py
+++ b/weasyprint/css/validation.py
@@ -1291,7 +1291,7 @@ def string_set(tokens):
             keyword = args[0].value.lower()
         if name != 'content':
             raise InvalidValues
-        if tokens[0].type == "IDENT" and keyword in ('none', 'text'):
+        if tokens[0].type == "IDENT" and keyword in ('none', 'text', 'after', 'before'):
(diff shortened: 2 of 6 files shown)


SimonSapin:
This looks great! I’ll comment on specifics inline.

Tests [fail](https://travis-ci.org/Kozea/WeasyPrint/jobs/55126940) on Travis-CI because:

* `flake8` complains about whitespace and overlong lines
* `xrange` doesn’t exist on Python 3.x. Add `from ..compat import xrange` at the top of the file.

weasyprint/css/properties.py
```diff
@@ -97,6 +97,8 @@
     'quotes': list('“”‘’'),  # depends on user agent
     'position': 'static',
     'right': 'auto',
+    #default not defined in spec http://dev.w3.org/csswg/css-gcpm/#named-strings
+    'string_set': '',
```
SimonSapin: The initial value *is* defined in the spec, in the property definition table. It is `none`, which should be represented as a string `'none'` since returning `None` in `validation.py` indicates a parse error.

weasyprint/css/validation.py
```diff
@@ -685,6 +685,8 @@ def validate_content_token(base_url, token):
             style = args[-1]
             if style in ('none', 'decimal') or style in counters.STYLES:
                 return (name, args)
+        elif prototype in (('string', ['IDENT']),('string', ['IDENT', 'IDENT'])):
+            return (name, args)
```
SimonSapin: When there are two arguments, this should check that the second one is one of the supported keywords.

SimonSapin: These keywords should also be case-insensitive. (But not the string/variable’s name!)

weasyprint/css/validation.py
```diff
@@ -1267,6 +1269,23 @@ def bookmark_level(token):
         return 'none'
 
 
+@validator(prefixed=True)  # CSS3 GCPM, experimental
+def string_set(tokens):
```
SimonSapin: I have a hard time following this, sorry. Could you clarify what set of syntax you mean to support? All unsupported syntax should be rejected.

weasyprint/css/validation.py
```diff
@@ -1267,6 +1269,23 @@ def bookmark_level(token):
...
+    """Validation for ``string-set``."""
+    if len(tokens) != 2:
+        raise InvalidValues("Only one value is supported. Received %s" %len(tokens)-1)
+    function = parse_function(tokens[1])
+    if function is None:
+        keyword = 'text'
```
SimonSapin: When the second token is not a function, it could be anything else. This should check what it is and only accept the expected syntax.

weasyprint/css/validation.py
```diff
@@ -1267,6 +1269,23 @@ def bookmark_level(token):
...
+    else:
+        name, args = function
```
SimonSapin: `name` should be checked here.

weasyprint/css/validation.py
```diff
@@ -1267,6 +1269,23 @@ def bookmark_level(token):
...
+        keyword = args[0].value
+    var_name = get_keyword(tokens[0])
+    if tokens[0].type == "IDENT" and keyword in ('text', 'before', 'after'):
```
SimonSapin: Keywords should be case-insensitive. (Use `.lower()`.)

weasyprint/css/validation.py
```diff
@@ -1267,6 +1269,23 @@ def bookmark_level(token):
...
+        return ('keyword', keyword, 'name', var_name)
```
SimonSapin: Do the `'keyword'` and `'name'` strings add information here, or could the return value be just `(keyword, name)`? (Or `(name, keyword)`, to match the order of the syntax?)

weasyprint/formatting_structure/build.py
```diff
@@ -235,6 +236,9 @@ def content_to_boxes(style, parent_box, quote_depth, counter_values,
                 counters.format(counter_value, counter_style)
                 for counter_value in counter_values.get(counter_name, [0])
             ))
+        elif type_ == 'string':
+            text = context.get_string_set_for(*tuple(value))
```
SimonSapin: Avoid crashing here when `context` is `None`. (Maybe just ignore `string()` functions.)

weasyprint/layout/__init__.py
```diff
@@ -71,6 +70,8 @@ def __init__(self, enable_hinting, style_for, get_image_from_uri):
         self.get_image_from_uri = get_image_from_uri
         self._excluded_shapes_lists = []
         self.excluded_shapes = None  # Not initialized yet
+        self.string_set = defaultdict(lambda:defaultdict(lambda:[]))
```
SimonSapin: Nit: replace `lambda: []` with `list`.

weasyprint/layout/__init__.py
```diff
@@ -89,3 +90,33 @@ def finish_block_formatting_context(self, root_box):
...
+    def get_string_set_for(self, name, keyword=None):
+        """ Resolve value of string function (as set by string set)
+
+            We'll have something like this that represents all assignments on a given page
+            {1: [u'First Header'], 3: [u'Second Header'], 4: [u'Third Header', u'3.5th Header']}
+
+            Value depends on current page - http://dev.w3.org/csswg/css-gcpm/#funcdef-string
+
+            :param name: the name of the named string.
+            :param name: indicates which value of the named string should be used.
```
SimonSapin: Should this be `:param keyword:`?

weasyprint/layout/__init__.py
```diff
@@ -89,3 +90,33 @@ def finish_block_formatting_context(self, root_box):
...
+                         Default is the first assignment on the current page else the most recent assignment (entry value)
+            :returns: text
+
+        """
+        last = keyword == "last"
+        if self.current_page in self.string_set[name]:
```
SimonSapin: Does this crash if `name` is not in `self.string_set`? Maybe replace all occurrences of `self.string_set[name]` with a single `assignments = self.string_set.get(name)` and check for `assignments is None`.

weasyprint/formatting_structure/build.py
SimonSapin: `*tuple(value)` can be just `*value`

weasyprint/layout/__init__.py
```diff
@@ -89,3 +90,33 @@ def finish_block_formatting_context(self, root_box):
...
+            elif last:
+                # most recent assignment
+                return self.string_set[name][self.current_page][-1]
+            return self.string_set[name][self.current_page][0]
+        else:
+            # no assignment on this page - search backwards; through previous page
+            for previous_page in xrange(self.current_page, 0, -1):
+                if previous_page in self.string_set[name]:
+                    return self.string_set[name][previous_page][-1]
+        return ""
```
SimonSapin: Yeah, the spec doesn’t say what to do in this case, but I suppose this is sensible.

weasyprint/layout/pages.py
```diff
@@ -453,8 +454,19 @@ def page_width(box, context, containing_block_width):
 def page_height(box, context, containing_block_height):
     page_width_or_height(VerticalBox(context, box), containing_block_height)
 
-
-def make_page(context, root_box, page_type, resume_at, content_empty):
+def box_text_contents(box):
+    if isinstance(box, boxes.TextBox):
+        return box.text
+    elif isinstance(box, boxes.ParentBox):
+        return ''.join(box_text_contents(child) for child in box.children)
```
SimonSapin: Although that is not the common case, this code will create lots of intermediate strings if the tree is deep. Please this line with:

```
return ''.join(b.text for b in box.descendants() if isinstance(b, boxes.TextBox))
```
(

weasyprint/layout/pages.py
```diff
@@ -511,6 +523,13 @@ def make_page(context, root_box, page_type, resume_at, content_empty):
     context.finish_block_formatting_context(root_box)
 
     page = page.copy_with_children([root_box])
+    descendants = list(page.descendants())
```
SimonSapin: This list might consume a lot of memory and doesn’t seem to serve any purpose. Please use the generator directly in the for-loop: `for child in page.descendants()`:.

SimonSapin:
Alright, I’m done reviewing up to 7db9be4005b2188b9674d4baa0b832afd67d88dc. There are some minor issues, but the overall design is simpler than I expected. Good job!

weasyprint/layout/__init__.py
E-M-P-I-R-E: It doesn't crash. The way the defaultdict is structured and evaluated [when resolving the value] is checking the current page for assignments, then previous pages... finally returning an empty string if there were no assignments.

weasyprint/layout/__init__.py
E-M-P-I-R-E: I'm not sure exactly which case you're referring to.

E-M-P-I-R-E:
Oops, I guess I should have responded to some of your comments before pushing again [I was planning on it]. I think I addressed everything in this latest commit. Mostly, changed the validation.

One oddity - you suggested a change  in `TEXT_CONTENT_EXTRACTORS['box_text_contents']`. I wasn't actually using that function (meant to remove that code), so I made that change to the original function in `formatting_structure/build`. I also added a new key `text` which calls that function.

Oh.. I just figured out how to see the specific issues that are causing continuous integration to fail. I'll take care of that.

Two things that I'm unclear about
First - from the spec.:

> User agents must be able to recall many values of the named string, as the string() function can return past, current, or **future values** of the assignment.

Second: should there be a warning when attempting to use named strings outside of the margin-boxes?

Thanks a lot!

weasyprint/layout/__init__.py
SimonSapin: Oh, right, I forgot it was a defaultdict.

E-M-P-I-R-E:
Question was raised:
>Do the 'keyword' and 'name' strings add information here, or could the return value be just (keyword, name)? (Or (name, keyword), to match the order of the syntax?)

I was following the example of the validation function for `bookmark-label` [which returns `('string', token.value)`] but I liked your suggestion and went with that.

`string-set` validation now returns `(var_name, keyword)`. It's only accessed in `layout/pages.make_page()` (when updating the `LayoutContext` with new assignments for a given named-string).

E-M-P-I-R-E:
It was also requested that I explain what I intended to support for `string-set` (in terms of validation).

The answer is one value pair ex:
`-weasy-string-set: header content();`

If no argument is passed to `content()` it defaults to 'text'.

liZe:
Wow. So cool!

I'll review that and merge it soon. Thanks @E-M-P-I-R-E for your work, and @SimonSapin for your review :).

PR #264, opened 2015-08-15
Support 'q', 'vh', 'vw', 'vmin', 'vmax', 'rem' units

liZe:
'q' is a quarter millimeter and is easy to support.

Viewport-percentage lengths are related to the size of the initial containing block, which is the size of the first page area for paged media. Many ways exist to solve them in WeasyPrint, but none is simple as the size of the first page needs to be known (the layout has started) before solving.

The functions solving percentages are already doing this, so I've tried to merge the viewport-percentage and percentage solving functions. The size of the viewport is stored in the layout context and set as soon as the size is known.

This implementation works for most cases, but some problems need to be solved:
- these units may not work yet for properties that don't rely on the `length` computer (quite easy to fix, but may be long to test),
- these units don't work yet for properties relying on the `pixels_only` attribute of the `length` computer (easy but boring to fix),
- these units don't work yet for properties that don't allow percentages (seems easy, but I'm not sure),
- WeasyPrint refuses to render the document when these units are used in properties needed before defining the viewport size. I don't know what to do with that. Moreover, some corner cases are not even solved in the spec. I think that we can safely ignore these cases.

@SimonSapin are you OK with this implementation, or do you have a better idea?

Storing the font size of the root element in the context should be enough to support 'rem' once the first three problems listed above are solved.

### DIFF
--- a/weasyprint/css/computed_values.py
+++ b/weasyprint/css/computed_values.py
@@ -30,6 +30,7 @@ LENGTHS_TO_PIXELS = {
     'in': 96.,  # LENGTHS_TO_PIXELS['pt'] * 72
     'cm': 96. / 2.54,  # LENGTHS_TO_PIXELS['in'] / 2.54
     'mm': 96. / 25.4,  # LENGTHS_TO_PIXELS['in'] / 25.4
+    'q': 96. / 25.4 / 4.,  # LENGTHS_TO_PIXELS['mm'] / 4
 }
 
 
--- a/weasyprint/css/computed_values.py
+++ b/weasyprint/css/computed_values.py
@@ -296,6 +296,13 @@ def length(computer, name, value, font_size=None, pixels_only=False):
             result = value.value * logical_width
         elif unit == 'em':
             result = value.value * font_size
+    elif unit in ('vh', 'vw', 'vmin', 'vmax'):
+        if pixels_only:
+            # TODO: remove the pixels_only attribute from this function and
+            # resolve these values during layout as it's done for percentages
+            return 0
+        else:
+            return value
     else:
         # A percentage or 'auto': no conversion needed.
         return value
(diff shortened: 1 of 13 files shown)


SimonSapin:
`q` unit looks good.

I’m against resolving viewport-percentage units like percentage both because it’s against the spec and because it sprinkles the implementation through the layout code, while resolving them early to pixels allows keeping the implementation contained in the style system.

How viewport-percentage units interact with Paged Media is indeed an open spec issue. For example, how to deal with:
```css
:root { font-size: 2vw; /* 2% of 'size' minus margins/borders/padding on @page */ }
@page { margin: 0; size: 100em; /* em is based on 'font-size' inherited from :root */  }
```

In WeasyPrint, I suggest:

* Optional: have the parser reject viewport units inside `@page` and `@top-center` and friends.
* Extract the beginning of `layout.pages.make_all_pages` into a function that determines whether the first page is a left or right page.
* Add a viewport size parameter to `css.set_computed_styles`, and use it in the `css.computed_values` module to resolve viewport-percentage units to pixels.
* In `css.get_all_computed_styles`:
  * Pick a dummy viewport size. Maybe zero, any stylesheet that relies on it is doing it wrong anyway.
  * Call `set_computed_styles` for the root element with the dummy viewport size. 
  * Use that as the inherited styles (the `parent` parameter) and the dummy viewport size when calling `set_computed_styles` for `@page`.
  * Determine the real viewport size based on `@page` computed values for the first page.
  * Use that for calling `set_computed_styles` for elements, including calling it once again for the root element.

liZe:
This branch is old. @SimonSapin I've tried to follow your hints and even if it looks simple, it was a bit too difficult for me :wink:. If anybody wants to add this feature, it'll be easier to start from the current code, not from this branch. `q` and `rem` are supported, and viewport-percentage lengths are not that useful for print.

PR #266, opened 2015-08-19
Fix letter-spacing handling at block boundaries

jasco:
Pango letter-spacing attribute does not affect the last character in the text range.  This is a particular problem when there are nested tags since each text fragment is treated as a separate block.

This fix appends a zero-width-space to each TextBox so that pango will add letter-spacing following the final character.


jasco:
The flake8 test failed because of some long lines.  I will fix the code line length, but the long lines in test are part of block quotes I'd rather not split for clarity.  Is there a way to disable the length check for specific lines?

liZe:
Thanks for this pull request. Almost one year without a review, shame on me.

I've spent some time during the last months to remove extra characters added for various workarounds (see f0912e4 for example). Adding characters looks harmless, but it sometimes adds bugs that are very hard to find and to fix. If you (or anyone else) find a solution without adding extra characters, I would be perfect!

liZe:
I've fixed this bug without adding an extra character. Thanks!

liZe:
I've merged this PR to get the test, thank you!

jasco:
Thanks for getting to this.  Great to see you found an alternate solution.

PR #280, opened 2015-11-25
Replace utf8 with utf-8 for gettext/django compatibility

alexef:
See: http://stackoverflow.com/questions/4370035/django-makemessages-errors-unknown-encoding-utf8

This will fix:
<pre>
CommandError: errors happened while running xgettext on __init__.py
xgettext: ./venv/lib/python3.4/site-packages/weasyprint/__init__.py:1: Unknown encoding "utf8".    Proceeding with ASCII instead.
xgettext: Non-ASCII string at ./venv/lib/python3.4/site-packages/weasyprint/__init__.py:145.
          Please specify the source encoding through --from-code or through a comment
          as specified in http://www.python.org/peps/pep-0263.html.
</pre>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,4 @@
-# coding: utf8
+# coding: utf-8
 """
     WeasyPrint
     ==========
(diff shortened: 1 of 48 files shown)


SimonSapin:
Sounds like a stupid bug in django-makemessages, but oh well.

PR #291, opened 2016-01-12
Use David Baron's proposal for atomatic table layout

liZe:
Kids,

When you were young, I bet that your dream was to accomplish something big. I mean, really big. Becoming a vet to save puppies and dolphins, going to the moon in a gigantic rocket with lots of blinking buttons, playing the guitar in front of crowds of naked men and women singing your interstellar hits. That's what you wanted, and there was joy, and there was hope, that's the purity of being young.

But now you feel like you're too old, and you think that there's nothing you can do to fix this.

**Good news: you're wrong**. There's one thing bigger than these dreams, even bigger than everything you can imagine. And this bigger thing is AUTOMATIC TABLE LAYOUT. And YOU CAN NOW BE PART OF IT. Magic.

Defining the size of columns in an HTML table is hard. Really, really hard. Even harder. So hard that after years and years, the different browsers continue to give us different results for a given table. So hard that even the CSS-WG gave up on giving an official specification for CSS2.1. So hard that the CSS3-tables draft tells us that "The remainder of this section is non-normative", with a stupid algorithm beginning with these words: "This algorithm may be inefficient".

Yes, kids: during this XXIst century, we want to go to Mars and to cure AIDS, but sadly we know that nobody can find a solution to this automatic table layout problem.

Nobody? Oh my daughter, oh my son, don't cry and don't be so sad, because someone came from the skies to save us, our souls and our tables. And this new kind of god is @dbaron. He wrote what we can call a Bible, because he must have been totally enlightened (or on drugs, or maybe both) to write such a piece of text: [More Precise Definitions of Intrinsic Widths and Table Layout](http://dbaron.org/css/intrinsic/#col-constrain).

I've just implemented quite a big part of it (Automatic Table Layout) and even understood some lines of it. The other parts are available for your own exercise, kids. Just read the Abstract, you'll understand that "min-content width" is just the perfect terminology we'll all have been dreaming about for centuries (replacing "preferred minimum width", and seriously, who's responsible for this stupid naming?).

What's your favourite verse? Mine is:

```
If there are non-constrained columns
That have originating cells
With intrinsic percentage width of 0%
And with nonzero max-content width
(Columns allowed to grow by this rule),
The distributed widths of the columns
Allowed to grow by this rule
Are increased in proportion to max-content width
So the total increase adds to the excess width.
```

This. Is. Fucking. Poetry. Go ahead and read it. Now.

Or, if you're a bit tired, just write more unit tests, and use this branch with your favorite real use cases with percentages and constrained columns, and be enlightened by the beauty of the rendered tables, and try again, report and forgive my bugs as He may forgive yours, until we together finally find the peace and reach the paradise (aka merge).

Amen.

PS: If some of the WeasyPrint prophets (like the famous @SimonSapin, may he be blessed) read this text, they can read and clean the code, as it's been written by a naive mind. I can't personally currently clean it, because each time I open my code editor, I spend days and nights dreaming of border-spacing and "intermediate min-content width for span N (N > 1)" (true story). I'd like to sleep a night or two.

[Review on Reviewable](https://reviewable.io/reviews/kozea/weasyprint/291)

### DIFF
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -86,9 +86,9 @@ CSS 2.1
 The `CSS 2.1`_ features listed here are **not** supported:
 
 * The `::first-line`_ and `::first-letter`_ pseudo-elements.
-* On tables: `empty-cells`_ and `visibility: collapse`_.
-* Minimum and maximum width_ and height_ on table-related boxes and
-  page-margin boxes.
+* On tables: `visibility: collapse`_.
+* Minimum and maximum height_ on table-related boxes.
+* Minimum and maximum width_ and height_ on page-margin boxes.
 * Conforming `font matching algorithm`_. Currently ``font-family``
   is passed as-is to Pango.
 * Right-to-left or `bi-directional text`_.
--- a/weasyprint/css/validation.py
+++ b/weasyprint/css/validation.py
@@ -260,6 +260,13 @@ def border_collapse(keyword):
     return keyword in ('separate', 'collapse')
 
 
+@validator()
+@single_keyword
+def empty_cells(keyword):
+    """``empty-cells`` property validation."""
+    return keyword in ('show', 'hide')
+
+
 @validator('color')
 @single_token
 def color(token):
--- a/weasyprint/draw.py
+++ b/weasyprint/draw.py
@@ -111,7 +111,8 @@ def draw_box_background_and_border(context, page, box, enable_hinting):
             for row_group in box.children:
                 for row in row_group.children:
                     for cell in row.children:
-                        draw_border(context, cell, enable_hinting)
+                        if cell.style.empty_cells == 'show' or not cell.empty:
+                            draw_border(context, cell, enable_hinting)
         else:
             draw_collapsed_borders(context, box, enable_hinting)
     else:
@@ -302,7 +303,9 @@ def draw_table_backgrounds(context, page, table, enable_hinting):
         for row in row_group.children:
             draw_background(context, row.background, enable_hinting)
             for cell in row.children:
-                draw_background(context, cell.background, enable_hinting)
+                if table.style.border_collapse == 'collapse' or (
+                        cell.style.empty_cells == 'show' or not cell.empty):
+                    draw_background(context, cell.background, enable_hinting)
 
 
 def draw_background_image(context, layer, image_rendering):
--- a/weasyprint/draw.py
+++ b/weasyprint/draw.py
@@ -105,15 +105,7 @@ def draw_page(page, context, enable_hinting):
 def draw_box_background_and_border(context, page, box, enable_hinting):
     draw_background(context, box.background, enable_hinting)
     if isinstance(box, boxes.TableBox):
-        for column_group in box.column_groups:
-            draw_box_background_and_border(
-                context, page, column_group, enable_hinting)
-        for row_group in box.children:
-            draw_background(context, row_group.background, enable_hinting)
-            for row in row_group.children:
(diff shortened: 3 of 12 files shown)


SimonSapin:
I’ve only skimmed the core algorithm. Nothing stands out as horribly wrong, but I don’t have the courage of double-checking every line against the spec :]

---

Reviewed 4 of 4 files at r1, 1 of 4 files at r2, 3 of 3 files at r4.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[weasyprint/formatting_structure/build.py, line 560 \[r2\]](https://reviewable.io:443/reviews/kozea/weasyprint/291#-K7vgvNrXMmeIDtVZ4oN:-K7vgvNsvWuG4sQU0Yna:113791355) ([raw file](https://github.com/kozea/weasyprint/blob/72e214ae2ea33adfddb985c741ecad7d3c65e8b2/weasyprint/formatting_structure/build.py#L560)):*
This comment can be replaced by one that says that e.g. `<colgroup span=3>` generates three `TableColumnBox` children:
https://github.com/Kozea/WeasyPrint/blob/b617122af0beec6851528960a937e8b044fadc1c/weasyprint/html.py#L217-L219

---

*[weasyprint/layout/pages.py, line 233 \[r1\]](https://reviewable.io:443/reviews/kozea/weasyprint/291#-K7veh8GVY-SW2lFGTHl:-K7veh8GVY-SW2lFGTHm:658782138) ([raw file](https://github.com/kozea/weasyprint/blob/4898ad980bdba566d84a3234fc5411d94e3dc0c1/weasyprint/layout/pages.py#L233)):*
No need to change this now, but http://rhodesmill.org/brandon/2012/one-sentence-per-line/ would avoid diffs like this

---

*[weasyprint/layout/preferred.py, line 10 \[r1\]](https://reviewable.io:443/reviews/kozea/weasyprint/291#-K7vfH4ENDW91jBdvSON:-K7vfH4ENDW91jBdvSOO:2118675948) ([raw file](https://github.com/kozea/weasyprint/blob/4898ad980bdba566d84a3234fc5411d94e3dc0c1/weasyprint/layout/preferred.py#L10)):*
Also https://drafts.csswg.org/css-sizing/#terms

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/kozea/weasyprint/291#-:-K7vq1bdt8Ke36KDOXBW:-1975054978)*

liZe:
> This comment can be replaced by one that says…
Fixed.

> No need to change this now, but http://rhodesmill.org/brandon/2012/one-sentence-per-line/ would avoid diffs like this
I'm not sure that I'm able to sleep well knowing that a line is 50 characters when is could be 80.

> Also https://drafts.csswg.org/css-sizing/#terms
After "The Terminology", here is "The Return of the Terminology"!

I'm trying right now to generate reports with various real-life tables.

SimonSapin:
> I'm not sure that I'm able to sleep well knowing that a line is 50 characters when is could be 80.

I could be even less than that! Exercise your radical freedom to write short lines.

liZe:
Tested with real-life reports, works very well for me.

liZe:
I've also fixed background for columns and rows. Time to merge?

SimonSapin:
I’m ok with merging this. I’ll leave it to you to push the green button, to the version number dance, etc.

liZe:
The results of the W3C test suite are WONDERFUL. I've fixed some minor regressions, crashes on corner cases, and even added the support of the `empty-cells` property.

PR #295, opened 2016-02-10
Add 'rem' support

fdemmer:
I am using Foundation, that uses the 'rem' unit a lot.

I am not sure about what i am doing here is correctly implementing the spec and the implementation follows "weasyprint conventions". Is there documentation on how weasy works? sequence of operations, what happens where etc?

I read through https://github.com/Kozea/WeasyPrint/pull/264 to maybe figure this out, but I don't need all the viewport related relative units, so I attempted to just solve 'rem' for now.

It's based on the 'font-size' of the root element, so I am passing on the computed style of the 'body' element to `compute()` and use it in `length()`.

My pdf output now works as I'd expect it.

### DIFF
--- a/weasyprint/css/__init__.py
+++ b/weasyprint/css/__init__.py
@@ -264,8 +264,8 @@ def add_declaration(cascaded_styles, prop_name, prop_values, weight, element,
         style[prop_name] = prop_values, weight
 
 
-def set_computed_styles(cascaded_styles, computed_styles,
-                        element, parent, pseudo_type=None):
+def set_computed_styles(cascaded_styles, computed_styles, element, parent,
+                        root=None, pseudo_type=None):
     """Set the computed values of styles to ``element``.
 
     Take the properties left by ``apply_style_rule`` on an element or
@@ -495,9 +495,12 @@ def get_all_computed_styles(html, user_stylesheets=None):
     # their children, for inheritance.
 
     # Iterate on all elements, even if there is no cascaded style for them.
+    root = None
     for element in element_tree.iter():
+        if element.tag == 'body':
+            root = element
         set_computed_styles(cascaded_styles, computed_styles, element,
-                            parent=element.getparent())
+                            root=root, parent=element.getparent())
 
     # Then computed styles for @page.
 
--- a/weasyprint/css/__init__.py
+++ b/weasyprint/css/__init__.py
@@ -274,11 +274,12 @@ def set_computed_styles(cascaded_styles, computed_styles, element, parent,
 
     """
     parent_style = computed_styles[parent, None] if parent is not None else None
+    root_style = computed_styles.get((root, None), {})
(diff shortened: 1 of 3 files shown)


weasyprint/css/__init__.py
```diff
@@ -273,18 +273,18 @@ def set_computed_styles(cascaded_styles, computed_styles,
...
 
     """
-    if parent is None:
-        parent_style = None
-    else:
-        parent_style = computed_styles[parent, None]
-
+    parent_style = computed_styles[parent, None] \
+        if parent is not None else None
+    root_style = computed_styles.get((root, None), {})
```
SimonSapin: `.get()` here should be `[]` indexing. We want this to fail early if the root element’s style can’t be found. This should be fine since `get_all_computed_styles` deals with the root element first.

But this brings the question: what about computing styles for the root element itself? Let’s look at the spec:

https://drafts.csswg.org/css-values/#rem
> Equal to the computed value of font-size on the root element. When specified on the font-size property of the root element, the rem units refer to the property’s initial value.

All in all, please replace this line with:

```python
root_style = computed_styles[root, None] if root is not element else properties.INITIAL_VALUES
```

SimonSapin: Ah, sorry, it’s more subtle than I first thought. `rem` refers to the initial value for `font-size` on the root element, but it still has its usual definition for *other properties* of the root element.

I’ll leave it to you to figure out how to do this exactly. Unless you’d rather I do it, let me know.

SimonSapin:
Thanks for your contribution! This looks good overall, there are just some minor issues noted inline.

It’d be good to add tests for this as well: change `weasyprint/tests/resources/doc1.html` to use `rem` (both in `font-size` and other properties, both on the root element and other elements) and `test_annotate_document()` in `weasyprint/tests/test_css.py` to check computed values.

Let me know if you want to do these fixes and add tests or if you’d rather I do it.

fdemmer:
updated the branch:
- using set()
- use actual root element, not body
- take the initial font-size as input into the style computation for the root element

in my dev environment i get a lot of failing tests, would be great and probably save me a lot of time if you could complete that. thanks!

SimonSapin:
I didn’t know about `merge --squash`, thanks! (I’ve been using `rebase -i`.)

I’ll look into tests later if @liZe doesn’t get to it first ;)

liZe:
Everything is OK on Travis, even with Python 2.6, good news. I've updated the tests and it works very well, but I can't push as it's not my branch.

@SimonSapin if everything is OK for you, I can merge and commit the test after.
@fdemmer I can squash the branch and add you in the AUTHORS :).

fdemmer:
:+1: and thanks for weasyprint!

liZe:
Thank you for your contribution!

SimonSapin:
@lize, wanna add this to http://weasyprint.org/docs/features/ ?

liZe:
Will be done in #297.

PR #299, opened 2016-02-26
Enhance the support of SVG images

liZe:
This commit adds two features:
- don't render SVG files twice anymore; and
- support SVG images with no intrinsic ratio.

This PR is useful for these two features (mainly the SVG rendered only once for me), but also for the tests: the W3C test suite relies on SVG files with no intrinsic size even for some tests unrelated to this feature, we won't be polluted anymore on these tests by this missing feature.

The related W3C tests pass (except one pathological case) with the latest git master version of CairoSVG. The current PyPI version of CairoSVG (1.0.20) has the needed API but has a poor support of SVG files with no intrinsic size.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -12,7 +12,7 @@ WeasyPrint |version| depends on:
 * cairocffi_ ≥ 0.3
 * tinycss_ = 0.3
 * cssselect_ ≥ 0.6
-* CairoSVG_ ≥ 0.5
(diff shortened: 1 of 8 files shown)


liZe:
@SimonSapin anything to say about this?

SimonSapin:
I’ve only skimmed it and don’t see anything egregiously wrong. I’m not planning to do an in-depth review.

PR #323, opened 2016-06-17
Fix CairoSVG 2.0 pre-release dependency in Python 2.x

kdeldycke:
CairoSVG 2.0 drop Python 2.x support. See:
* https://github.com/Kozea/CairoSVG/blob/2.0.0rc2/NEWS.rst#version-20-not-released-yet
* https://github.com/Kozea/CairoSVG/issues/106

While the final revision of CairoSVG 2.0 hasn't been released yet, there is a `CairoSVG-2.0.0rc2` pre-released package available in PyPi. As a result, WeasyPrint deployments on Python 2.x environment with the `pip --pre` parameter fails.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,17 @@
 WeasyPrint changelog
 ====================
 
+
+Version 0.29
+------------
+
+*Not released yet.*
+
+Bug fixes:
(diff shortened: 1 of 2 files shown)


liZe:
Thank you!

kdeldycke:
Thanks for the merge! That was fast! :)

PR #345, opened 2016-08-16
Fix the Windows support

liZe:
I've tried many little things to make WeasyPrint work with both Python 2 and 3, and here's the best solution I've found.

- As explained in #132, the problem is that `pathname2url` doesn't accept bytestrings with Python 3 on Windows. As Python 2 allows bytes and Python 3 allows unicode with all OSes, I've added the `encode` in the `compat` module's `pathname2url` implementation.
- Contrary to what I've said, there's no need to be afraid of returning unicode strings in `path2url`: unicode is actually already returned because of the implicit cast due to `u'abc' + b'abc'` at the end of the function.
- @SimonSapin's idea to use `open` instead of `urlopen` is a good idea, but it's not related to this problem. For the record, I've tried to use `open`, but URIs for Windows filenames have an extra slash before the volume (something like `/C:/path`). This leading path is explained by the RFC and it's handled correctly by `urlopen`, duplicating code to handle it (and other needed features such as guessing mimetypes) in WP may be useless. We can do this later if we want.
- I've put the `FILESYSTEM_ENCODING` variable in the `compat` module: it's not a bad place and avoids a circular import.

Finally, this PR is pretty equivalent to #132, but a bit cleaner :smile:.

It's been tested with Python 2.7 and 3.5, on Windows 10 and Linux. Some tests launched with `py.test` even work on Windows (but some tests make Python crash, fun fun fun).

### DIFF
--- a/weasyprint/compat.py
+++ b/weasyprint/compat.py
@@ -12,6 +12,7 @@
 
 from __future__ import division, unicode_literals
 
+import codecs
 import sys
 import email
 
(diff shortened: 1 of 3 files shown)


SAPikachu:
Thanks! Just tried and it seems to work fine here.

liZe:
Let's go!

Issue #253, opened 2015-05-18
Support HTML presentational attributes


liZe:

SimonSapin:
The `style` attribute is already supported. Do you mean presentational hints? https://html.spec.whatwg.org/multipage/rendering.html#the-css-user-agent-style-sheet-and-presentational-hints

liZe:
Yes, I do!

liZe:
I've tried to support presentational attributes using custom `-weasy-*` functions in user-agent CSS. It's just handling size-related attributes, but this mechanism should be able to handle colors and text alignements too.

See 173d294.

Good news:
- most of the rules described in http://www.w3.org/TR/html5/rendering.html#the-css-user-agent-style-sheet-and-presentational-hints are easy to write in CSS,
- fallback values are possible.

Bad news:
- it's impossible to handle attributes that modify their children's properties (`cellpadding` for example),
- we add another prefixed CSS function.

I'm not sure that's it's done the RightWay™, but I'm interested on working on that feature if anybody's got better ideas (or if my idea is not that bad).

PR #352, opened 2016-08-29
Support presentational hints

liZe:
See #253.

This feature has been implemented in a clean way.

Random thoughts:
- I've added this feature because it's useful for many W3C tests.
- The diff is big but really simple.
- Many hints are given in CSS in the spec, they could (should?) be added in another stylesheet in WP with priority set to 0 (instead of code).
- Merging `find_style_attributes` and `find_presetational_hints` is possible, it would avoid going through the whole tree twice.
- The code is not tested, but the test suites available on `test.weasyprint.org` pass without a crash.
- Calling `check_style_attribute` with constant strings is probably useless.
- This features makes the tests last ~5% more for me, and it's probably much worst on real cases. Should we allow the users to enable/disable it with an option?
- Some rules have not been implemented, mainly because they would be much easier to handle with CSS than with code (one more reason to use an extra stylesheet).
- Some parsing errors are logged, some aren't (who cares?).

Comments are welcome, as usual. I won't merge it right now, I'd like (at least):
- [x] a new option to activate this feature (and keep it deactivated by default),
- [x] a stylesheet for rules described in CSS in the spec,
- [x] tests.

### DIFF
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -33,8 +33,11 @@ Some elements need special treatment:
   or in SVG with CairoSVG_. SVG images are not rasterized but rendered
   as vectors in the PDF output.
 
-HTML `presentational hints`_ (like the ``width`` attribute on an ``img``
-element) are **not** supported. Use CSS in the ``style`` attribute instead.
+HTML `presentational hints`_ are not supported by default, but can be supported:
+
+* by using the ``--presentational-hints`` CLI parameter, or
+* by setting the ``presentational_hints`` parameter of the ``HTML.render`` or
+  ``HTML.write_*`` methods to ``True``.
 
 .. _CairoSVG: http://cairosvg.org/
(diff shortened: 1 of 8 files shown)


liZe:
This PR is ready to be merged. I'm happy with it because:
- it's optional and doesn't impact performances by default,
- its code is clean :bouquet:, with CSS rules when possible and Python code when it's not,
- it's tested and quite reliable,
- it's implementing almost all the rules defined in the spec,
- it's really useful for the W3C tests.

The main problems remaining are not that important:
- CSS rules' attributes values are case-sensitive (we have to implement CSS Selectors Level 4 to fix this),
- some values should raise errors instead of being accepted or silently rejected.

@SimonSapin I think that you don't care about presentational hints, but you can look at the code if you want :smile:.

PR #390, opened 2016-11-17
check for font_config

keithcallenberg:
check for font_config before attempting to add_font_face
This is an attempt to resolve this error that I get from `css/__init__.py:690` upon upgrading to 0.32:
'NoneType' object has no attribute 'add_font_face'


liZe:
Thank you!

jeffgabhart:
Is this a crasher and I should wait for another release before upgrading ?

liZe:
I'll release a new version before the end of the week.

PR #423, opened 2017-02-06
Warn broken cairo version

andreas-p:
Since the broken cairo version is all around (e.g. Debian Jessie), IMHO a warning might help other users.

### DIFF
--- a/weasyprint/fonts.py
+++ b/weasyprint/fonts.py
@@ -23,6 +23,9 @@ from .text import (
     cairo, dlopen, ffi, get_font_features, gobject, pango, pangocairo)
 from .urls import fetch
 
+if cairo.cairo_version() <= 11400:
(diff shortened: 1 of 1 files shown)


liZe:
Good idea, thanks for the PR.

I'd prefer to avoid warnings at the import of the module, because they sometimes appear at the wrong time (for example when you launch `weasyprint --help`). It may be better in `Document._render`, what do you think?

andreas-p:
I see your point, but the Pango>1.38 warning appears at import time of fonts.py as well. Putting it in Document._render would have it appear possibly multiple times.

liZe:
> I see your point, but the Pango>1.38 warning appears at import time of fonts.py as well.

True (and **I** did that :fearful:).

> Putting it in Document._render would have it appear possibly multiple times.

I think that it's not a bad thing.

So let's merge, I'll try to find a better solution before releasing the next version.

PR #446, opened 2017-03-28
Let the logging module do the string interpolation

sunu:
Interpolating manually sometimes leads to UnicodeDecodeError when the argument is a unicode string. Better to let the logging module take care of the interpolation.


liZe:
Thanks a lot!

PR #464, opened 2017-05-04
Assert fail details

hughsw:
This is an RFC, WIP, not a PR for merging.

This PR shows a few examples of how I'd like to add a terse detail string for assert failures.  The rationale for the pattern (stringify a tuple of local objects) is that an assert failure will be dealt with by an expert who will have to examine the code in question, so there's no need for helpful prose.  What can be helpful to the expert (or the end-user who reports a one-off failure) is to report details about the local state that gave rise to the failure.

I intend to apply this pattern to the other 103 asserts in the non-test code of the project.

### DIFF
--- a/weasyprint/layout/blocks.py
+++ b/weasyprint/layout/blocks.py
@@ -370,7 +370,7 @@ def block_container_layout(context, box, max_position_y, skip_stack,
                            device_size, page_is_empty, absolute_boxes,
                            fixed_boxes, adjoining_margins=None):
     """Set the ``box`` height."""
-    assert isinstance(box, boxes.BlockContainerBox)
+    assert isinstance(box, boxes.BlockContainerBox), str(type(box))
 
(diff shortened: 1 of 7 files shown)


weasyprint/layout/blocks.py
```diff
@@ -466,8 +466,8 @@ def block_container_layout(context, box, max_position_y, skip_stack,
             continue
 
         if isinstance(child, boxes.LineBox):
-            assert len(box.children) == 1, (
-                'line box with siblings before layout')
+            assert len(box.children) == 1, str((
+                'line box with siblings before layout:', len(box.children)))
```
hughsw: I will get rid of the prose...

liZe:
> This is an RFC, WIP, not a PR for merging.

Closing, as it will become harder and harder to merge this PR in the future. I've created #600 to keep the idea.

PR #485, opened 2017-07-07
Use cssselect2 instead of cssselect + lxml

### LINKED ISSUE #57: Named pages
css3-page defines "named pages": http://dev.w3.org/csswg/css3-page/#using-named-pages

The only non-trivial part of this is the cascade. Currently we compute in advance styles for every possible page. With named pages we might want to do this lazily.

### LINKED ISSUE #440: Get rid of lxml dependency
If possible, I think it'd be great to get rid of the dependency on `lxml`. I've found it to be quite annoying to install; on Linux it takes a minute to compile, on Mac OS X the first time it'll take like an hour to compile (due to having to `brew install gcc` as well), and on Windows one by default doesn't have a compiler at hand - leaving you to install a pre-compiled version of `lxml`, but apparently not every version has pre-compiled DLLs in place..
By replacing `lxml` by stdlib `xml` some of this pain (imo) goes away. Of course will need some simple unittests to ensure compatibility remains, but other than that.. :)

liZe:
Using cssselect2 makes WeasyPrint free from lxml (close #440), and gives us hope on our long way to close #57. A lot of things in the private API have changed, mainly to bring the base URL everywhere, but there's no real big change.

The only real downside is the loss of the source line in various logging messages. ElementTree and html5lib don't store the `sourceline` attributes in elements (see html5lib/html5lib-python#97), WeasyPrint can't do anything without this.

If anyone is interested in this change (and in cssselect2), please read Kozea/cssselect2#5. And don't hesitate to contribute if you're interested in [CSS Selectors Level 4](https://drafts.csswg.org/selectors-4/)!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,5 +18,6 @@
 /.cache
 
 # Tests
+/weasyprint/tests/.cache
 /weasyprint/tests/test_results
 /weasyprint/tests/w3_test_suite/test_results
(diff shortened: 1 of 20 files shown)


liZe:
Let's go!

PR #500, opened 2017-08-23
Improve readibility of _select_source function

Pacu2:
During development, I've seen this function, thought it would improve readability a little


liZe:
Thanks a lot!

PR #508, opened 2017-09-19
Add missing libpangocairo dependency

jonashaag:
Without this, you'll get 

```
OSError: cannot load library pangocairo-1.0: pangocairo-1.0: cannot open shared object file: No such file or directory.
Additionally, ctypes.util.find_library() did not manage to locate a library called 'pangocairo-1.0'
```

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -98,19 +98,19 @@ Debian 9.0 Stretch or newer, Ubuntu 16.04 Xenial or newer:
 
 .. code-block:: sh
 
-    sudo apt-get install build-essential python3-dev python3-pip python3-cffi libcairo2 libpango-1.0-0 libgdk-pixbuf2.0-0 libffi-dev shared-mime-info
+    sudo apt-get install build-essential python3-dev python3-pip python3-cffi libcairo2 libpango-1.0-0 libpangocairo-1.0.0 libgdk-pixbuf2.0-0 libffi-dev shared-mime-info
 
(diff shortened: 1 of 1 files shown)


liZe:
Thank you! PangoCairo seems to be provided by Pango with other distributions / OSes.

geographika:
Just to note this is also the case on Windows for 0.41

`OSError: cannot load library pangocairo-1.0: error 0x7e.  Additionally, ctypes.util.find_library() did not manage to locate a library called 'pangocairo-1.0'`

Is there an additional step to add to http://weasyprint.readthedocs.io/en/stable/install.html#windows ?

liZe:
> Is there an additional step to add to http://weasyprint.readthedocs.io/en/stable/install.html#windows ?

pangocairo is included in pango and supposed to be compiled when cairo is available. Both gtk2-runtime and gtk3-runtime have pango and cairo included, I don't see why they wouldn't include pangocairo. Could you check that it's included (or not) in your Gtk+ install?

geographika:
@liZe - it does appear to be there - there is a libpangocairo-1.0-0.dll file in GTK3-Runtime Win64\bin
I'll do a few more tests to check it is not something in my environment. Rolling back to 0.4.0 does work though with the same GTK files.

geographika:
@liZe - apologies. I ran through the installation on a clean Windows install and it works fine with 0.4.1, please ignore the above!

PR #534, opened 2017-11-11
Bugfix: handle SVGs containing images embedded as data URIs

### LINKED ISSUE #529: "Failed to draw an SVG image" when including SVG image with raster content
When generating PDF from a HTML file that includes an SVG image which, in turn, contains raster data in the form of `<image xlink:href="data:image/png;base64..." ... />`, an error is generated:
`ERROR: Failed to draw an SVG image at file:///C:/test/logo_error.svg : 'file_obj'`

The platform is Windows 10 Pro, 64-bit, Python 3.6.3 64-bit, GTK from `gtk3-runtime-3.22.21-2017-09-25-ts-win64`

Attached is a set of files that show the problem.

There are three versions of the image with corresponding HTML files that include it.
The PNG version and SVG version without raster content work fine but when trying to generate PDF from 
the file `test_svg_error.html` which uses the image `logo_error.svg`, the error occurs.

[weasyprint_raster_data_in_svg_bug.zip](https://github.com/Kozea/WeasyPrint/files/1440567/weasyprint_raster_data_in_svg_bug.zip)

martijnthe:
The `dict` returned from `open_data_url` does not contain a `file_obj` key. This line was throwing a key error before.
Fix: avoid accessing the non-existent key.

Fixes #529.

### DIFF
--- a/weasyprint/images.py
+++ b/weasyprint/images.py
@@ -121,7 +121,9 @@ class SVGImage(object):
 
     def _cairosvg_url_fetcher(self, src, mimetype):
         data = self._url_fetcher(src)
-        return data.get('string', data['file_obj'].read())
+        if 'string' in data:
+            return data['string']
+        return data['file_obj'].read()
 
     def get_intrinsic_size(self, _image_resolution, font_size):
         # Vector images may be affected by the font size.


martijnthe:
@liZe I'm sorry for not adding a unit test.
I had trouble getting them to run locally properly.

martijnthe:
hrmm.. the failed unit test looks unrelated to the change I made, it also only failed on the macOS build...

```
=================================== FAILURES ===================================
______________________________ test_overflow_wrap ______________________________
    @assert_no_logs
    def test_overflow_wrap():
        def get_lines(wrap, text):
            page, = parse('''
                <style>
                    body {width: 80px; overflow: hidden; font-family: ahem; }
                    span {overflow-wrap: %s; white-space: normal; }
                </style>
                <body style="hyphens: auto;" lang="en">
                    <span>%s
            ''' % (wrap, text))
            html, = page.children
            body, = html.children
            body_lines = []
            for line in body.children:
                box, = line.children
                textBox, = box.children
                body_lines.append(textBox.text)
            return body_lines
    
        # break-word
        lines = get_lines('break-word', 'aaaaaaaa')
>       assert len(lines) > 1
E       AssertionError: assert 1 > 1
E        +  where 1 = len(['aaaaaaaa'])
weasyprint/tests/test_text.py:581: AssertionError
============== 1 failed, 303 passed, 1 xfailed in 100.05 seconds ===============
```

Any ideas @liZe?

liZe:

> hrmm.. the failed unit test looks unrelated to the change I made, it also only failed on the macOS build...

Yes, it's [a bug in Pango 1.40.13 fixed in Pango 1.40.14](http://ftp.gnome.org/pub/GNOME/sources/pango/1.40/pango-1.40.14.news), we can safely ignore the problem.

PR #552, opened 2017-12-16
Update docs/hacking.rst to include required pytest plugins

martijnthe:
The doc says:

> Use the `py.test` command from the `WeasyPrint` directory to run the test suite.

This doesn't work unless these additional `pytest` plugins are installed as well.


liZe:
Thanks for the report and the fix! I've used `.[test]` instead of listing the plugins, we won't have to change this if we use other pytest plugins in the future.

PR #579, opened 2018-02-24
Initial implementation of flexbox

### LINKED ISSUE #324: Support Flexbox
Hi there, I'm looking to switch from wkhtmltopdf to weasyprint but just wanted to check about flex-box support. I can't see any mention of it in issues. Will I have trouble using weasyprint whilst using flex-box in my design?

liZe:
Fix #324, thanks to Lucien!

This PR doesn't provide a perfect support of [flexbox](https://www.w3.org/TR/css-flexbox-1/), but it shouldn't introduce any regressions and doesn't crash with any of the W3C suite tests. And it's not that bad with real-life cases :wink:.

What's not supported yet:
- floats (inside or outside the flex container),
- position of some absolutely-positioned flex children,
- forced and avoided page breaks in flex container and flex items,
- some alignments of flex items with margins,
- `table` and `table-*` flex items,
- many page breaks in flex containers (especially with reordered flex items),
- many `min-width`, `max-width`, `min-height` and `max-height` values of flex items,
- shrink-to-fit size of flex containers,
- collapsed items…

Many TODOs are in `flex.py` and could be quite easily fixed following the spec.

I'll fix some easy-to-fix problems before merging this PR, using the W3C tests and some real case uses. I'll need to add tests in WP too.


liZe:
Time to merge. We have almost 50% of the W3C test suite passing, with a lot of tests relying on Regions or Grid, and some old tests based on old properties that have been removed. There's no crash with the test suite and hopefully no regressions.

Page breaks and floats for example could be highly improved, but fixes can be added on `master`.

Issue #376, opened 2016-10-28
Support @font-face on Windows


liZe:
I've successfully made `@font-face` work on Windows (using fontconfig too) with small changes:
- I've removed `if sys.platform.startswith('win')` from `fonts.py`.
- I've added `'libfontconfig-1'` in `dlopen` for fontconfig (we can probably remove `'fontconfig'` or `'libfontconfig'`) in `fonts.py`.
- I've created a file called `fonts.config` with the content from Shoes3/shoes3#46.
- I've set the environment variable `FONTCONFIG_FILE` to the config file path.
- ~~I've added a `try/except` block around `fonts.py`'s `clean` (removing the temporary files) because Windows says that the file is used by another process.~~ (see #396).

With these changes, `@font-face` works, but some problems remain:
- We have to find a reliable way to create and make fontconfig use `fonts.config`.
- We have to find a way to make this file accurate (find folders containing fonts used by Windows, depending on how Windows is installed).

If anyone is interested in solving this problem, I can provide some help (as long as the question is not related to Windows :smile:).

Issue #587, opened 2018-03-10
pytest fails on Windows 7


Tontyna:
Running unit tests on a clean fresh clone reported lots of failures. Installing Ahem font reduced the failings to 22.

They can be grouped into 4 categories:

- CairoError
- No `@font-face` on Windows
- Encoding and other platform specific issues
- Rendering issues

After de-bugging to the best of my knowledge I was able to reduce the failing test number to 8.

Tested with Python 3.6.4 (v3.6.4:d48eceb, Dec 19 2017, 06:54:40) on Windows 7 (64 bits) and Cairo 1.15.6.
Updating Cairo to 1.15.10 didn't help a lot (passed +=1).

CairoError
-----------

Exploring the reason for the 55 occurences of
```
E cairocffi.CairoError: cairo returned 41: b'error occurred in the Windows Graphics Device Interface'
```

I finally spotted the culprit in `test_api.py` at line 635 in `test_bookmarks()`:
```python
assert_bookmarks('''
    <style>* { height: 10px; font-size: 0 }</style>
    ....
```

The `font-size: 0` kills Pango and provokes the CairoErrors in subsquent unit tests.

Inside pytest the Pango-WARNING isn't visible. When I run the snippet outside of pytest I see
```
(python.exe:7200): Pango-WARNING **: scaled_font status is: error occurred in the Windows Graphics Device Interface
```
Why doesn't this happen on Linux machines? Don't know. On Windows it's probably related to the way how DLLs are loaded. I guess, Python loads Pango/Cairo only once at startup/when required...

In any case, replacing `font-size: 0` with `line-height:0` left me with 12 failures.

No `@font-face` on Windows
------------------------------

The 3 tests in test_fonts.py cannot pass on Windows because of:

```
UserWarning: @font-face is currently not supported on Windows
```

Out of curiosity I installed the weasyprint.otf from the ./resources/ folder and have the sneaking suspicion that this font **must** be loaded via special treatment (using FontConfig instead of Cairo) because otherwise not even the kernings and ligatures (being on by default) are respected, let alone turning them on and off via `font-feature-settings` or `font-variant`.

Thats how "kkliga", styled with font-family weasyprint, looks when rendered as PDF or in a M$Word 2010 document:
![test_font_face](https://user-images.githubusercontent.com/35614937/37236818-94f25d5a-2403-11e8-9693-a94640dedd26.png)
Red color applied to "liga" for clarification.

[...]

PR #592, opened 2018-03-15
support @font-face on Windows

Tontyna:
UPPs, the branch name is a bit ... misleading ...

Since at the moment I have no access to another OS than Windows and the existing code in `weasyprint/fonts.py` probably was ok for other platfoms, I implemented the fix in a way that only affects `sys.platform.startswith('win')`.

Switching to FreeType rendering not only enables `@font-face`; suddenly other CSS properties, like `font-stretch` and `font-variant`, start to work as expected -- see  #587 -- even the killer css `font-size: 0` doesn't crash Pango anymore!

There is one minor drawback on Windows: FontConfiguration can't delete the temporary font files it creates during rendering. I think Pango keeps the file handles open (until the Python process is finished), couldn't find a way to persuade it to let them go...

FontConfig and PangoFT2 not only require the proper libraries but also a working font configuration.

Maybe on Nix it never happens that those config files are missing, but [Tom Schoonjans' GTK3 Runtime](https://github.com/tschoonj/GTK-for-Windows-Runtime-Environment-Installer), recommended in the WeasyPrint docs [Installing on Windows](http://weasyprint.readthedocs.io/en/latest/install.html#windows), though installing all required DLLs, doesn't create the `/etc/fonts` folder. I created an issue, status == open.

Lucky me having Inkscape installed with an appropriate `etc/fonts`...

That's why (besides trying and catching `dlopen`), before actually using FontConfig, the code checks whether it has access to at least one font.
If that's the case, font rendering is done via FreeType, otherwise it falls back to the, `@font-face`-less, native system fonts rendering mode.

I tried to support the fontconfig library with an `os.environ['FONTCONFIG_PATH'] = 'default/etc/fonts/we/could/add/to/weasyprint'` when it complains about a missing `fonts.conf`.
Doesn't work. Probably again due to the way how DLLs are loaded in Python. The DLL only sees the global environment that the python script is running in. Changes during runtime don't reach the DLL.
Contemplated about `FcConfigAppFontAddDir(<WINDOWSFONTDIR>)` to circumvent such a situation...

Windows-testers welcome!

### DIFF
--- a/weasyprint/fonts.py
+++ b/weasyprint/fonts.py
@@ -33,17 +33,33 @@ class FontConfiguration:
         """Add a font into the application."""
 
 
-if sys.platform.startswith('win'):
-    warnings.warn('@font-face is currently not supported on Windows')
-elif pango.pango_version() < 13800:
+fontconfig = None
+pangoft2 = None
+
+if pango.pango_version() < 13800:
     warnings.warn('@font-face support needs Pango >= 1.38')
 else:
+    try:
+        fontconfig = dlopen(ffi, 'fontconfig', 'libfontconfig',
+                            'libfontconfig-1.dll',
+                            'libfontconfig.so.1', 'libfontconfig-1.dylib')
+        pangoft2 = dlopen(ffi, 'pangoft2-1.0', 'libpangoft2-1.0-0',
+                          'libpangoft2-1.0.so', 'libpangoft2-1.0.dylib')
+    except Exception as err:
+        warnings.warn("'@font-face not supported: {0}".format(err))
+        fontconfig = None
+        pangoft2 = None
+
+
+# if both libraries are present: Use them
+if fontconfig and pangoft2:
     ffi.cdef('''
         // FontConfig
 
         typedef int FcBool;
         typedef struct _FcConfig FcConfig;
         typedef struct _FcPattern FcPattern;
+        typedef struct _FcStrList FcStrList;
         typedef unsigned char FcChar8;
 
         typedef enum {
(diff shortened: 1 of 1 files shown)


liZe:
Thank you :heart:

> Since at the moment I have no access to another OS than Windows and the existing code in weasyprint/fonts.py probably was ok for other platfoms, I implemented the fix in a way that only affects sys.platform.startswith('win').

That's a good idea.

> Switching to FreeType rendering not only enables @font-face; suddenly other CSS properties, like font-stretch and font-variant, start to work as expected -- see #587 -- even the killer css font-size: 0 doesn't crash Pango anymore!

Great! It would be a good idea in the future to find how to avoid crashes with font-size: 0, but that's another story. There were crashes on Unix too, but they were avoided by fed326d.

> There is one minor drawback on Windows: FontConfiguration can't delete the temporary font files it creates during rendering. I think Pango keeps the file handles open (until the Python process is finished), couldn't find a way to persuade it to let them go...

That's not the first time we hit this problem, I'm not sure there's a solution.

> Maybe on Nix it never happens that those config files are missing, 

True.

> but Tom Schoonjans' GTK3 Runtime, recommended in the WeasyPrint docs Installing on Windows, though installing all required DLLs, doesn't create the /etc/fonts folder. I created an issue, status == open.

(tschoonj/GTK-for-Windows-Runtime-Environment-Installer#14 for the record)

> That's why (besides trying and catching dlopen), before actually using FontConfig, the code checks whether it has access to at least one font.
> If that's the case, font rendering is done via FreeType, otherwise it falls back to the, @font-face-less, native system fonts rendering mode.

It's perfect.

I merge and I'll try to make sure that it doesn't hurt other OSes and Windows users with no usable FontConfig.

liZe:
Fix #376.

Tontyna:
As promised, Tom Schoonjans' [today's update](https://github.com/tschoonj/GTK-for-Windows-Runtime-Environment-Installer/commit/1595c18695c42a6c1c0ba6e7accdaf2f216b9b3e) of the *GTK+ for Windows Runtime Environment Installer* includes the fontconfig configuration files.

PR #594, opened 2018-03-19
Clean tests

liZe:
- Use pytest features when possible.
- Split long repetitive tests into small unit tests.
- Split long files.


Tontyna:
While you're at it:
I'd like to see `/.pytest_cache` in `.gitignore`

PR #595, opened 2018-03-20
Ensure 'str' and trailing slash in path2url

Tontyna:
This PR fixes the errors caused by `weasyprint/urls/path2url`, detected when pytesting on Windows, see  #587.

It ensures that Python's `urllib.request.pathname2url()` isn't fed with binary pathnames and takes care of the trailing slash for directories.


liZe:
Looks like a good idea and shouldn't break anything, that's cool. Thank you!

PR #599, opened 2018-03-22
Windows default encoding isn't utf-8!

Tontyna:
See #587, UnicodeDecodeError

### DIFF
--- a/weasyprint/tests/test_api.py
+++ b/weasyprint/tests/test_api.py
@@ -151,7 +151,7 @@ def test_html_parsing():
 
     py.path.local(os.path.dirname(__file__)).chdir()
     filename = os.path.join('resources', 'doc1.html')
-    with open(filename) as fd:
+    with open(filename, encoding='utf-8') as fd:
         string = fd.read()
     _check_doc1(FakeHTML(string=string, base_url=filename))
     _check_doc1(FakeHTML(string=string), has_base_url=False)


liZe:
Thank you!

It's OK for the UnicodeDecodeError.

For Cairo, I'd really like (if possible) to avoid the crash with `font-size: 0` instead of changing tests. Do you get a useful backtrace with these crashes?

Tontyna:
There are more than 60 snippets using `font-size: 0` in the test suite -- and only the one in test_api.test_bookmarks seems to crash Pango. Why? What's so special there? Why doesn't it crash when FreeType rendering is used? I don't get it.

The Pango crash itself is invisible/suppressed in the pytest; test_api.**test_bookmark** passes, no problem.

Parsing the snippet on its own produces the following bewildering messages:
```
_cairo_win32_scaled_font_set_metrics:GetTextMetrics: Das System kann die angegebene Datei 
nicht finden.

(python.exe:5168): Pango-WARNING **: failed to create cairo scaled font, expect ugly output. 
the offending font is 'DejaVu Serif Bold 0'

(python.exe:5168): Pango-WARNING **: font_face status is: no error has occurred

 <PageBox PageType(side='right', blank=False, first=True, name=None)> (python.exe:5168): 
Pango-WARNING **: scaled_font status is: error occurred in the Windows Graphics Device Interface
= 643 x 972 @ 0, 0
  _cairo_win32_scaled_font_set_metrics:GetTextMetrics: Der Vorgang wurde erfolgreich beendet.
```
What? What is Windows trying to tell?
In the first line it complains: "Das System kann die angegebene Datei nicht finden" -- meaning: Cannot find the requested file.
In the last line it has changed its mind: "Der Vorgang wurde erfolgreich beendet" -- Operation successfully completed, thank you very much.
Inbetween Pango, unable to decide whether an error happended or not. 

BTW: 'DejaVu Serif Bold 0' can't be ugly because it is invisible.

The induced crashing cairocffi.context.Context produces about 1000 to 2000 log lines per failing test. Dont think that the backtraces are very helpful. Nevertheless, if you want me to, I'll try to make an extract of maybe relevant lines.

I suspect it's another missing DLL, will investigate `_cairo_win32_scaled_font_set_metrics`.

Tontyna:
Managed to capture useful error log by parsing the following snippet:

```html
<div>
   prepare crash
</div>
<div style="font-size: 0; break-before:always">
  crash with font-size 0 on page 2
</div>
<div style="break-before:always">
   page 3
</div>
```

The "prepare crash" is important. Without it -- no problems. Page breaks for proof of speculations.

Output with `logging.INFO` level:

```
Step 1 - Fetching and parsing HTML - crashtest.html
C:\...\weasyprint\fonts.py:173: UserWarning: @font-face not supported: Cannot load default config file
  warnings.warn(msg)
Step 3 - Applying CSS
Step 4 - Creating formatting structure
Step 5 - Creating layout - Page 1
```

Until now: everything is ok/as expected.
On Page 1 no offending font in use. Probably a default fallback font with a size other than zero.

On Page 2 the div with font-size 0 hits Cairo/Pango:

```
Step 5 - Creating layout - Page 2
_cairo_win32_scaled_font_set_metrics:GetTextMetrics: Der Vorgang wurde erfolgreich beendet.
(python.exe:6656): Pango-WARNING **: failed to create cairo scaled font, expectugly output. the offending font is 'DejaVu Serif 0'
(python.exe:6656): Pango-WARNING **: font_face status is: error occurred in the Windows Graphics Device Interface
(python.exe:6656): Pango-WARNING **: scaled_font status is: error occurred in the Windows Graphics Device Interface
```

The offending code seems to be located in the function `_cairo_win32_scaled_font_set_metrics` in the file [`cairo/src/win32/cairo-win32-font.c`](https://cgit.freedesktop.org/cairo/tree/src/win32/cairo-win32-font.c). Though not being a C programmer the GDI handling therein looks dubious to me. I suspect, when GetTextMetrics fails it leaves the global(!) static HDC in a critical state (the spoilt Zero-Size-Font isn't deselected).
Cf. my previously posted error log. The messages seem to vary. "Der Vorgang wurde erfolgreich beendet" is the Windows System Message for ERROR_SUCCESS. Probably retrieved by `GetLastError`. Which -- if not called immediately after the error happened -- returns, yes: the updated last error...

In any case: No reason to stop or crash -- not yet.
Carry on with pagination.

```
Step 5 - Creating layout - Page 3
```

The Zero-Font-Size crashes Cairo -- not Pango! -- when we actually try to draw on a surface.
Thats the reason why in the test suite the killer test_api.test_bookmarks seems to pass but all the subsequent tests that call `write_pdf`, produce this vast amout of CairoErrors.

```
Step 6 - Drawing
  File "C:\...\weasytest.py", line 280, in execute
    document.write_pdf(target=output)
  File "C:\...\weasyprint\document.py", line 494, in write_pdf
    surface.show_page()
  File "c:\...\python36\lib\contextlib.py", line 99, in __exit__
    self.gen.throw(type, value, traceback)
  File "C:\...\weasyprint\draw.py", line 89, in stacked
    context.restore()
  File "c:\...\python36\lib\site-packages\cairocffi\context.py", line 178, in restore
    self._check_status()
  File "c:\...\python36\lib\site-packages\cairocffi\context.py", line 108, in _check_status
    _check_status(cairo.cairo_status(self._pointer))
  File "c:\...\python36\lib\site-packages\cairocffi\__init__.py", line 79, in _check_status
    raise exception(message, status)
cairocffi.CairoError: cairo returned 41: b'error occurred in the Windows Graphics Device Interface'
```

That's the short version of the backtrace. Unshortened there are lots of "During handling of the above exception, another exception occurred" only repeating that the `<cairocffi.context.Context>` is scrambled.

**Summary:** 
Wait for (or contribute to) the next Cairo release. Until then avoid `font-size: 0` on Windows when your libfontconfig is inoperative.

Will undo the last commit in this PR.

liZe:
Thank you really much, that's an impressive report.

I've had bad times reporting Cairo bugs for corner cases, we should try to find a workaround in WeasyPrint in my opinion.

https://github.com/Kozea/WeasyPrint/blob/34e29926b56277c3714edf28896455be4f629a0b/weasyprint/draw.py#L1008-L1011

I don't exactly remember what's been done to avoid text with `font-size: 0`, but I remember that WeasyPrint used to crash. This assertion and its comment seem to imply that 0-sized text is not drawn anymore. As you said, the problem comes for sure when font metrics for a 0-sized font are required and then some normal text is drawn.

I think that the font metrics are requested when the layout is created. Maybe we should try to prevent the creation of these layout (they're useless by the way). I'll try to find a quick fix if possible.

liZe:
Here's a patch that may help:

```diff
2 files changed, 12 insertions(+)
weasyprint/css/computed_values.py |  2 ++
weasyprint/text.py                | 10 ++++++++++

modified   weasyprint/css/computed_values.py
@@ -304,6 +304,8 @@ def length(computer, name, value, font_size=None, pixels_only=False):
             result = value.value * font_size * ex_ratio(computer.computed)
         elif unit == 'ch':
             # TODO: cache
+            if font_size == 0:
+                return 0
             layout = text.Layout(
                 context=None, font_size=font_size,
                 style=computer.computed)
modified   weasyprint/text.py
@@ -626,6 +626,9 @@ def first_line_metrics(first_line, text, layout, resume_at, space_collapse,
 class Layout(object):
     """Object holding PangoLayout-related cdata pointers."""
     def __init__(self, context, font_size, style):
+        # Cairo crashes with font-size: 0
+        assert font_size
+
         self.context = context
         hinting = context.enable_hinting if context else False
         self.layout = ffi.gc(
@@ -690,6 +693,8 @@ class Layout(object):
 
     def set_tabs(self, style):
         if isinstance(style['tab_size'], int):
+            if style['font_size'] == 0:
+                width = 0
             layout = Layout(
                 context=self.context, font_size=style['font_size'],
                 style=style)
@@ -838,6 +843,11 @@ def create_layout(text, style, context, max_width, justification_spacing):
         or ``None`` for unlimited width.
 
     """
+    if style['font_size'] == 0:
+        # Cairo crashes with font-size: 0
+        # See https://github.com/Kozea/WeasyPrint/pull/599
+        return Layout(context, 1, style)
+
     text_wrap = style['white_space'] in ('normal', 'pre-wrap', 'pre-line')
     if not text_wrap:
         max_width = None
```

Tontyna:
Youre right, the assertion in draw.draw_text comes too late. Respectively it doesn't help. When the given snippet hits `draw_text` the first time, it's with the "prepare crash" text, having a `font-size: 16`. And CAIRO CRASHES.
I'll try the patch tomorrow and report what happens.

Tontyna:
Why assert something when we can prevent it?

```python
# weasyprint/text.py, line 626 ff
class Layout(object):
    """Object holding PangoLayout-related cdata pointers."""
    def __init__(self, context, font_size, style):
        # Cairo crashes with font-size: 0
        # See https://github.com/Kozea/WeasyPrint/pull/599
        if font_size == 0:
            font_size = 1
```

No need to alter other source files. One location to catch them all.

Works for the above test case (hitting text.create_layout).
Did one test with `width: 10ch`, hits computed_values.length with `unit == 'ch'` -- looks ok.
Did one test with `tab-set: 20`, hits text.set_tabs -- looks ok.

I'm not as familiar with the code base as you, maybe there are reasons not to do it like I propose?

liZe:
> I'm not as familiar with the code base as you, maybe there are reasons not to do it like I propose?

If you set font-size to 1 and don't catch `font-size: 0` everywhere else, you'll draw small letters instead of drawing nothing, and you'll break the layout: text doesn't have the same size when drawn with `font-size: 0` and `font-size: 1px`.

Tontyna:
You're right, was an error in reasoning. 

`computed_styles.length` should indeed return 0 immediately.
But then, your patch for `text.create_layout` isn't soo much inerter than mine ;)
And `text.set_tabs` should be patched with an additional `else` and indent.

No, this doesn't make me happy. Feels like makeshift.

Considering that this error only happens on Windows when Pango/Cairo switch to Win32 API.
Considering that text rendered with Win32 API is anyway buggy/ugly.
Considering that there will be a wonderful documentation on how to install GDT/FontConfig on Windows.

Considering that, I think using a non-crashing 1px font instead of a 0px font is ok.

Anyway, will implement and test the `assert font-size` stopgap if you vote for it.

liZe:
> And text.set_tabs should be patched with an additional else and indent.

True.

> Considering that this error only happens on Windows when Pango/Cairo switch to Win32 API.
> Considering that text rendered with Win32 API is anyway buggy/ugly.
> Considering that there will be a wonderful documentation on how to install GDT/FontConfig on Windows.
> Considering that, I think we should ignore the 0px issue.

You're right, let's forget about this for now.

(I like the third point :smile:)

liZe:
I've used `@font-face` for tests, removing the need of installing Ahem. I'll also fix the list of fonts used for some tests depending on the OS.

Tontyna:
Good idea. I'll run the unit tests with your latest commits. Feedback asap.

Tontyna:
Uninstalled my Ahem font.

pytest with working fontconfig/FreeType -- no errors.
pytest with WinAPI rendering -- 36 tests fail. **No CairoError**

Didnt yet check how many tests fail because of unsupported `@font-face`. Those tests could be decorated with `@pytest.mark.skipif(fonts.ZERO_FONTSIZE_CRASHES_CAIRO == True)` 
Ok, the variable could/should be renamed to `CAIRO_SWITCHED_TO_TOY_TEXT_API` or
 `FONTCONFIG_OUT_OF_ORDER`

liZe:
@Tontyna Guess [what's happening](https://travis-ci.org/Kozea/WeasyPrint)?

Tontyna:
@liZe Now, what's the appropriate emoji? Blink-and-smile, blink-and-smile :sheep: :sheep: :sheep:

PR #612, opened 2018-04-03
Dont floor PDF surface size

### LINKED ISSUE #606: Round PDF width and height instead of flooring
My generated pdf  using weasyprint has size of 8.27*11.68 inch.
Could you please let me know how can I fix it via weasyprint?

Tontyna:
Should fix #606

### DIFF
--- a/weasyprint/document.py
+++ b/weasyprint/document.py
@@ -487,10 +487,10 @@ class Document(object):
         LOGGER.info('Step 6 - Drawing')
         for page in self.pages:
             surface.set_size(
-                math.floor(scale * (
-                    page.width + page.bleed['left'] + page.bleed['right'])),
-                math.floor(scale * (
-                    page.height + page.bleed['top'] + page.bleed['bottom'])))
+                int(round(scale * (
+                    page.width + page.bleed['left'] + page.bleed['right']))),
+                int(round(scale * (
+                    page.height + page.bleed['top'] + page.bleed['bottom']))))
             with stacked(context):
                 context.translate(
                     page.bleed['left'] * scale, page.bleed['top'] * scale)


liZe:

There's no good solution about PDF size. We must have a PDF size with an integer number of PDF points (1 PDF point = 1/72 inch), so your solution doesn't really fix your problem: you can now report a new issue asking for a document size of 8.268×11.693 :wink:.

More seriously, rounding instead of flooring can create small white stripes at the right and at the bottom of the page, when rounding makes the PDF size greater than the drawing surface. That's why we preferred flooring. Moreover, [ISO 216 gives a tolerance of ±2.0mm](https://en.wikipedia.org/wiki/ISO_216#Tolerances) for A4 sizes and our floored A4 size is 209.9×296.6mm. Our A4 document is OK according to the specification!

A better solution would to round the size but also to scale the drawing surface to match the rounded size. What do you think about that?

Tontyna:
I think its not worth the effort.

Neither the floored nor the rounded MediaBox results in a perfect A4 paper size. No chance with 1/72" units.

As you said: The floored values are within the spec and my PDF readers identify it as A4. No problems when printing.

@hajarghaem: Why do you **need** 11.69 inches?

hajarghaem:
@Tontyna: becasue I want to make the pdf same as an example pdf that I have. And its size is 8.27 inch * 11.69inches. I can make 11.69inch by using 
"@page{
 size: 8.278in 11.70in;"
}"
infact when I say 11.70in then it gaves me a pdf with 11.69 that is also an error!
And when I use 8.27in then it gives me 8.28inch that is not same way it works for height!
So I can make 11.69 using 11.70 but I cannot make 8.27 using 8.28!

Could you please let  me know your thought?

Tontyna:
As @liZe said: PDF paper sizes can only be multiples of 1/72" aka pt.
But though, you're right, it's a case of trial-and-error to produce a MediaBox with a height of 842pt / 11.6944444". Thats -- I think -- mainly due to the internal calculations done in css pixels plus the flooring to avoid white stripes... 

Finally succeded with that counter-intuitive css:
```
@page{
  size: 595pt 842.5pt;
}
```

PR #620, opened 2018-04-22
rectify index

Tontyna:
Did I get you right? 
In the `content()` function in weasyprint.css.validation.properties -- the purpose of the `parsed_tokens` is only a stub to cover the spec's comma separated URI list. Since this 'alternate resources with fallback' concept isn't (yet) implemented they are silently discarded.


liZe:
> Did I get you right?

Of course.

PR #621, opened 2018-04-23
Dont crash when element/target-collector is None

Tontyna:
Concerning `compute_attr_function()`: Why not return the fallback?

Concerning `_content_list()`: target_collector is None when called in `@page` context. Not shure whether the spec forbids target-* in that context. In any case: Since all the anchors are already collected when we hit `compute()` within the page margins, no harm is done by not calling `target_collector.collect_computed_target()`. A css like that works as expected:

```css
@page {
  @top-center {
    content: "countervalue at #id: " target-counter("#id", countername);
  }
```
That is: Page margins gain access to targets in the document without a detour via `string-set/string()`. OMG, I fear thats not allowed by the spec...

BTW: With your cleanups it's almost a pleasure to read the code :smiley:

### DIFF
--- a/weasyprint/css/computed_values.py
+++ b/weasyprint/css/computed_values.py
@@ -400,8 +400,11 @@ def compute_attr_function(computer, values):
     func_name, value = values
     assert func_name == 'attr()'
     attr_name, type_or_unit, fallback = value
-    attr_value = computer.element.get(attr_name, fallback)
+    # computer.element sometimes is None
+    # computer.element sometimes is a 'PageType' object without .get()
+    # so wrapt the .get() into try and return None instead of crashing
     try:
+        attr_value = computer.element.get(attr_name, fallback)
         if type_or_unit == 'string':
             pass  # Keep the string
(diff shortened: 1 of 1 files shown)


liZe:
> Concerning compute_attr_function(): Why not return the fallback?

Returning some type when the caller is waiting for another type may break the caller. Oh, and I'm pretty sure nobody actually cares!

> Page margins gain access to targets in the document without a detour via string-set/string(). OMG, I fear thats not allowed by the spec...

:stuck_out_tongue_closed_eyes:

Looks like it's not forbidden.

> BTW: With your cleanups it's almost a pleasure to read the code

It's not perfect but it's much better like this :wink:.

One last question: when is computer.element `None`?

Tontyna:
> Oh, and I'm pretty sure nobody actually cares!

Right you are.

computer.element is `None` when called (via computed_from_cascaded) by 
layout.inlies.first_letter_to_box(), layout.pages.make_margin_boxes.make_box()  and formatting_structures.boxes.Box.anonymous_from()

PR #624, opened 2018-05-01
no need to try/catch dlopen()

Tontyna:
If there's no fontconfig library, cairocffi already crashed the script, same is true for pangocairo requiring pangoft2 -- so no need to try/catch in fonts.py


liZe:
> so no need to try/catch in fonts.py

Removing code is my favorite activity.

PR #626, opened 2018-05-02
Consolidate space after/before list markers

### LINKED ISSUE #608: No space after inside list markers
```
<ul style="list-style-position: inside"><li>foo
```

This produces "•foo". Per CSS Counter Styles and CSS Lists, it should be "• foo" (which matches what every browser does).

Tontyna:
Should fix #608

### DIFF
--- a/weasyprint/formatting_structure/build.py
+++ b/weasyprint/formatting_structure/build.py
@@ -478,6 +478,7 @@ def add_box_marker(box, counter_values, get_image_from_uri):
         if type_ == 'none':
             return
         counter_value = counter_values.get('list-item', [0])[-1]
+        # TODO: rtl numbered list has the dot on the left!
         marker_text = counters.format_list_marker(counter_value, type_)
         marker_box = boxes.TextBox.anonymous_from(box, marker_text)
     else:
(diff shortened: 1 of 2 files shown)


Tontyna:
I used a 0.5em (in the boxes font size) margin because that was the "space" that WeasyPrint applied until now to the `outside` lists, and until now nobody was complaining.

Didn't check the spec, always wondered how the browsers decide on the various distances in their lists and always had to apply css fixes&forces to make them display the way I wanted them.

gsnedders:
@Tontyna `outside`'s positioning is basically undefined and really inconsistent (and causes web compat bugs!) and the space suffix is white-space collapsed. `inside` at least is relatively consistent.

Tontyna:
Ok. A space/blank char after an inlined **text** marker. What should separate an **image** aka `list_style_image` marker?

gsnedders:
See https://drafts.csswg.org/css-lists-3/#content-property (ignore the "not ready for implementation" box; this bit is pretty accurate).

liZe:
I'd also like to avoid modifying the `style` objects, as it's shared by different boxes and may be frozen in the future if I ever find a workaround for the few cases left.

(And I like the space too, but you already knew that :innocent:.)

Tontyna:
OMG! Not even my browser implements that draft about `::marker`s. I confess: I have no desire at all to dive into that...

What i **can** offer instead of an 0.5em space and although the difference will almost be imperceptible: 
A blank char after those inline markers.

Tontyna:
@liZe: So we stay with the 0.5em space? (Didnt spot your comment at once, was busy reading the draft :smile: ) 
And some day in the future somebody will implement the full *CSS Lists Module Level 3*

liZe:
> So we stay with the 0.5em space?

No problem. I'd just love to modify `marker.margin_*` and let `marker_box.style['margin_*']` unchanged.

Tontyna:
You mean: `marker.margin_*` for the `outside` markers and a space char after `inline` text marker and an additional anonymous TextBox with a space cah for `inline` image markers?

liZe:
OK. As inside markers can't be modified in `markers.py`, let's forget what I said. I'll add a `style.copy` to avoid any possible problem with `style` modification, and this code will go away when List 3 is implemented.

Issue #635, opened 2018-06-07
Stable version user opinion survey


liZe:
It's time to make WeasyPrint great (again)!

We're about to release the first stable version of Weasyprint and we would like to know the users' opinions. Whether you're a former, regular or massive user of WeasyPrint, we'd like to have your answers to these questions:

- Why did you use WeasyPrint?
- Why was it useful for you?
- What were the main problems? 
- What best describes your choice of WeasyPrint?

(You'll find more about new features in [the next milestone page](https://github.com/Kozea/WeasyPrint/milestone/10?closed=1)).

Thank you!

Tontyna:
Though my favourite writing mode is plain text (with lightweight markup), sooner or later there is always a fuelling demand for formatting and proper pagination of joined documents. Not to mention that paper protects against digital dark age.

Converting to nicely styled HTML is no problem at all. But...

...until I discovered WeasyPrint I imported the HTMLs into Word or OpenOffice and exported as PDF. Textprocessors are good when it comes to pagination but no script no regex no macro no template could prevent them from applying unwanted styles and doing other funny things, causing a lot of stupid manual reworking.

Since my css are not too complex, WeasyPrint's output is perfect.

The only thing I missed was its lack of automatic TOC generation -- no problem: it's Open Source and there's docs.python.org. Now I have a (rather ugly, but working) script which does what I need.

Ok, there is another obstacle, at least for Windows people: The installation. 
When I first read the list of prerequisites and that kittens will die because of me, I decided to give *wkhtmltopdf* a chance and even started learning *LaTex*. But OMG! No!
So I finally followed the instructions, mumbling, grumbling, what an expense for only a little tool!, almost didn't cheat -- and no kitten died.

Browsers will eventually implement CSS Paged Media, but until then: I'll use WeasyPrint!

pothos:
I convert pandoc markdown+html/css documents to PDF with WeasyPrint. The documents range from letters to scientific writings. I specially like it because of good native hyphens, and multi column support.

Unfortunately I sometimes need to fallback to wkhtmltopdf or chromium because of JS usage for mathjax and reveal.js (until I find a way to preprocess this with pandoc).

tokyovigilante:
I've only just discovered WeasyPrint after dabbling with LaTeX years ago for scientific writing, but going back to Word after being disappointed with the complexity and (IMO) unattractive output. 

[...]

PR #384, opened 2016-11-03
Better performance

liZe:
The main point of this pull request is to make `StyleDict` inherit from `dict` instead of the copy-on-write-like strategy used before.

This change really improves the speed (yeah, profiling helps), but unfortunately has a high memory footprint. That's why memory usage has been also improved, including an impressive don't-store-the-dummy-cairo-context-in-the-layout-forever in 152193c (already incluled in the `font-face` branch).

The commits are small and hopefully easy to understand and to revert if needed. I've got exactly the same PDFs before and after these commits with many examples, but I'm sure that bugs have been introduced.

Here are the results (with reliable science inside):

## Offline HTML5 Specification

https://www.w3.org/TR/html5/

Long document with a lot of lists and underlined links.

0.31:
- 6.0s
- 146MB memory peak

font-face:
- 6.3s (+5%)
- 130MB memory peak (-11%)

font-face + perf:
- 4.5s (-25%)
- 159MB memory peak (+9%)

## Offline Alice Adventures in Wonderland

https://www.gutenberg.org/files/11/11-h/11-h.htm

Long document with repetitive justified text.

0.31:
- 12.5s
- 162MB memory peak

font-face:
- 9.8s (-21%)
- 134MB memory peak (-17%)

font-face + perf:
- 7.1s (-43%, **too good to be true**)
- 161MB memory peak (=)

## Online WeasyPrint

http://weasyprint.org/

Simple website, uses online fonts (that's why it's slower now).

0.31:
- 1.6s
- 66MB memory peak

font-face:
- 2.2s (+38%)
- 68MB memory peak (+3%)

font-face + perf:
- 2.0s (+25%)
- 70MB memory peak (+6%)

## Online Wikipedia

https://en.wikipedia.org/w/index.php?title=HTML5&printable=yes

Printable version of a Wikipedia page, long left-aligned paragraphs with floats.

0.31:
- 7.9s
- 134MB memory peak

font-face:
- 8.2s (+4%)
- 124MB memory peak (-7%)

font-face + perf:
- 6.6s (-16%)
- 141MB memory peak (+5%)

Issue #70, opened 2013-04-04
Profile big tables


SimonSapin:
> mop___: finally found time for the performance test :S having one large table which spreads across 100 pages consumes LOADS of memory :S filesize is still 67kb :S
> mop___: 6gb memory and counting :D

http://pastebin.com/HLVrTmxt

liZe:
I can't reproduce this issue anymore. I've tried to render a 10000-line table, it took 108 seconds on my computer with about 3GB of RAM used (between 10MB and 15MB per page). I've just used 10000 instead of 100 in the script above. The output is a 239-page, 164kB PDF.

Of course, we could be better. Firefox takes about 2s and 100MB to render the same page.

liZe:
The example provided by @si13b still fails.

liZe:
Good news:
![perf](https://user-images.githubusercontent.com/952243/28490849-703f6a54-6ee4-11e7-821d-7ab8afd6d5cf.png)

- The red line is Python 3.5 before 344cb08.
- The blue line is Python 3.6 before 344cb08.
- The black line is Python 3.6 after 344cb08.

Python 3.6 is a huge improvement thanks to [compact dicts](https://mail.python.org/pipermail/python-dev/2016-September/146327.html).

344cb08 prevents style dicts to be copied each time a box is duplicated. I have to check that it doesn't break anything with the W3C suite, it may have introduced subtle bugs but I'm pretty confident.

Oh, and I think that the "40 minutes" from the last comment were not true :wink:.

xrmx:
@liZe amazing, thanks!

liZe:
The results are very good for other documents too. I've tested the examples of #384, they're both significantly faster and less memory consuming thanks to Python 3.6 and this commit.

There's room for more improvement though (before I close this issue). Inline boxes need really more memory than block boxes, I don't really know why. That's why you get memory problems with tables: there are lots of text lines in tables (at least one per cell).

liZe:
I've done my best to both add optimizations and clean the code, and I'm really happy with the result. I've done some benchmarks with Python 3.6 on Linux, for 4 different versions:
- 0.39,
- 344cb08 (first step with important optimizations),
- 033f473 (style dicts frozen for "security" purpose, but less optimized), and
- 84bdee1 (style dicts that can be modified, but modification only happens at "secure" moments).

I had also tested 0.31 **with Python 3.5** in #384.

[...]

PR #638, opened 2018-06-10
Possible optimizations - do not merge

chancyk:
I was profiling WeasyPrint and came across a few optimizations that were a 25% speed increase on my particular use case.

3210ea8 simply caches `Box.margin_height()` because it gets called thousands of times. This is the largest speedup though all three are respectable. Is this safe?

89e269a see if `hypothetical_position` is already greater than `shape.position_y` before performing attribute lookups.

71f7415 just assign locals instead of performing 9 different attribute lookups in the original list comprehension.

Thoughts?

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,38 @@
 WeasyPrint changelog
 ====================
 
+
+Versions 0.42.x will only get simple bug fixes backported from the master
+branch. New features, opitimizations and complex bug fixes will only be added
+to the 43+ versions that don't support Python 2 anymore.
+
+Do not rely on future versions, development on the 0.x branch may be stopped at
+any moment.
+
+
+Version 0.42.1
+--------------
+
+Released on 2018-02-01.
+
+Bug fixes:
+
+* `#566 <https://github.com/Kozea/WeasyPrint/issues/566>`_:
+  Don't crash when using @font-config.
+* `#567 <https://github.com/Kozea/WeasyPrint/issues/567>`_:
+  Fix text-indent with text-align: justify.
+* `#465 <https://github.com/Kozea/WeasyPrint/issues/465>`_:
+  Fix string(*, start).
+* `#562 <https://github.com/Kozea/WeasyPrint/issues/562>`_:
+  Handle named pages with pseudo-class.
+* `#507 <https://github.com/Kozea/WeasyPrint/issues/507>`_:
+  Fix running headers.
+* `#557 <https://github.com/Kozea/WeasyPrint/issues/557>`_:
+  Avoid infinite loops in inline_line_width.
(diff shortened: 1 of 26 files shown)


liZe:
@chancyk Thanks a lot for your optimizations! I know how speed improvements can sometimes be frustrating and I really appreciate when other people take the time to 

> To be safe the _cached_margin_height needs a reset to None when any of the 7 involved attributes is changed. Probably requires a crafty Box.__setattr__() -- not being a Python expert I fear this will affect the speed of code, heavily: 7 ifs to execute on every assignment of a Box attribute.

That's perfectly right! 3210ea8 makes WeasyPrint faster but probably breaks many use cases.

> The condition for the short circuit isnt hypothetical_position >= y but hypothetical_position >= y + h -- no chance to avoid calling margin_height().

It's logically true, but in this case `h` can't be negative, so the `hypothetical_position >= y` condition is right!

> But wow! 25% speedup by not directly accessing attributes -- didnt know that attribute lookup is that slow in Python. Good to know. There must be many (other, better?) places, especially in the layout code, where local variables could be used to avoid the lookups...

Actually, they're not that slow, it's WeasyPrint's fault, here's why (long story, you can skip):

---

There used to be a class called `StyleDict` to store CSS attributes and values. This dict-like class had a copy-on-write feature (for memory footprint reasons) and its items could be accessed as attributes (`style.width` instead of `style['width']`, for convenience reasons). Style dicts are used everywhere in WeasyPrint and optimizing them is a main way to make WeasyPrint faster.

Dicts in Python are used for everything and are optimized in each Python version (sometimes heavily, like in 3.6, see [#70](https://github.com/Kozea/WeasyPrint/issues/70#issuecomment-317180608)). A simple way to make WeasyPrint really faster was to replace the `StyleDict` by a real `dict`, we can get Python's improvements in WeasyPrint for free.

`StyleDict` has been transformed to inherit from `dict`, and copy-on-write was removed in #384. It gave amazing results for speed but cost extra memory. This change is already in 0.42.x versions.

That was not enough. Inheriting from `dict` was a great improvement, but pure `dict` objects are really better. ad11edf removed the `StyleDict` class and style items can't be accessed as attributes anymore. This big improvement is in the `master` branch, but not in 0.x versions.

---

I've cherry-picked 71f7415, mainly as it avoids calling `margin_height()` multiple times. As `StyleDict` objects have been removed, using local variables instead of attribute accesses are not required anymore (at least with the testing documents I've tried).

If you find other optimizations on `master`, I'll be happy to merge them!

PS: @chancyk I'd be really interested in your performance results using the current `master` branch with your documents. I'd be happy too to try your samples (if possible), and to know why you use WeasyPrint before releasing the stable version (see #635)!

PR #641, opened 2018-06-13
Use actual page as containing block for fixed boxes

### LINKED ISSUE #636: Incorrectly positioned static divs when using multiple page sizes
I have a PDF I'm generating that has the first page as A4 Portrait and the remaining pages are in landscape (dynamic number of pages depending on the amount of data in the report). The issue is that  the statically positioned header and footer elements are rendered in the incorrect place. The header is correct on the first page and the footer is correct on the second page. On the first page the footer shows at the height it would be if the page was landscape and the header on the second page shows half way across the page as if the page was portrait. Simplified example below.

@page:first {
    size: a4 portrait;
}
@page  {
    size: a4 landscape;
}
.page-footer {
    position: fixed;
    text-align: center;
    background-color: rgb(212, 218, 92);
    font-size: 0.8rem;
    line-height: 1.7rem;
    height: 1.7rem;
    bottom: -10mm;
    width: 100%;
}
#page-logo {
    width: 170px;
    position: fixed;
    right: 0;
    top: -22mm;
}

screen shot 2018-06-08 at 2 43 00 pm

Tontyna:
Fixes #636

### DIFF
--- a/weasyprint/layout/__init__.py
+++ b/weasyprint/layout/__init__.py
@@ -26,14 +26,16 @@ from .backgrounds import layout_backgrounds
 from ..formatting_structure import boxes
 
 
-def layout_fixed_boxes(context, pages):
-    """Lay out and yield the fixed boxes of ``pages``."""
+def layout_fixed_boxes(context, pages, containing_page):
+    """Lay out and yield the fixed boxes of ``pages`` on the
+    ``containing_page``.
+    """
     for page in pages:
         for box in page.fixed_boxes:
             # Use an empty list as last argument because the fixed boxes in the
             # fixed box has already been added to page.fixed_boxes, we don't
             # want to get them again
-            yield absolute_box_layout(context, box, page, [])
+            yield absolute_box_layout(context, box, containing_page, [])
 
 
 def layout_document(enable_hinting, style_for, get_image_from_uri, root_box,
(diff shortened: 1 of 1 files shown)


ElRoberto538:
This works great except for images which still appear in the wrong location, taking a quick look at the code now that I can see where it processes fixed boxes but not sure I can fix it.

Tontyna:
Indeed: TextBoxes are always copied -- `split_inline_level` calls `split_text_box` which calls `box.copy` -- thats why adada5b did the trick.

Images / InlineReplacedBoxes are passed through `atomic_box` which (until d46dd01) only returned the altered box. 
This commit moves the images in the logo and footer to their proper places.

Tontyna:
Yes, list-markers of `list-style-position: outside` need special treatment, too.
I'm almost shure that there are no more special beasts...

@liZe: I hope the copying I introduced has no weird side-effects. But I couldn't find a way to determine whether we are in a "fixed position".
Besides: The code in weasyprint/layout is **no fun to read**

Tontyna:
Hope, the `BlockReplacedBox` was the last box requiring special treatment.

liZe:
@Tontyna Thanks a lot.

I've put the copy of replaced boxes into `layout_fixed_boxes`, I think that it's better to keep it a single place. The only special case left is markers, we should copy it in `layout_fixed_boxes` too but that's too much useless work.

And I've added a test, of course :wink:.

PR #649, opened 2018-06-24
Fix some bugs related to content-lists

Tontyna:
Ignore CSS values like the following -- invalid value!

```css
#test1 {
   string-set: "some nameless text";
}
#test2 {
   string-set: mystring "some text", "more nameless text";
}
```

### DIFF
--- a/weasyprint/css/targets.py
+++ b/weasyprint/css/targets.py
@@ -64,7 +64,8 @@ class TargetCollector(object):
         if anchor_name:
             self.items.setdefault(anchor_name, TargetLookupItem())
 
-    def lookup_target(self, anchor_token, source_box, parse_again_function):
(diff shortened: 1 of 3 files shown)


Tontyna:
The `content_lists` of css content, bookmark-label and of each named string are distinct creatures wich are evaluated separately. When their target is pending they must be evaluated once more in `target_collector.check_pending_targets()`

E.g.
```css
#id {
  string-set: string1 "counter 1 is " target-counter('#anchor', counter_1),
              string2 "counter 2 is " target-counter('#anchor', counter_2),
}
```

If we only store one parse_again() per anchor and box, the pending `string2` is never parsed again.

liZe:
I had a bad time trying to understand 62bada0 but I get it now.

Tontyna:
... not enough comments in source :smirk:

PR #650, opened 2018-06-25
Prevent embedded file streams from being bytes

### LINKED ISSUE #644: I can attach files via command line api but not the python api?
Hi,

Using -a on the command line I can attach pdfs fine, when I try 

`<link rel="attachment" href="path/to/pdf.pdf"> `

tags in the html I get no attachments, using relative or absolute paths makes no difference. 

I've also tried passing an array of attachments in the python like so...

`doc.write_pdf(outputPath,1,attachments)`

where doc is returned from an html.render() call. No dice that way either. I think I'm missing something basic, but I've looked through the docs and the issues on here and I can't see why it fails. 

Any ideas?

Tontyna:
Should fix #644

### DIFF
--- a/weasyprint/pdf.py
+++ b/weasyprint/pdf.py
@@ -94,6 +94,8 @@ def _create_compressed_file_object(source):
         md5.update(data)
         pdf_file_object.stream += compress.compress(data).decode('latin-1')
     pdf_file_object.stream += compress.flush(zlib.Z_FINISH).decode('latin-1')
+    if isinstance(pdf_file_object.stream, bytes):
+        pdf_file_object.stream = str(pdf_file_object.stream)
     pdf_file_object.Params = PdfDict(
         CheckSum=PdfString('<{}>'.format(md5.hexdigest())), Size=size)
     return pdf_file_object


liZe:
As
- `pdf_file_object.stream` is initially an empty `str`
- `decode` always returns `str` objects
- adding `str` and `bytes` objects raises an exception,

I don't know how we could ever get a `bytes` object here!

Tontyna:
I don't know whether we ever get a `bytes` object from decode (zlib is built-in, cannot debug), it was just an idea. It was just an experiment. And by forcing the stream to `bytes` the embedded stream in the PDF was changed into `b'....'`, exactly what the buggy PDFs contain.

No exceptions here. Neither with `pdf_file_object.stream = str(pdf_file_object.stream)`  nor with `pdf_file_object.stream = bytes(pdf_file_object.stream, 'latin-1')`.

BTW: My pdfrw version is 0.4

liZe:
> I don't know whether we ever get a bytes object from decode (zlib is built-in, cannot debug), it was just an idea. It was just an experiment.

No problem :wink:.

The problem is in 0.42.3 but is already fixed in master. I've backported the fix into the 0.x branch (as written in #644) and I'll release a new 0.x version later.

Tontyna:
Ah, got it -- I (of course?) used the master branch when exploring the issue...

PR #662, opened 2018-08-01
Minor issues with build_sphinx

Tontyna:
After creating an editable environment with Sphinx and everything for WeasyPrint -- BTW: On Windows the `activate` command is located in `env/Scripts`, not in `env/bin`, will edit `hacking.rst` later  -- I tried to build the `docs` by issuing `python setup.py build_sphinx`. 

Got no htmls but *Theme error: sphinx_rtd_theme is no longer a hard dependency since version 1.4.0. Please install it manually.(pip install sphinx_rtd_theme)*.
Don't know how / in which config-or-what-file this can be fixed or whether it should be fixed at all...  problem solved by running `pip install`.

After that the expected htmls were created successfully, though with some WARNINGs. The 4 warnings about `:ref:`not finding its target are solved by 2ea0c6c.

### DIFF
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -66,7 +66,7 @@ point to http://weasyprint.org/news/ in PDF files.
 PDF bookmarks are also called outlines and are generally shown in a
 sidebar. Clicking on an entry scrolls the matching part of the document
 into view. By default all ``<h1>`` to ``<h6>`` titles generate bookmarks,
-but this can be controlled with CSS (see :ref:`bookmarks`.)
+but this can be controlled with CSS (see :ref:`Bookmarks <bookmarks>`.)
 
 Attachments are related files, embedded in the PDF itself. They can be
 specified through ``<link rel=attachment>`` elements to add resources globally
@@ -299,6 +299,8 @@ defining "new properties and values, so that authors may bring new techniques
 (running headers and footers, footnotes, leaders, bookmarks) to paged media".
 
 Two features from this module have been implemented in WeasyPrint.
+
+.. _bookmarks:
 
 The first feature is `PDF bookmarks`_.  Using the experimental_
 ``bookmark-level`` and ``bookmark-level`` properties, you can add
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -90,6 +90,8 @@ WeasyPrint! Otherwise, please copy the full error message and
        JPEG, GIF and others are not available.
 
 
+.. _linux:
+
 Linux
 -----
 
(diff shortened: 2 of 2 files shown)


Tontyna:
There is another WARNING about not finding the intersphinx inventory of cairocffi. This one can only be solved by cairocffi maintainers. But as far as I could see: In the whole doc no intersphinx-link to cairocffi. 

Since there are no intersphinx links at all to nothing, the whole `intersphinx_mapping` in `config.py` is obsolete, but maybe it's for future use...

Tontyna:
Was tempted to eliminate the last WARNING about *favicon file 'icon.ico' does not exist*. That's a lie, the icon lives in the `_static` subfolder. Which, according to the comments in `config.py` should be the right place.

In contrast, the current Sphinx doc  tells me that html_favicon path should be relative to the configuration directory. 

Sounds like they changed that recently and I better live with the warning -- the files from the `_static` folder are copied to `_build\html\_static`, anyway.

Tontyna:
...until today I had no idea what those `/env`and `/venv` entries in `.gitignore` are supposed to prevent...

liZe:
Thanks a lot for the fixes.

> Don't know how / in which config-or-what-file this can be fixed or whether it should be fixed at all... problem solved by running pip install.

As we host docs on ReadTheDocs, the style is already installed. As there's already a nice message, I think it's OK.

> After that the expected htmls were created successfully, though with some WARNINGs. The 4 warnings about :ref:not finding its target are solved by 2ea0c6c.

Thank you.

> There is another WARNING about not finding the intersphinx inventory of cairocffi. This one can only be solved by cairocffi maintainers. But as far as I could see: In the whole doc no intersphinx-link to cairocffi.

I'll check that.

> Sounds like they changed that recently and I better live with the warning -- the files from the _static folder are copied to _build\html\_static, anyway.

I'll check that too.

> ...until today I had no idea what those /env and /venv entries in .gitignore are supposed to prevent...

These are quite standard virtual environment folders, but I'll remove them, people should put their virtualenvs somewhere else.

Tontyna:
> but I'll remove them, people should put their virtualenvs somewhere else.

Dont!

If you do and someone confidentlly follows the instructions in [Hacking WeasyPrint](http://weasyprint.readthedocs.io/en/latest/hacking.html#hacking-weasyprint), sets up an environment where `build_sphinx` works because she wants to contribute to the documentation and wants to check her work before pushing...

I mean: people who know how to setup virtualenvs and Sphinx and everything wont need those instructions anyway and the .gitignore entries dont hurt.

Issue #589, opened 2018-03-14
Weasyprint not working on Windows 7 x64 / Python 3.6.2 x64 / GTK3 runtime x64


micsan13br:
Hello,

I tried to follow the installation instructions as carefully as possible (and repeated the process twice with python 3.6.4 and 3.6.2 without getting installation errors). I am running windows 64 bit, using python 64 bit and GTK 3 64 bit (gtk3-runtime-3.22.26-2017-11-15-ts-win64.exe).
Visual C++ Build Tools were already installed on the system previously as well as Visual Studio 2017.
I did run the line "pip install --upgrade setuptools" as instructed in Python's wiki.

WeasyPrint just prints multiple error messages when testing using the command:

python -m weasyprint http://weasyprint.org weasyprint.pdf

Traceback (most recent call last):
  File "C:\Users\sad1lo\AppData\Local\Programs\Python\Python36\lib\runpy.py", line 183, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "C:\Users\sad1lo\AppData\Local\Programs\Python\Python36\lib\runpy.py", line 142, in _get_module_details
    return _get_module_details(pkg_main_name, error)
  File "C:\Users\sad1lo\AppData\Local\Programs\Python\Python36\lib\runpy.py", line 109, in _get_module_details
    __import__(pkg_name)
  File "C:\Users\sad1lo\AppData\Local\Programs\Python\Python36\lib\site-packages\weasyprint\__init__.py", line 375, in <module>
    from .css import preprocess_stylesheet  # noqa
  File "C:\Users\sad1lo\AppData\Local\Programs\Python\Python36\lib\site-packages\weasyprint\css\__init__.py", line 29, in <module>
    from . import computed_values
  File "C:\Users\sad1lo\AppData\Local\Programs\Python\Python36\lib\site-packages\weasyprint\css\computed_values.py", line 16, in <module>
    from .. import text
  File "C:\Users\sad1lo\AppData\Local\Programs\Python\Python36\lib\site-packages\weasyprint\text.py", line 18, in <module>
    import cairocffi as cairo
  File "C:\Users\sad1lo\AppData\Local\Programs\Python\Python36\lib\site-packages\cairocffi\__init__.py", line 41, in <module>
    cairo = dlopen(ffi, 'cairo', 'cairo-2')
[...]

PR #665, opened 2018-08-06
Don't use pdfrw anymore

liZe:
pdfrw is a great piece of software, but we don't know PDF enough to debug the
problems we've met. It's safer to use the new cairo API and get back to manual
edition for attachments and bleed boxes.

We only have two regressions for now:
- some internal links are broken,
- PDF producer is not overwritten.

A mail has been sent to cairo's mailing-list about that:
https://lists.cairographics.org/archives/cairo/2018-August/028694.html

Fix #639, #615, fix #596, fix #565.

PR #668, opened 2018-08-12
Fix link

Tontyna:
Besides this little link I'd like to improve the installation instructions for Windows (as desired e.g. by [@liZe](https://github.com/Kozea/WeasyPrint/issues/589#issuecomment-373124636) but not sure how to do that.

The instructions for Win64 are, maybe a bit meager, but almost correct:
Install Python3, install Tom Schoonjans' GTK3 Runtime, install the appropriate MSVC,
install Weasyprint.
No need to reboot, but before `pip install WeasyPrint` do a `pip install --update setuptools`.

Problem is **Win32**.  No problem with Python, no problem with MSVC.
But. There isn't an up-to-date, ready-to-use GTK+ anymore. I know, WeasyPrint doesnt need GTK+ but it requires Pango, Cairo and GDK-PixBuf and GTK+ contains all the required stuff.

Spent a few hours to find out a smart way. But there isn't.

Pango instructs us to download its Windows binaries from a 404 gtk.org page, [gtk.org tells us ](https://www.gtk.org/download/windows.php) that GTK+ for Windows isn't meant for end-users anymore instead developers should use MSYS2 and deploy the libraries with their application. Fine.

So Win32 user have three choices:

1. Wait until someone creates and maintains a GTK+Project for Win32
2. Install msys2, install and maintain therein the `mingw-w64-i686-pango` and the `mingw-w64-i686-gdk-pixbuf2` packages and insert the MSYS2-mingw32-bin-Path into their PATH
3. Install GIMP (or Inkscape or another Win32 application that uses a current GTK+), keep it up-to-date and insert the appropriate folder into their PATH -- in GIMP it's the ./bin subfolder

That's how it is. Shall I try to re-write the WeasyPrint installation instructions accordingly? Which recipe is preferable? Msys2 is a good thing when you plan to become a developer, but otherwise a slight overkill...

BTW: Both, the msys2 and the GIMP solution work for Win64, too.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -81,7 +81,7 @@ WeasyPrint! Otherwise, please copy the full error message and
        you get incomplete SVG renderings, please read `#339
        <https://github.com/Kozea/WeasyPrint/issues/339>`_. If you get invalid
        PDF files, please read `#565
-       <https://github.com/Kozea/WeasyPrint/issues/565>`.
+       <https://github.com/Kozea/WeasyPrint/issues/565>`_.
 
 .. [#] pango ≥ 1.29.3 is required, but 1.38.0 is needed to handle `@font-face`
        CSS rules.


liZe:
Fun fact: the link had also been fixed in #665.

> That's how it is. Shall I try to re-write the WeasyPrint installation instructions accordingly?

I'd love that!

> Which recipe is preferable? Msys2 is a good thing when you plan to become a developer, but otherwise a slight overkill...

Two solutions:
- Msys2, or
- drop Windows Win32 support.

Next version will rely on pretty recent libraries, it would be coherent to drop Win32 just as we drop Python 2.

Tontyna:
> drop Windows Win32 support

Don't know how many people run WeasyPrint on Windows, dont know how many of them run Win32, but there exists even a 32bit Win10, Win 7 extended support ends in 2020.

Not really an option as long as Python3 supports Win 32bit?

liZe:
> Not really an option as long as Python3 supports Win 32bit?

Python 2 is supported until 2020, it doesn't prevent us from dropping support right now.

Official Win32 support doesn't only mean that WeasyPrint works with Win32. It means that we are responsible for maintaining an install documentation page that is clear enough for users so that they don't have to ask us as soon as they have a problem. I'm OK with msys2 as long as we don't get issues about how to install/use/config it :wink:.

Tontyna:
No problem dropping Python 2 -- but Python **3** has installers for / supports Win32.

We'll always get issues with the stuff WeasyPrint depends on but cant control. 

Contemplating a `weasyprint --selfcheck` command...

liZe:
> No problem dropping Python 2 -- but Python 3 has installers for / supports Win32.

I understand. The point is that we don't have to support every platform that's supported by other companies or projects. Not supporting Win32 for WeasyPrint is an option even if Python3 installers support Win32.

> We'll always get issues with the stuff WeasyPrint depends on but cant control.

It's not a fatality.

We've got almost no complains about Linux users, and I'm convinced that it's because our Linux documentation is much better that the Windows documentation. As I'm a Linux user, I know what to expect from the Linux documentation, it's easier to write the right documentation.

We could say that Win32 users are on their own, just as we do for [Macports users](http://weasyprint.readthedocs.io/en/latest/install.html#macos). That's what I'd do with other OSes like *BSD or Solaris for example: it doesn't mean that it won't work, it means that we don't have to help them.

(And to be honest, Windows and macOS users ask for an installer.)

Tontyna:
Despite disputing OSses,  I'm happy that Python, GTK+ and WeasyPrint can be convinced to run on this proprietary system called Windows.

When looking for a pagination/PDF tool for HTML I was told that not only kittens will die when I try to setup WeasyPrint on Windows, but that one needs superhuman powers to make it work unless one is a Linux expert, and every sane Windows person should decide for Prince, Prince, Prince. And moreover I was told that Weasyprint is buggy and not maintained and a bad choice, yeah, that's what I was told. HA!

I'm cocksure it will do the average Windows user -- whatever that is -- no harm when he starts using his shell and brain and learns something new. Knowledge is power!

Tontyna:
BTW: What's the recommended Python version? The latest? Python 3? 3.6 or newer?

liZe:
> I'm cocksure it will do the average Windows user -- whatever that is -- no harm when he starts using his shell and brain and learns something new. Knowledge is power!

:smile:

> BTW: What's the recommended Python version? The latest? Python 3? 3.6 or newer?

The latest.

Tontyna:
Good news: No Visual C++ Tools (aka Windows compiler) required to install WeasyPrint on Windows! CFFI ist the only required package in need of a .pyd and it's distributed as wheel and wheels dont require the C compiler. 
Verified by successfully installing WeasyPrint on systems without MSVC.

PR #674, opened 2018-08-21
Extensive installation instructions for Windows

Tontyna:
Really extensive, doubles the size of install.rst -- tried not to explain everything from big bang on, but wordy is my second name, delete superfluous babbling as you like.

There is one issue: In `.. code-block:: sh`, backslash is combined with the next character into a blue rendered `<span class="se">`. Whats the proper markup for Windows command prompt / Windows paths?

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -179,35 +179,256 @@ Windows
 
 Dear Windows user, please follow these steps carefully.
 
-Really carefully. Don't cheat.
+Really carefully. Don’t cheat.
 
-**If you decide to install Python or GTK 32 bit on Windows 64 bit, you're on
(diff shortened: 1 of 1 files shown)


liZe:
:heart:

Tontyna:
I see: `.. code-block:: bat` -- should be the proper lexer, but still applies crazy highlights.
`.. code-block:: console` seems to do the trick, `<span class="go">`, no funny colors for special chars and words at all.

liZe:
> I see: .. code-block:: bat -- should be the proper lexer, but still applies crazy highlights.
> .. code-block:: console seems to do the trick, <span class="go">, no funny colors for special chars and words at all.

Then let's go! I was quite happy with `bat`, but the `C:/` part was strange.

PR #682, opened 2018-09-03
Added section to tutorial on adjusting `@page` attributes

rightbrace:
This adds a section to the tutorial page in the docs which explains:

- There are no command line flags to set page size/margin
- How to do this in CSS
- Links to the MDN page on the `@page` at-rule
- Links to weasyprint.org main page, which demonstrates adding page numbers and headers to the output file

This may be controversial because it is considered outside the scope of the documentation (as it is a CSS feature), however as an end user (especially one who has little experience with writing print-friendly HTML), I believe that at demonstrating at least those two options is incredibly useful. 

Further, an example using the `@page` rule _is_ on the main page, however it does not show how to change page size, which I believe to be important because the default size of A4 is not as common in North America.

This question is what prompted this addition to the docs:
https://stackoverflow.com/questions/29831775/weasyprint-page-size-wrong-8-27in-x-11-69-in

Postscript:
This should have been one commit not two - I did not initially realize that the docs/_build directory is omitted from the repository

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -28,6 +28,28 @@ If you have many documents to convert you may prefer using the Python API
 in long-lived processes to avoid paying the start-up costs every time.
 
 
+Adjusting Document Dimensions
+.............................
+
+Currently, WeasyPrint does not provide support for adjusting page size
+or document margins via command-line flags. This is best accomplished
+with the CSS ``@page`` at-rule. Consider the following example:
+
+.. code-block:: css
+  
+  @page {
+    size: Letter; /* Change from the default size of A4 */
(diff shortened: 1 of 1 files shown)


liZe:
Thanks a lot!

> This may be controversial because it is considered outside the scope of the documentation (as it is a CSS feature), however as an end user (especially one who has little experience with writing print-friendly HTML), I believe that at demonstrating at least those two options is incredibly useful.

You're right. Actually, we're currently working on the new website with a lot of samples, and we want to use them as tutorials. There are so many paged-media-related features in CSS that tutorials would be really useful.

I keep your section for now, I'll probably move it to a dedicated tutorial page as soon as we have a better idea about what we want!

PR #762, opened 2019-01-05
Ship LICENSE in sdist/bist, document minimum setuptools version

kitterma:
Since the code says:
    :copyright: Copyright 2011-2018 Simon Sapin and contributors, see AUTHORS.
    :license: BSD, see LICENSE for details.
It is more complete/clearer if the referenced files are included in the dists.  This is particularly true for the BSD license, since there are many variants of the BSD license.

Additionally, it took me awhile to figure out what version of setuptools I needed.  I was finally able to figure it out from the meta data section of https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files.  I think it would be helpful to document the version requirements for people that can't just upgrade to the latest.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,2 @@
+include LICENSE
+include AUTHORS
--- a/README.rst
+++ b/README.rst
@@ -17,3 +17,5 @@ pagination, and meant to be easy to hack on.
 * Source code and issue tracker: https://github.com/Kozea/WeasyPrint
 * Tests: https://travis-ci.org/Kozea/WeasyPrint
 * Support: https://www.patreon.com/kozea
+
+Minimum setuptools version required is 39.2.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,6 +34,7 @@ classifiers =
 [options]
 packages = find:
 zip_safe = false
+include_package_data=True
 setup_requires = pytest-runner
 install_requires =
   cffi>=0.6


liZe:

> It is more complete/clearer if the referenced files are included in the dists. This is particularly true for the BSD license, since there are many variants of the BSD license.

You're right. I'd love to find a solution to include it without adding a `MANIFEST.in` file, I've spent a lot of time cleaning the repository and only keeping `setup.cfg` and `setup.py` :smile:.

> Additionally, it took me awhile to figure out what version of setuptools I needed.

It was at the very beginning of the [Installing](https://weasyprint.readthedocs.io/en/latest/install.html) page.

kitterma:
I looked at the installing page and missed it.  Thanks.  Because you use the 'version' key, the minimum setuptools requirements needs to be bumped to 39.2.0 (see the metadata section of the page I linked in the original pull request.

I could tell you were trying to avoid having a MANIFEST.in, so I tried to avoid it, but didn't come up with something that works, but looking at the documentation again, I think maybe I see what I was doing wrong.  I'll see if I can figure it out and submit a new pull request.

Thanks for the quick review.

liZe:
> I looked at the installing page and missed it. Thanks. Because you use the 'version' key, the minimum setuptools requirements needs to be bumped to 39.2.0 (see the metadata section of the page I linked in the original pull request.

I think that 39.2.0 is only needed to generate the package, not to install it. I've tried with 39.0.1 installed on my system, it seems to work well.

> I'll see if I can figure it out and submit a new pull request.

I'll try to find a solution too…

kitterma:
OK.  Perhaps a separate note on the setuptools build requirement being a higher version.

kitterma:
setuptools-40.8.0 was just released and it support license_file in setup.cfg, so this can be solved now.

Issue #699, opened 2018-09-25
WeasyPrint install instructions for Alpine Linux


elyak123:
Just wanted to recommend adding to documentation instructions to install WeasyPrint in Alpine Linux 3.6. It took me a while but I made it work with this command:
`apk --update --upgrade add gcc musl-dev jpeg-dev zlib-dev libffi-dev cairo-dev pango-dev gdk-pixbuf`

If someone else have a more efficient aproach (important when working with docker containers for small size images) can comment on this.

VDShark:
Also spent some time for finding non redundant solution.

As a result of experiments, found next
1) these packages required for running `cairo pango gdk-pixbuf`
2) these packages required only on the building stage `gcc musl-dev libffi-dev jpeg-dev zlib-dev` (and could be deleted after that for keeping image as small/clean as possible )

VDShark:
Finally!
With the latest versions of alpine & weasyprint we can keep build pretty small & clean.

Here my current approach:
```
FROM python:3.7-alpine
ENV PYTHONUNBUFFERED 1

RUN apk --update --upgrade --no-cache add \
    cairo-dev pango-dev gdk-pixbuf

...

ADD /etc/pip/requirements.txt /code/requirements.txt
RUN set -ex \
    && apk add --no-cache --virtual .build-deps \
        musl-dev gcc jpeg-dev zlib-dev libffi-dev \
    && pip install --no-cache-dir -r /code/requirements.txt \
    && apk del .build-deps

...
```

Packages `cairo-dev pango-dev gdk-pixbuf` need in runtime, so we kept them.
Other ones need only on building stage - so we clean.

In requirements.txt just weasyprint and my specific dependencies, not related to this issue.

dploeger:
For somebody who came around this issue because they couldn't install weasyprint on Alpine now, here's the two lines to get it installed:

    apk add python3 py3-pip pango gcc musl-dev py3-wheel python3-dev libffi-dev zlib-dev jpeg-dev
    pip install weasyprint

dploeger:
And this is also [documented](https://weasyprint.readthedocs.io/en/stable/install.html#alpine) which I somehow overread 🙄

jo-bitsch:
I recently used the following commands for my docker containers (`alpine 3.17`):

```
apk add python3 py3-pip pango
pip install weasyprint
weasyprint -h
```

[...]

PR #764, opened 2019-01-07
gdk-pixbuf Error on Alpine 3.8

Ballrock:
Hi,

Got some problems with gdk-pixbuf on Alpine 3.8.
@elyak123's command at #699 seems ok at beggining but after some builds (without images), i finally got some error on JPEG images, like : 
```bash
ERROR: Failed to load image at "file:///builds/1.jpg" (Could not load GDK-Pixbuf. PNG and SVG are the only image formats available.)
ERROR: Failed to load image at "file:///builds/2.jpg" (Could not load GDK-Pixbuf. PNG and SVG are the only image formats available.)
[...]
```

I try some extra packages without success. Eventually, looks like add the devel version of gdk-pixbuf fix the problem.

Unfortunatly i didn't try on other Alpine version but since `gdk-pixbuf-dev` include `gdk-pixbuf` it does not seem too risky.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -159,7 +159,7 @@ For Alpine Linux 3.6 or newer:
 
 .. code-block:: sh
 
-    apk --update --upgrade add gcc musl-dev jpeg-dev zlib-dev libffi-dev cairo-dev pango-dev gdk-pixbuf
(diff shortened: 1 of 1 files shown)


liZe:
Looks safe… Thank you!

PR #784, opened 2019-02-01
note on alpine installation and ctypes.util

cinatic:
fix for broken ldconfig / ctypes.util.find_library as mentioned in:

https://github.com/docker-library/python/issues/111#issuecomment-437625964
https://github.com/Kozea/WeasyPrint/issues/699#issuecomment-455180488

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -162,6 +162,21 @@ For Alpine Linux 3.6 or newer:
     apk --update --upgrade add gcc musl-dev jpeg-dev zlib-dev libffi-dev cairo-dev pango-dev gdk-pixbuf-dev
 
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


liZe:
Thank you.

liZe:
@cinatic Does adding `'libcairo.so.2'` as last argument of cairocffi's `dlopen` fix the problem? If so, I should add it as it's pretty harmless and may help people.

cinatic:
the problem is that the find_library does not work, i tried it also with alpine3.9 this still does not work:

https://github.com/Kozea/cairocffi/blob/master/cairocffi/__init__.py#L31

maybe it would be worthy to fallback trying to load libcairo.so.2 (or more) directly if all find_library tries failed. i can provide a PR if you like

PR #789, opened 2019-02-05
Add LICENSE to sdist

kitterma:
It turned out it wasn't possible to add the LICENSE using setup.cfg, but it is now.  This PR implements what we discussed a few months ago (now that it's possible).  This is safe even though it uses a new feature because older setuptools will just not install the file.  I didn't get any errors when I tested it.


liZe:
Cool, thank you!

PR #792, opened 2019-02-08
Document no alignment character support

Smylers:
Make it clear that text-align: '.' and similar (in table cells) isn't supported.


liZe:
Thank you!

PR #796, opened 2019-02-09
Fix a small typo in the tutorial

PaperBag42:
While reading the tutorial, I've noticed it says "a few random example" instead of "examples".


liZe:
Thank you!

PR #804, opened 2019-02-15
refs Kozea#412 Avoid default Logger

xusy2k:
If you want to use Weasyprint under django, you can configure
weasyprint's logger via django, but this setting never is set

### DIFF
--- a/weasyprint/logger.py
+++ b/weasyprint/logger.py
@@ -23,7 +23,8 @@
 import logging
 
 LOGGER = logging.getLogger('weasyprint')
-LOGGER.setLevel(logging.WARNING)
-LOGGER.addHandler(logging.NullHandler())
+if not LOGGER.handlers:
+    LOGGER.setLevel(logging.WARNING)
+    LOGGER.addHandler(logging.NullHandler())
 
 PROGRESS_LOGGER = logging.getLogger('weasyprint.progress')


liZe:

I'm not sure to fully understand how this can a problem with Django. Could you please provide a simple example showing the problem you're fixing?

xusy2k:
Yes: Django allows to configure logger. This is mine:

```python
# settings.LOGGING:
LOGGING = {
 'disable_existing_loggers': False,
 'formatters': {'verbose': {'datefmt': '%Y%m%d %H:%M:%S',
                            'format': '[%(asctime)s] %(levelname)s '
                                      '[%(name)s:%(lineno)s] %(message)s'}},
 'handlers': {'console': {'class': 'logging.StreamHandler',
                          'formatter': 'verbose',
                          'level': 'DEBUG'},
              'file': {'class': 'logging.FileHandler',
                       'filename': '/tmp/logging.log',
                       'formatter': 'verbose',
                       'level': 'DEBUG'}},
 'loggers': {
             # ...
             'weasyprint': {'handlers': ['console'],
                            'level': 'ERROR',
                            'propagate': True},
             ,
            }
 'root': {'handlers': ['file'], 'level': 'DEBUG'},
 'version': 1
}
```

As you can see, I set default logLevel for `weasyprint` as ERROR, but this has no effect because the package set always logLevel to WARNING ignoring the django's setting.

In my case I have thousands of warnings like `Error: Expected <ident> for declaration name, got literal. at 9:91136` taken from CSS (tested with Bootstrap 2.3.2). Also, I can't modify bootstrap.css, but I want to mute this warnings configuring the logLevel as ERROR

I hope this help you :-) and Thank you for this awesome package!

liZe:
> As you can see, I set default logLevel for `weasyprint` as ERROR, but this has no effect because the package set always logLevel to WARNING ignoring the django's setting.

OK, I can understand now. Thanks a lot for the example.

> I hope this help you :-) and Thank you for this awesome package!

:heart:

PR #805, opened 2019-02-15
Addition of standard print media sizes

rmackinnon:
css/computed_valued.py: Added dimensions for output to defined common book sizes (mass market paperback/novella/digest/us_tradeback, textbook small/medium/fullsize).

### DIFF
--- a/weasyprint/css/computed_values.py
+++ b/weasyprint/css/computed_values.py
@@ -112,6 +112,34 @@ PAGE_SIZES = dict(
         Dimension(11, 'in'),
         Dimension(17, 'in'),
     ),
+    paperback_massmarket=(
+        Dimension(4.25, 'in'),
+        Dimension(6.87, 'in'),
+    ),
+    paperback_novella=(
+        Dimension(5, 'in'),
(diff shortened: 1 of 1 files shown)


rmackinnon:
Looks like the build is failing on python 3.6 tests with the cairocffi version, but all the other versions pass.

liZe:
Hello @rmackinnon,

Thanks a lot for taking the time to create a pull request!

WeasyPrint tries to blindly follow the specifications, and unfortunately these book sizes are not included in https://drafts.csswg.org/css-page-3/#typedef-page-size-page-size, and not even in [IEEE ISTO Printer Working Group's reference document](ftp://ftp.pwg.org/pub/pwg/candidates/cs-pwgmsn10-20020226-5101.1.pdf).

I know that this PR may look harmless (and they always are … at the beginning :smile:), but each time I've tried to include this kind of changes, I've met problems afterwards. So please forgive me for being the bad guy, but … I'll be the bad guy. I hope you won't be mad at me, but as it's not a blocking problem it shouldn't prevent you from using WeasyPrint :pray:.

PR #825, opened 2019-03-14
Use correct shrink values, keep width and height positive

Tontyna:
Upps, copy & paste is one of the roots of evil.


liZe:
Oh, sorry, I had fixed that yesterday but not committed yet.

Tontyna:
No problem. When I saw my mistake I was astonished that it had escaped your eagle eyes.

PR #847, opened 2019-04-04
allow ability to ignore ssl verification within default_url_fetcher

mohdnajdy:
This fix is very helpful if i want to have images in html that are coming from un-trusted host.
This is applicable when i have a load balancer for all test servers and their global CA is not signed.

### DIFF
--- a/weasyprint/urls.py
+++ b/weasyprint/urls.py
@@ -202,7 +202,7 @@ HTTP_HEADERS = {
 }
 
 
-def default_url_fetcher(url, timeout=10, verify=True):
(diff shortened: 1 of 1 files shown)


mohdnajdy:
@liZe

liZe:
Hello,

Having this feature is interesting, but I'd prefer a `ssl_context` parameter (with a default value of `None`, directly passed to `urlopen`) instead of `verify`. It's more explicit and gives users the possibility to tweak the SSL connection to fit their needs.

mohdnajdy:
Done @liZe

liZe:
Thank you!

Issue #571, opened 2018-02-10
Support full text-decoration syntax


mdekstrand:
WeasyPrint does not currently support the full `text-decoration` syntax (`text-decoration-line`, `text-decoration-style`, `text-decoration-color`, and `text-decoration` as a shorthand for them).

These are in the CSS Text Decoration Level 3 CR, and are [documented by Mozilla](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration) and supported by Firefox, Chrome, and (mostly) Safari.

randomstuff:
I started a branch to try to add support for this.

mdekstrand:
@randomstuff fantastic! If you want a real-world test case, this document is the reason I filed this issue: https://md.ekstrandom.net/cv

liZe:
> @randomstuff fantastic! If you want a real-world test case, this document is the reason I filed this issue: https://md.ekstrandom.net/cv

Looks like [it works](https://github.com/Kozea/WeasyPrint/files/3178845/cv.pdf)!

As @randomstuff said in #855, we have small limitations (no wave, glitches with nested tags) but it should fit your needs!

PR #855, opened 2019-04-23
Add support for full CSS3 text-decoration [RFC]

randomstuff:
* parse text-decoration as shorthand
* draw different colors
* basic text-decoration-style
* no 'wavy' support for now

One problem of the way it's currently done is that it makes weird glitches with dashed/dotted style when elements are found in the decorated text:

```html
<p style="text-decoration: underline green dashed">Lorem <span>ipsum</span>
dolor sit amet, <span>consectetur</span> adipiscing elit, sed do eiusmod tempor 
ncididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud
exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
Duis aute irure dolor in reprehenderit in voluptate velit esse cillum
dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt
in culpa qui officia deserunt mollit anim id est laborum.</p>
```


randomstuff:
This is trying to address #571.

PR #859, opened 2019-04-27
Implement box-shadow with Azure Blur

Afoucaul:
This PR is an attempt to solve #13.

I merged and updated what @liZe did back in 2013 in #149.
I also wrote the property rule to support `box-shadow`.

While I understand the points raised in #13 and #149, I strongly believe that it's better to have this implementation, improvable as it may be, rather than nothing.
Furthermore, @liZe did some awesome work, even though it's been deemed not so efficient.
Now that it's been integrated, we can consider improving the implementation, but at least, we can generate basic shadows.

Here's an example of how shadows are rendered:

```html
<style>
div {
    width: 100px;
    height: 100px;
    background-color: green;
    box-shadow: 10px 10px 5px 5px black;
}
</style>
<div></div>
```

Screenshot from 2019-04-27 12-56-28

Not that bad looking IMO, and can be improved with a little CSS tweaking.


Afoucaul:
So... Based on @SimonSapin's suggestion in #13, and his work on providing Python bindings for [AzureBlur](https://github.com/SimonSapin/azureblur), I implemented the box shadow drawing using `cairo` and `azureblur`. It looks pretty awesome, to be honest.

This is the new shadow, using the same code as above:

Screenshot from 2019-04-27 16-46-06

I still have to work on the `inset` shadows.
What do you guys think about it?

liZe:
I'm closing this PR for now (see #13).

Issue #648, opened 2018-06-22
bookmark-state


mikosullivan:
This is a feature request. It would be cool if weasyprint recognized the bookmark-state CSS property: https://www.w3.org/TR/css-gcpm-3/#bookmark-state

Currently it appears that bookmark-state isn't a recognized property:

    WARNING: Ignored `bookmark-state: closed` at 121:2, unknown property.

Thanks for all the great work on this tool.

PR #870, opened 2019-05-17
Implement bookmark-state support

randomstuff:
Addresses #648


randomstuff:
Tests are currently broken. I think this is because the bookmarks are now all opened by default (as per the `bookmark-state` spec). I'm looking at it.

randomstuff:
I fixed the tests. The notable ugly hack is the `BookmarkSubtree` which tries to be API-compatible with the tuples which where used before.

liZe:
Thanks a lot! The code is clean and readable, it's a really nice feature to have. :heart:

liZe:
For the record: I've cleaned a couple of things (mainly removing the tuple workaround) in 2ce0309.

PR #875, opened 2019-05-31
Feature object-fit and object-position

### LINKED ISSUE #669: Support for object-fit css property
Hi:

I'm using WeasyPrint for generate PDF and the css property object-fit is not supported yet.

Do you have a plan to support this feature as this is really common used in my project.

Thanks.

mbarkhau:
Fixes #669


liZe:
See #669 for further discussion.

Issue #719, opened 2018-10-25
Support for column-span


fabiobatalha:
Is any kind of envisioning to give support for column-span.

Example of use of column-span using the tool PagedJS:

![screenshot from 2018-10-25 10-23-24](https://user-images.githubusercontent.com/504142/47507475-3dbb6d00-d840-11e8-930d-1f46d1f8b741.png)

The same PDF file produced with WeasyPrint using the same HTML and CSS's.

![screenshot from 2018-10-25 10-23-50](https://user-images.githubusercontent.com/504142/47507528-562b8780-d840-11e8-974a-025e1c3d2234.png)

liZe:
> Is any kind of envisioning to give support for column-span.

Nothing has been done yet to support column-span, but it would be quite easy to handle simple cases (ie at least without fixed height).

The overall goal is to change the `columns_layout` function (in `layout/columns.py`) to split the `box` at the beginning of the function into multiple boxes. For example:

```css
div {columns: 2}
h2 {column-span: all}
```

```html
<div>abc def <h2>test</h2> ghi jkl</div>
```

In this case, we have to:
- go through `box` (the `div` element) to find the `h2` element,
- render `<div>abc def ` (with the current code) and `<h2>test</h2>` (with some added code),
- render ` ghi jkl</div>` for example by calling `columns_layout` with the current `skip_stack`.

If anyone is interested in implementing this feature, I'll be happy to help!

PR #878, opened 2019-06-04
Add partial support for column-span property.

grewn0uille:
Related to #719.

### DIFF
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -559,7 +559,7 @@ are supported.
 The ``break-before``, ``break-after`` and ``break-inside`` properties are
 **not** supported.
 
-The ``column-span`` property is **not** supported.
+The ``column-span`` property is supported for direct children of columns.
 
 The ``column-fill`` property is supported, with a column balancing algorithm
 that should be efficient with simple cases.
--- a/weasyprint/css/validation/properties.py
+++ b/weasyprint/css/validation/properties.py
@@ -411,8 +411,7 @@ def column_width(token):
 @single_keyword
 def column_span(keyword):
     """``column-span`` property validation."""
-    # TODO: uncomment this when it is supported
-    # return keyword in ('all', 'none')
+    return keyword in ('all', 'none')
 
 
 @property()
--- a/weasyprint/layout/columns.py
+++ b/weasyprint/layout/columns.py
@@ -21,13 +21,15 @@ def columns_layout(context, box, max_position_y, skip_stack, containing_block,
                    adjoining_margins):
     """Lay out a multi-column ``box``."""
     # Avoid circular imports
-    from .blocks import block_box_layout, block_level_width
+    from .blocks import (
+        block_box_layout, block_level_layout, block_level_width,
+        collapse_margin)
 
     # Implementation of the multi-column pseudo-algorithm:
     # https://www.w3.org/TR/css3-multicol/#pseudo-algorithm
-    count = None
     width = None
     style = box.style
+    original_max_position_y = max_position_y
 
     if box.style['position'] == 'relative':
         # New containing block, use a new absolute list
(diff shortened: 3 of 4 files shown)


JohannesMunk:
Hey liZe!

This new balancing algorithm is the cause of quite a big slowdown in document generation in our projects. We have several documents with colspans>=2 that span multiple pages and contain complex block stuff. Rendering all columns content on every pixel the height is increased was getting to slow. 90mins for 230 pages.. 

While keeping the basic idea alive instead of a linear scan to find the smallest columns height for a "balanced" layout I implemented a simple binary search.

```python
    # Balance.
    #
    # The current algorithm starts from the ideal height (the total height
    # divided by the number of columns). We then iterate until the last column
    # is not the highest one. At the end of each loop, we add the minimal
    # height needed to make one direct child at the top of one column go to the
    # end of the previous column.
    #
    # We rely on a real rendering for each loop, and with a scanning algorithm
    # like this it can last minutes…

    adjoining_margins = []
    current_position_y = box.content_box_y()
    new_children = []
    for column_children_or_block in columns_and_blocks:
        if not isinstance(column_children_or_block, list):
            # We get a spanning block, we display it like other blocks.
            block = column_children_or_block
            resolve_percentages(block, containing_block)
            block.position_x = box.content_box_x()
            block.position_y = current_position_y
            new_child, _, _, adjoining_margins, _ = block_level_layout(
                context, block, original_max_position_y, skip_stack,
                containing_block, page_is_empty, absolute_boxes, fixed_boxes,
                adjoining_margins)
            new_children.append(new_child)
            current_position_y = (
                new_child.border_height() + new_child.border_box_y())
            adjoining_margins.append(new_child.margin_bottom)
            continue

        excluded_shapes = context.excluded_shapes[:]

        # We have a list of children that we have to balance between columns.
        column_children = column_children_or_block

        # Find the total height of the content
        current_position_y += collapse_margin(adjoining_margins)
        adjoining_margins = []
        column_box = create_column_box(column_children)
        new_child, _, _, _, _ = block_box_layout(
            context, column_box, float('inf'), skip_stack, containing_block,
            page_is_empty, [], [], [])
        height = new_child.margin_height()
        total_height = height
        if style['column_fill'] == 'balance':
            height /= count

        # Try to render columns until the content fits, increase the column
        # height step by step by binary search limits.

        # helper to derive the step_size
        def step_size_calc(value):
            return max(1, value)
            # coercing to power of 2 has no benefit, but increases number of steps..
            # result = round(value).bit_length()
            # return 1 if result == 0 else 1 << result-1

        # difference between upper and lower height is the current search step size
        # start value is arbitrary height / 4
        step_size = step_size_calc((height)/4)
        lower_height = height
        upper_height = 0
        # while the right height has not been found
        while True:
            column_skip_stack = skip_stack
            # Remove extra excluded shapes introduced during previous loop
            new_excluded_shapes = (
                len(context.excluded_shapes) - len(excluded_shapes))
            for i in range(new_excluded_shapes):
                context.excluded_shapes.pop()

            # through all columns
            for i in range(count):
                # render the column.
                new_box, resume_at, next_page, _, _ = block_box_layout(
                    context, column_box, box.content_box_y() + height,
                    column_skip_stack, containing_block, page_is_empty,
                    [], [], [])

                # did we render anything?
                if new_box is None:
                    # Give up and use the total content height.
                    height = total_height
                    break

                # render of the next column shall pickup where this one left
                column_skip_stack = resume_at

                # Stop if we already rendered the whole content
                if resume_at is None:
                    break

            # gave up or reached the total_height of all content combined?
            if height >= total_height:
                break

            # rendered the whole content?
            if column_skip_stack is None:
                if step_size == 1:
                    break
                else:
                    # found a working height, but our binary step is still to big.
                    # need to continue and check if height is lowest acceptable
                    upper_height = height
                    # we need to step backwards until our stepsize is 1.
                    step_size = step_size_calc((upper_height-lower_height)/2)
                    height -= step_size
            else:
                # overall height was not enough to fit the whole content =>

                # height value itself is not acceptable - height+1 might be
                lower_height = height+1
                
                # if a real upper limit is not known. we do not half the stepsize
                # Increase the columns heights and render them once again
                if not upper_height == 0:
                    step_size = step_size_calc((upper_height-lower_height)/2)
                    height = min(height+step_size, upper_height)
                    # optimization: avoid last block_box_layout if we already know that our new height is acceptable
                    if step_size == 1 and height == upper_height:
                        break
                else:
                    height = height+step_size
                

```

My question is: Why is the height information of each child - as is used to determine the total content height - not enough to fill the children into columns as they fit? Why call block_box_layout in every iteration of the height scan?

Hope your pdf-backend migration is progressing well!

Johannes

PR #882, opened 2019-06-17
[ADD] position: running(); content: element();

hbrunn:
@liZe this is a result of our discussion in https://github.com/OCA/reporting-engine/issues/254

This doesn't work yet, but I've a few questions so I figure it's a good idea to open this draft PR in order to talk about code


weasyprint/css/utils.py
```diff
@@ -740,3 +740,7 @@ def get_content_list_token(token, base_url):
         elif arg.type == 'string':
             string = arg.value
         return ('leader()', ('string', string))
+    elif name == 'element':
+        if len(args) != 1 or args[0].type != 'ident':
+            return
+        return ('element()', ('element', args[0]))
```
hbrunn: the second element of the tpuble could be just `args[0]` I think

weasyprint/layout/__init__.py
```diff
@@ -209,6 +209,9 @@ def __init__(self, enable_hinting, style_for, get_image_from_uri,
         self.tables = {}
         self.dictionaries = {}
 
+        # TODO: this is probably the wrong place to put this
+        self.running_elements = {}
```
hbrunn: we'll have to carry around the running elements somewhere. Do you think this belongs here or should I extend the state tuple?

hbrunn:
generally, my problem is that in https://github.com/Kozea/WeasyPrint/blob/master/weasyprint/layout/pages.py#L348-L349, we assume inline boxes, which crashes in https://github.com/Kozea/WeasyPrint/blob/master/weasyprint/formatting_structure/build.py#L1247. What would be the smartest way to change this? (I just started studying this library, so I'm not yet well versed with available helper functions)

weasyprint/css/utils.py
liZe: It can, as `element()` only accepts idents.

weasyprint/layout/__init__.py
liZe: I think that the specification isn't clear enough. It's hard to know which running element you'll have in the page margin when multiple running elements are in the page. The `string()` function at least has a second argument for that.

Even a sigle value, in [example 5](https://drafts.csswg.org/css-gcpm/#running-syntax) for example, is not clear to me: what's on the second page?
- We could put '2 / Miranda v. Arizona in Context'.
- We could put '1 / Miranda v. Arizona in Context', as the running element is defined on page 1.
- We could put nothing, as no running element is defined on page 2.
(I think that we want the first answer.)

The place where to put running elements depends on the answer to this question. But if we think that running elements somehow behave like named strings, then it's the right place to set

liZe:
> generally, my problem is that in https://github.com/Kozea/WeasyPrint/blob/master/weasyprint/layout/pages.py#L348-L349, we assume inline boxes, which crashes in https://github.com/Kozea/WeasyPrint/blob/master/weasyprint/formatting_structure/build.py#L1247. What would be the smartest way to change this? (I just started studying this library, so I'm not yet well versed with available helper functions)

For now, you can call the functions that are in `build_formatting_structure`, as we now may have blocks:
https://github.com/Kozea/WeasyPrint/blob/e080d97ee8651aa2edb7ccbb2b418b46ba0008a2/weasyprint/formatting_structure/build.py#L74-L80

We may want to create a function for this later.

weasyprint/layout/__init__.py
hbrunn: If it's not the first answer I wouldn't want to implement it because then it wouldn't be useful for me :-)
I define a [box](https://github.com/Kozea/WeasyPrint/pull/882/files#diff-bc88be45800e96e69c57218fd5c788c7R718) now that does nothing, save for [setting the running element for the page](https://github.com/Kozea/WeasyPrint/pull/882/files#diff-9f5170b54e3d607d6e89661639e531e7R671). This way, the last `running()` on the page wins, that's how I interpret the spec.

The not-doing-anything-box seems clumsy to me, can this be more elegant?

hbrunn:
@liZe thanks for the pointers. now this PR actually works, and I just need to wrap it up. last question: I need to have `get_content_list` called on the running box for every time it's rendered, how do you propose to do that?

hbrunn:
@liZe this is ready for review now

liZe:
> @liZe this is ready for review now

OK, thanks! I'll look at it.

hbrunn:
great. I just noticed I failed to remove the note about running elements in the docs, just pushed a fixup commit for that.
And I notived late yesterday that `content: element(...)` elements aren't always laid out the way I'd expect it (not take the full width of the page), but that might also be a wrong expectation. We'll see about that during review I think.

Here the file I mostly used for testing:

```html
<html>
    <head>
        <style type="text/css">
            .header {
                position: running(header);
            }
            .footer {
                position: running(footer);
                text-align: center;
            }
            @page {
                @top-center {
                    content: element(header);
                }
                @top-right {
                    content: string(header_string);
                }
                @bottom-center {
                    content: element(footer);
                    /* without this, there's a line break before the pages
                       span
                    */
                    width: 100%;
                }
                @bottom-right {
                    content: counter(page);
                }
            }
            h1 {
                margin-bottom: 15cm;
                string-set: header_string content();
            }
            .page:before {
                content: counter(page);
            }
            .pages:before {
                content: counter(pages);
            }
            .header_string:before {
                content: string(header_string);
            }
        </style>
    </head>
    <body>
        <div class="header">
            Hello world file: <span class="header_string" />
        </div>
        <div class="footer">
            <span class="page" /> of <span class="pages" />
        </div>
        <h1>test1</h1>
        <h1>test2</h1>
        <h1>test3</h1>
        <h1>test4</h1>
        <h1>test5</h1>
        <h1>test6</h1>
        <h1>test7</h1>
        <h1>test8</h1>
        <h1>test9</h1>
        <h1>test10</h1>
        <h1>test11</h1>
        <h1>test12</h1>
        <div class="header">
            Hello world2
        </div>
        <h1>test13</h1>
    </body>
</html>
```

which then yields [test.pdf](https://github.com/Kozea/WeasyPrint/files/3641350/test.pdf)

liZe:
Thanks a lot :clap:. The global way it works seems to be OK for me.

I can merge this PR, but I'll change some little things:
- We should use `('running()', name)` instead of a dedicated class, because that's how it's done for other functions. One day we'll use real objects for everything I suppose…
- `__deepcopy__` should be a normal method.
- The `inline_in_block` and `block_in_inline` should work out of the box, there's probably something wrong elsewhere if we have to add `or box.is_running()`.
- The `RunningPlaceholder` probably has to be removed. Having a placeholder inheriting from `BlockBox` with random used values set is a huge lie that will bring much more harm in the future than the problems it solves right now :smile:.

> I'd expect it (not take the full width of the page), but that might also be a wrong expectation. We'll see about that during review I think.

I have to dive into the code by fixing the points listed above, I'll check this problem afterwards.

weasyprint/layout/__init__.py
liZe: > This way, the last `running()` on the page wins, that's how I interpret the spec.

Actually, `element()` takes an optional second argument to define which one wins (just as in `string()`).

liZe: > This way, the last `running()` on the page wins, that's how I interpret the spec.

There's a second optional parameter for `element()` to define this. We have to use the same behavior as `string()`.

hbrunn:
thanks! If you give me a few pointers I'll be happy to investigate myself, integrating weasyprint into odoo is my current pet project :-)

To my understanding we will need some way to know where some box would be in the flow in order to use the correct element in the page margins, that's why I came up with the placeholder. Is your critique specifically that I made this a `BlockBox` and I should be using a more generic class higher up in the inheritance chain as base, or do you disagree with the placeholder altogether?

If you can't answer without doing the code dive yourself, I'll wait and check out the diff.

liZe:
> integrating weasyprint into odoo is my current pet project :-)

:heart:

> thanks! If you give me a few pointers I'll be happy to investigate myself

I've made some commits already, but I'll let you some work if you prefer, no problem :wink:.

> To my understanding we will need some way to know where some box would be in the flow in order to use the correct element in the page margins, that's why I came up with the placeholder. Is your critique specifically that I made this a `BlockBox` and I should be using a more generic class higher up in the inheritance chain as base, or do you disagree with the placeholder altogether?

The problem is that the box is removed from the page, we don't need to keep a placeholder to render it at the place it should have been, as we do for absolute boxes for example. When rendered in the margins, it's totally unrelated to its place in the page content (I hope!), unlike absolute boxes. You can see in 0a403d9d that everything works well without it.

liZe:
Hello @JanPretzel

> I just tried this feature and it is awesome! But I encountered a problem, when I add a `table` in the header of the example from @hbrunn I get an error.

Could you please open an issue about that? Thank you!

rjoly:
Nice job!

It would be awesome if we could use this feature in a near-future deployed application. Do you think that you could give a release date for v51?

weasyprint/layout/__init__.py
liZe: For the record:

> I think that the specification isn't clear enough. It's hard to know which running element you'll have in the page margin when multiple running elements are in the page. The `string()` function at least has a second argument for that.

They’ve added a second parameter to the `element()` function, just as it’s done for `string()`.

PR #887, opened 2019-06-24
Implement text-decoration: wavy

randomstuff:
Maybe using bezier cuve instead of arc os fircle would give a nicer result.


liZe:
Cool, thank you!

liZe:
I've cleaned a little bit your code, but nothing bad at all :heart:.

I love the rendering, I find it much better than Firefox, Safari and Chrome, especially when you change the color in one wave. Thanks a lot!

PR #902, opened 2019-07-15
Add state to bookmark tree documentation.

stanim:
There is a breaking change in make_bookmark_tree, which was not documented.


liZe:
Thank you!

PR #903, opened 2019-07-16
tables: add -weasy-break-inside-threshold proprietary extension

pmjdebruijn:
break-inside: avoid-page is an all of nothing affair, when applied
on large table row group the result may be less than stellar with
regard to large amounts of useless semi-unintentional white space.

this patch introduces a threshold where the largest size of a
table row group is defined that is to honour the break-inside.
attribute.

usage example:

tbody
{
  break-inside: avoid-page;
  -weasy-break-inside-threshold: 100mm;
}

### DIFF
--- a/weasyprint/css/properties.py
+++ b/weasyprint/css/properties.py
@@ -203,6 +203,7 @@ INITIAL_VALUES = {
     'anchor': None,  # computed value of 'none'
     'link': None,  # computed value of 'none'
     'lang': None,  # computed value of 'none'
+    'break_inside_threshold': float('inf'),
 
     # Internal, to implement the "static position" for absolute boxes.
     '_weasy_specified_display': 'inline',
--- a/weasyprint/css/validation/properties.py
+++ b/weasyprint/css/validation/properties.py
@@ -16,10 +16,10 @@ from ...formatting_structure import counters
 from .. import computed_values
 from ..properties import KNOWN_PROPERTIES, Dimension
 from ..utils import (
-    InvalidValues, check_var_function, comma_separated_list, get_angle,
-    get_content_list, get_content_list_token, get_custom_ident, get_image,
-    get_keyword, get_length, get_resolution, get_single_keyword, get_url,
-    parse_2d_position, parse_function, parse_position, single_keyword,
+    LENGTHS_TO_PIXELS, InvalidValues, check_var_function, comma_separated_list,
+    get_angle, get_content_list, get_content_list_token, get_custom_ident,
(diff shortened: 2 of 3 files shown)


pmjdebruijn:
Some sample testcases:

[testcase_nothreshold.pdf](https://github.com/Kozea/WeasyPrint/files/3396982/testcase_nothreshold.pdf)
[testcase_threshold.pdf](https://github.com/Kozea/WeasyPrint/files/3396983/testcase_threshold.pdf)

[testcase_nothreshold.txt](https://github.com/Kozea/WeasyPrint/files/3396992/testcase_nothreshold.txt)
[testcase_threshold.txt](https://github.com/Kozea/WeasyPrint/files/3396993/testcase_threshold.txt)

Aside from whitespace issues, there are some resulting issues where the table header isn't being repeated on new pages.

pmjdebruijn:
There are some caveats to this patch:

[1] I wasn't sure where to do the units to internal 96px-per-inch conversion, I went with css/validation/properties.py since that already depended on utils.py, and therefore not creating new interdependancies in the code.

[2] this construct may extend to other object-types as well, this only implements it for tables.

liZe:
Hello!

First of all, thank you for this pull request.

WeasyPrint tries hard to avoid proprietary properties. It's already not easy to implement and maintain the properties proposed by the W3C, and even if it's very tempting to add a property when there's something you can't do with CSS, it's a bad idea in the long term. Even simple properties are very hard to define precisely, and we need this precision to have interoperability between implementations, which is what WeasyPrint fights for. That's why we have to rely on a spec to add this feature.

It's hard to refuse new features, it's hard to refuse contributions (and this one is short and clean!), but I'm sorry: I won't merge this pull request.

The solution to solve this problem is to discuss this feature with the W3C. You can open a new issue [on their GitHub page](https://github.com/w3c/csswg-drafts/issues) and see if people are interested in defining this property. I know that it's a longer path to reach your goal, but it's an interesting one, and it's the best one to see the feature included in WeasyPrint in the future.

pmjdebruijn:
I see why you'd want to resist proprietary tags...

However if you take a look at my samples, there a bug that's at least workaround by this as well.

With break-inside: avoid-page defined, table headers aren't repeated as the should.

So there is a logic error somewhere, making break-inside conditional on height fixes this. While my patch makes this configurable through a proprietary tag, wouldn't it make sense to at least keep the extra conditional with the conditional being (hardcoded) the tbody height < page_height - page_bottom_margin - page_top_margin?

Issue #906, opened 2019-07-22
Table auto-layout differences compared to browsers


rtpg:
So I have this HTML 
```
<style>
td, div {
  border: 1px solid black;
}
</style>
<div style="width: 1cm;">
  Hi this is 1cm long
</div>

<div style="width: 1cm; white-space: nowrap">
  Hi this is also 1cm long
</div>

<table style="width: 10cm">
  <tr>
    <td style="width: 1cm;">
      This is 1cm
    </td>
    <td style="width: 1cm; white-space: nowrap">
      This is not 1cm
    </td>
    <td>
      This will take the rest
    </td>
  </tr>
</table>
```
(I have a nice fiddle for this https://jsfiddle.net/4wr36ouh/)

Here I have some 1cm divs, along with some tables that have 1cm columns (or rather, they should).

the `white-space: nowrap` property on the second table cell means that browsers end up _expanding this column/cell to be more than 1cm long_.

However, in weasyprint, this property causes the _content to escape the cell, with the cell staying 1cm_. See this PDF [test_pdf.pdf](https://github.com/Kozea/WeasyPrint/files/3416544/test_pdf.pdf)

Now, the auto-layout spec is User Agent-specific.... I don't know if this is just an incidental issue in the weasyprint algorithm leading to the difference, or if it's a fundamental difference in the interpretation of the `width` property on table cells, but the browser results tend to look better and feel like a nice target to emulate.

I'm going to try looking into this a bit more later, I believe there could be a very small patch that unifies behavior here. But I would like to know if there's any intentionality in this behavior that I should be aware about?

liZe:
> But I would like to know if there's any intentionality in this behavior that I should be aware about?

No, there's not. It's probably caused by a wrong value for the min width of table cells.

rtpg:
Yeah looking into this a bit more I found that `content_min_width` for the table cell was returning just the width, instead of a content-aware result that took into account that would adapt to the non-wrapping.

liZe:
> Yeah looking into this a bit more I found that `content_min_width` for the table cell was returning just the width, instead of a content-aware result that took into account that would adapt to the non-wrapping.

That's right, we treat cells as we treat other containing blocks when calculating minimum content width. We should check that content doesn't escape from the block for cells instead.

There's an extra `if` condition to add at the beginning of `min_content_width`. @rtpg Are you interested in adding this code? I can help if you want, or I can add it myself if you're not interested, no problem!

[...]

PR #907, opened 2019-07-23
Refactor LayoutContext building

rtpg:
This will provide a slightly nicer API for generating contexts in tests (doing this as part of resolving #906)


liZe:
It's better this way, thank you. I'll fix the lint for you!

PR #909, opened 2019-07-23
[WIP/RFC] Get auto-layout tables to handle non-whitepsace wrapping correctly

### LINKED ISSUE #906: Table auto-layout differences compared to browsers

rtpg:
This was my attempt to fix #906, though I think I'm a bit out of my depth at this point.

This includes #907 , which I used to write the test. The values being asserted are wrong, but I think that conceptually the test is a good way to cover a lot of ground.

### DIFF
--- a/weasyprint/document.py
+++ b/weasyprint/document.py
@@ -15,6 +15,7 @@ import shutil
 import warnings
 
 import cairocffi as cairo
+from weasyprint.layout import LayoutContext
 
 from . import CSS
 from .css import get_all_computed_styles
@@ -347,9 +348,11 @@ class Document(object):
     <weasyprint.fonts.FontConfiguration>`.
 
     """
+
     @classmethod
-    def _render(cls, html, stylesheets, enable_hinting,
-                presentational_hints=False, font_config=None):
+    def _build_layout_context(cls, html, stylesheets, enable_hinting,
+                       presentational_hints=False,
+                       font_config=None):
         if font_config is None:
             font_config = FontConfiguration()
         target_collector = TargetCollector()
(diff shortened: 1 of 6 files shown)


liZe:
Fixed by 8d10c01.

Issue #760, opened 2019-01-03
Support for @counter-style


TUSF:
On feature I'd like to see, is support for [@counter-style](https://www.w3.org/TR/css-counter-styles-3/). It appears to only be supported by [FireFox](https://developer.mozilla.org/en-US/docs/Web/CSS/@counter-style), but I'd really like to be able to use it through WeasyPrint.

In my case specifically, I want to be able to have list and page counters that that aren't base ten, or anything else not on the predefined list of styles.

liZe:
Good idea! A basic implementation should be quite easy to add (basically add new CSS rules and put `formatting_structure/counters.py`'s `STYLES` in the document). If anyone is interested, I'd be happy to help!

TUSF:
Think I'll try. Might as well contribute to WeasyPrint given I intend to be using it quite a bit :P

Will just have to acclimate myself to how the repo is structured, and where everything is actually defined and stuff. I also haven't used Python enough recently haha. Main thing I guess would be finding where `@` rules (`@counter-style`) are processed and adding a value function? (`symbols()`)

Adventure time, I guess!

liZe:
@TUSF Good news, thank you! Here are some steps that could help you.

**For now, you probably only want to handle the `disc` counter.** The other ones will be quite easy to add later.

1. You can parse the `@counter-style` rules in [`preprocess_stylesheet`](https://github.com/Kozea/WeasyPrint/blob/adb2e3704dc621c41b4413542306983d4c1c7ccd/weasyprint/css/__init__.py#L624). There's no need to do the validation part (i.e. discard unknown descriptors), we can do that later. You can just parse a few descriptors for now, no need to implement the whole specification :wink:.

2. The result of your parsing should be put in a dictionary (let's call it `counter_style`) whose keys are the counter types names, and whose values are functions that take one argument (`value`, the number, for example '3') and that return the corresponding counter value (for example 'iii' for lower roman). For now, you can only support one simple case, for example `disc`, using the `counter_style['disc'] = lambda value: '•'` code when you detect a `disc` counter style rule.

[...]

PR #911, opened 2019-07-25
Support @counter-style

liZe:


liZe:
> I followed the steps above, and I was able to load custom `disc` counters from a `counter_style` dict defined in `Document._render` and populated in `preprocess_stylesheet`. There is still no parsing for custom `@counter-style` rules, just a hard coded value for `disc`.

:+1:

> I tried hard coding additional `@counter-style` rules in `counter_style`, but I couldn't get them to render. From what I could gather, it looks like the function that parses the `list-style-type` rule defaults to `disc` when a non-standard value is passed. However, I'm not too sure where that logic is at in the codebase.

There's a validator that only allows values predefined in CSS2.

https://github.com/Kozea/WeasyPrint/blob/f0001604231e4c3c5bff07bfa1930d176712c767/weasyprint/css/validation/properties.py#L870-L874

You can `return True` for now, we'll validate this later.

> What steps should I take from here? Should I try to edit the `list-style-type` parser? Or just finish hard coding the default `list-style-type` rules in the new `counter_style` logic? Thanks in advance!

1. You can change the validator, as written above.
2. Contrary to what I've said, you can remove the `@counter-style` rules from the UA CSS, because it will add extra complexity for now (because the UA stylesheet is parsed when WeasyPrint is imported, not for each document). Put them in your testing document instead, you've already done the logic, it should work. (And you can remove the `counter_style={}` parameter from `HTML5_UA_STYLESHEET` too.)

When it's done, you're at the end of step 5 of [#760](https://github.com/Kozea/WeasyPrint/issues/760#issuecomment-461035577). Your `'foo'` counter should work. :v: :tada:

The next step is to get the counter name from the rule. You basically need to copy the first line of `parse_page_selectors` and take the value of the first token.

You'll then need to get the descriptors (`system`, `negative`, `prefix`…), that's a hard part. A pretty straightforward solution is to use the `preprocess_desrciptors` function, and declare the descriptors (in `css/validation/descriptors.py`) just as `font_variant`, `font_weight`… Only supporting `system` and `symbols` should be enough to have something interesting.

Last part is to replace the `lambda` with real code. Almost all you need already exists in `formatting_structure/counters.py`, but it will need to be adapted. Tell me when you're there!

(After that, we'll need to handle other descriptors, split allowed descriptors between `@font-face` and `@counter-style`, validate `list-style-type`, put the default counters back in the UA stylesheet, add tests… That's the boring stuff I can do if you're not interested!)

liZe:
(Fun fact: I'm currently implementing `::marker`. I'll do the merge, don't worry :smile:.)

liZe:
Duplicated by #1017.

PR #922, opened 2019-08-15
Improve performance

### LINKED ISSUE #700: Implement ::marker
Currently I can't find a reliable way to specify custom style for the numbers (markers) in numbered lists of paragraphs in WeasyPrint — that is, for HTML like this:

````HTML
<ol>
  <li><p>Urkkk!
  <li><p>Clank!
  <li><p>Glurpp!
</ol>

<ol>
  <li>Biff!
  <li>Ooooff!
  <li>Bang-eth!
</ol>
````

being able specify styles for the numbers.

* The spec defines [the `::marker` pseudo-element](https://drafts.csswg.org/css-lists/#marker-pseudo) ... but that's a draft which says it isn't ready for implementation yet.
* Putting [a counter in `li::before`](https://codepen.io/sawmac/pen/txBhK) and setting some margins and padding is suggested by some, but that only works for short list items that directly contain text; for list items that contain paragraphs, `<p>` starts a new line, under the `li::before`. (This isn't a WeasyPrint bug; it's a defect in this approach.)
* Floating or [absolutely positioning the `li::before`](https://jsfiddle.net/vandigroup/3Lvpt9rc/1/) avoids that limitation and works fine in browsers. But in WeasyPrint, if the list item starts near the bottom of a page, sometimes the `<li>` gets bumped on to the next page, leaving the pseudo-element hanging on the previous page. Chrome appears not to do this (though possibly the page-break simply wasn't being triggered in quite the right spot.) I fear that this is related to #36. I see also there are other issues involving list marker positions (such as #287 and #55), but I'm not sure if they are relevant once the marker has been positioned or floated.
* For lists with paragraphs in, a workaround could be to apply the desired number styling (such as colour and font) to the entire list, then undo that for `li > p`. That obviously wouldn't work for lists that _don't_ contain paragraphs. Handling both types of list would require applying different styling to an `<li>` depending on whether or not it contains any `<p>`s. The [`:has()` pseudo-class](https://drafts.csswg.org/selectors-4/#relational) would allow determining that ... but is also a draft, and not implemented by any browsers either.

So I don't know what specifically the bug is here — but surely WeasyPrint should provide some way of achieving this.

liZe:
Also fix #700.


Tontyna:
:+1: for the cleanups

PR #927, opened 2019-08-26
Dont use deprecated Logger.warn

Tontyna:
... use 'warning' instead


liZe:
Oops, it was supposed to be fixed a long time ago…

Issue #92, opened 2013-05-22
How to repeat on each page of complex headers (eg, tables)?


pikhovkin:
Hello!
Please tell me how to repeat on each page of complex headers (eg, tables) at the top and bottom of the page?

jsenecal:
FWIW, I'm using what @pikhovkin shared and it does wonders. Like @SimonSapin mentionned - it could be nice to document those APIs.

doekman:
Hi, based on the gist of @pikhovkin, I created an extension to Flask-WeasyPrint in [this gist](https://gist.github.com/doekman/4674dd2d05b643ef57e747cfdc81c3c9). It's still a bit rough, but I thought to share it, report my findings and possibly get feedback.

I understand it uses undocumented API's. I take my chances, because I don't know an other way to create the PDF's I need with WeasyPrint. But documenting them would be appreciated a lot!

It uses Jinja2 to render the Overlay HTML (header/footer-HTML), so page number and number of pages can be used in HTML. Custom positioning of content can be done by supplying positions through the context which will be rendered by Jinja2.

The render method shares the `stylesheets` with the Overlay pages to prevent errors like "CAIRO out of memory" (I probably got this error because I used @font-face definitions, and this way, the FontConfiguration is shared).

whitelynx:
Has there been any progress on GCPM in the last 3 years? Is it getting to a stage where we could start implementing something based on it?

liZe:
> Has there been any progress on GCPM in the last 3 years?

Yes, you can check [what the CSSWG did](https://drafts.csswg.org/css-gcpm/).

What we want is [running elements](https://drafts.csswg.org/css-gcpm/#running-elements). With named pages (added in #495), we should be able to repeat complex headers in page margins.

> Is it getting to a stage where we could start implementing something based on it?

Of course. Even if GCPM is "a long story" and "a big mess", running elements can IMHO be added safely to WeasyPrint. The spec didn't really change during the last years, the [only reported issue](https://drafts.csswg.org/css-gcpm/#issue-81a9d129) even includes an interesting alternative syntax to fix it.

Implementing both syntaxes looks like a good idea for me. Even if it's a draft, the feature is useful and having it implemented may even help the CSSWG.

Anyone interested? I'll be happy to help!

liZe:
@doekman

[...]

PR #930, opened 2019-08-29
feature to help book designers debug their line grid

rbergmair:
Dear WeasyPrint Team!

First of all: Thanks a lot to the maintainers here for creating WeasyPrint and making it open source.  I am in the middle of doing an expansive book-publishing project with WeasyPrint and have found it absolutely great.  I'm sure this software has a bright future.

This PR is a piece of functionality that I've added that helps me in debugging design issues for the book.  There seems to be a pretty hard rule within the design profession that lines of a book on two opposing pages should always line up to the same raster.  This is not always easy to do in HTML, since HTML will reflow stuff and if there is even a single box that is not the height you are expecting it to be, then it will move everything below it and push all the lines to go out of alignment.

So I've created this function where, by adding the parameter `--grid-debug somefile.csv`, weasyprint will create a file `somefile.csv` that the designer can expect to check if lines have the expected heights and their y_positions line up with each other.

I'm hoping other people might find this useful, hence the PR.

Thanks again, and keep up the good work!

cheers,

Richard

### DIFF
--- a/weasyprint/__init__.py
+++ b/weasyprint/__init__.py
@@ -135,7 +135,7 @@ class HTML(object):
         return get_html_metadata(self.wrapper_element, self.base_url)
 
     def render(self, stylesheets=None, enable_hinting=False,
-               presentational_hints=False, font_config=None):
+               presentational_hints=False, font_config=None, grid_debug=None):
         """Lay out and paginate the document, but do not (yet) export it
         to PDF or PNG.
 
@@ -165,11 +165,11 @@ class HTML(object):
         """
         return Document._render(
             self, stylesheets, enable_hinting, presentational_hints,
-            font_config)
+            font_config, grid_debug)
 
     def write_pdf(self, target=None, stylesheets=None, zoom=1,
                   attachments=None, presentational_hints=False,
-                  font_config=None):
+                  font_config=None, grid_debug=None):
         """Render the document to a PDF file.
 
         This is a shortcut for calling :meth:`render`, then
@@ -208,7 +208,8 @@ class HTML(object):
         return self.render(
             stylesheets, enable_hinting=False,
             presentational_hints=presentational_hints,
-            font_config=font_config).write_pdf(
+            font_config=font_config,
+            grid_debug=grid_debug).write_pdf(
                 target, zoom, attachments)
 
     def write_image_surface(self, stylesheets=None, resolution=96,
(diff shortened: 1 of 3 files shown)


liZe:
> First of all: Thanks a lot to the maintainers here for creating WeasyPrint and making it open source. I am in the middle of doing an expansive book-publishing project with WeasyPrint and have found it absolutely great. I'm sure this software has a bright future.

:heart: 

> This PR is a piece of functionality that I've added that helps me in debugging design issues for the book. There seems to be a pretty hard rule within the design profession that lines of a book on two opposing pages should always line up to the same raster. This is not always easy to do in HTML, since HTML will reflow stuff and if there is even a single box that is not the height you are expecting it to be, then it will move everything below it and push all the lines to go out of alignment.

That's really useful, thanks!

For [reasons explained in another comment](https://github.com/Kozea/WeasyPrint/issues/92#issuecomment-483991137), I think that it would be better to add this code in the documentation, not in WeasyPrint.

Do you want me to do this for you?

rbergmair:
Are you sure you're not getting confused between my two PRs now?  I submitted two PRs almost simultaneously:
https://github.com/Kozea/WeasyPrint/pull/932: facility to override font metrics
https://github.com/Kozea/WeasyPrint/pull/930: feature to help book designers debug their line grid

You're pointing to a comment explaining that it's a bad idea to start coloring outside the lines as far as the HTML/CSS specs are concerned.  Point well taken.  But it would seem to apply primarily to the other PR (https://github.com/Kozea/WeasyPrint/pull/932).

This PR is about creating a logfile that designers can use for debugging issues with line grids.  Book designers are likely to want to pay close attention to those grids, as there seems to be a hard and fast rule in the profession that lines on opposing pages of books are supposed to always line up with each other which is not always easy to achieve with HTML.  So a logfile like this can be of great help here.

I think that's a feature that makes sense, even while rejecting the other PR.

liZe:
> Are you sure you're not getting confused between my two PRs now?

No, I'm not. But my answer was not as clear as it was in my head, sorry :smile:.

Debugging the line grid is useful when you want to understand what's going on with text. We could have the same kind of features to track boxes positions and sizes, to show flex metrics, to explain table layout, and so on. That's what browsers do. The problem with this kind of features is that you can always add more, and even if they're small and clean (and your PR is), they add complexity and have to be maintained in the future. Moreover, information needed when debugging highly depends on what you're exactly trying to debug, and you'll always need an extra field that's not included yet.

That's why I'd prefer to have this in the documentation. I think that it can be added only using monkey patching, and the code snippet will be adapted for each use case if needed.

Issue #497, opened 2017-08-17
Don't copy styles only to change margins, borders and paddings


liZe:
Many situations force to override (often remove) margins, borders and paddings, including:
- table cells with collapsing borders,
- blocks split between pages,
- inline boxes split by block boxes.

Depending on the situation, the style is copied and modified, or the computed values are applied to the box (`box.padding_right` and its friends). There's no perfect solution, as the first one greatly increases memory usage, and the second one breaks `min/max-width` calculation and box copy.

We have to find a solution to:
- never copy nor modify style dicts,
- override margins, borders and paddings on styles without modifying the dict,
- avoid mixing `box.style['margin_left']` and `box.margin_left` (as it's done in `mix/max-width` calculation on boxes after dimensions are already set on the box),
- keep speed (i.e. don't use the copy-on-write method used before, as it was overriding the `__getitem__` method that's over-optimized).

liZe:
In [this wonderful blog entry](https://hacks.mozilla.org/2017/08/inside-a-super-fast-css-engine-quantum-css-aka-stylo/), the solution I was thinking of is given in the chapter called "A sidenote: style struct sharing".

liZe:
4057a92 uses real dicts for performance and use [ChainMap](https://docs.python.org/3/library/collections.html#collections.ChainMap)s to keep original dicts when changing padding/borders/margins, breaking the Python2 compatibility.

liZe:
See https://github.com/Kozea/WeasyPrint/pull/948#issuecomment-534608574 for more information about where to remove style edition. #950 removed the worse parts, but it’s not finished yet.

Having frozen styles with computed values is the way to go.

Tontyna:
> Having frozen styles with computed values is the way to go.

Of course. But grepping for modifications of `Box.style` it looks like since #950 we don't touch the immutable computed values anymore. The modfications rather ensure the correct initialization of computed values for specialized boxes.

What I don't like is the fact that the pagination stores used values (and other properties related to layout) as attributes in the root_box tree and its children. AFAIK it doesn't hurt, not sure whether it's required, but it definitely feels wrong.

A 'style struct sharing' would be great. At the moment each box of the formatted root_box tree has its own complete style dictionary -- style sharing only happens between the formatted box and its paginated counterpart(s).

PR #948, opened 2019-09-20
Fix comparison in ParentBox._remove_decoration

### LINKED ISSUE #943: Footer appears to be bumping up list items on a `<ul>` that's split across pages
I'm only noticing this in the table of contents so far. Items 2, 4, 6, and 8 in the attached pdf. They appear to be bumped up about the same distance as my lower @page declarations.
I'm making the PDF with the following commands:
```
pandoc -s --template="templates/default.html" -f markdown-smart --toc -c style.css "$filename.md" -o "$filename.html"
python3 -m weasyprint "$filename.html" "$filename.pdf"
```
[test.pdf](https://github.com/Kozea/WeasyPrint/files/3622554/test.pdf)

Tontyna:
Should fix #943

> Don't copy style if there's no need to, save some memory

Indeed.

### DIFF
--- a/weasyprint/formatting_structure/boxes.py
+++ b/weasyprint/formatting_structure/boxes.py
@@ -321,7 +321,7 @@ class ParentBox(Box):
             self._reset_spacing('top')
         if end:
             self._reset_spacing('bottom')
-        if (start or end) and old_style == self.style:
+        if (start or end) and old_style != self.style:
             # Don't copy style if there's no need to, save some memory
             self.style = old_style
 


liZe:
That's another reason why we should **never** edit the style dictionary.

The current code is actually correct. This function removes decoration and spacing when a box is split between pages (as you say in #943). As the style may be shared between different boxes, we copy it before modifying it. But when the modified dictionary is the same as the original one, there's no need to copy it.

Your pull request actually reverts the changes made to `style` by `_reset_spacing`. We don't want that for sure. We want to avoid the copy (code in the `if` block) when it's not been modified (i.e when it's the same, `==`). When it's been modified, the copy is mandatory, we don't want to go in the `if` block that reverts the copy.

#943 comes from the fact that the style attached to the box on the first page has been modified by the box on the second page (its top margin has been removed). Your pull request fixes this because it copies the style between the two renderings: the `_reset_spacing` has already been called during the first rendering, when it's called during the second rendering the generated style is the same (of course), and the dictionary is copied.

We actually want to always copy the style when it's edited.

But.

Do we really need to edit the `style` dictionary? The layout should only be using computed values when decorations are removed (that's why this PR works even if `style` have wrong values). Removing the edition (ca0fd023) doesn't break a test, and I'm pretty sure it fixes #943.

Do you have a sample that I could use to add a test :smile:?

liZe:
(See #497 too).

Tontyna:
I was aware of #497 and I was confused by the contraintuitive `==` vs. `!=` and I'm still confused why and when the `style` is touched or not touched instead of or in conjunction with the box's properties and why all this happens multiple times. -- And I recall there was something with absolute or fixed boxes that required the copy/edit of the `style`, but cannot find the related issue...

That's my snippet:

```html
body {
  orphans: 1;
  widows: 1;
}
p {
  margin: 0;
  padding: 0;
  background-color: #999;
}
.deko {
  border: dotted;
  padding: 1em;
  margin: 1em;
  background-color: #ccc;
}
.forceremake::after{
  content: " at " target-counter("#lastp", page);
  color: red;
}
@page {
  size: 12cm 8cm;
  border: 1px solid red;
}
</style>
<body>
  <p class="forceremake">
    force remake for page 1
  </p>
  <div class="deko">
    Lorem ipsum enough lines 
    to cross the page dolor sit 
    amet sed ipsimus dingsda 
    et crossing the page is 
    required mullo nennam 
    desirpitus to zero out the 
    top properties of this 
    block item in corpore desepiutum
    noneque dicit.
  </div>
  <p id="lastp" >
    the target
  </p>
</body>
</html>
```

Lot of deko stuff to visualize the disappearing top styles.

Tontyna:
@liZe - with your commit (ca0fd02) the top properties are perfect. But what about d4357e5, supposed to "fix blocks split between pages and inlines boxes split by block boxes."?

liZe:
> But what about [d4357e5](https://github.com/Kozea/WeasyPrint/commit/d4357e5a4a8570cc7e12be6d1e7311e78c2dc1fd), supposed to "fix blocks split between pages and inlines boxes split by block boxes."?

I wish I had added some tests. I'll spend some time trying to find broken use cases and playing with the W3C testing suite before merging.

Tontyna:
I'm not happy with ca0fd02. Though it obviously fixes #943. 
It modifies `_remove_decoration` of `ParentBox`, but what about `InlineLevelBox`? And what about `_reset_spacing` left with the border radius? There must have been a reason why these functions exist to zero out the whole `style` decoration.

`pages.make_box()` calls `_reset_spacing` for MarginBoxes, `ParentBox.copy_with_children` and `block_container_layout` want to `_remove_decoration` -- there must be a reason why these functions exist and zero out the whole `style` decoration, not only the box's properties.

But maybe that reason has become obsolete and editing the style dict isn't required anymore?

BTW: During my research for #943 I was wondering why the box on the second page had no problems with its *bottom decoration*...

I whish there were more **comments in the source**! It's hard to find reasons for source code in the commit history when there are almost no hints but the limited commit titles.

liZe:
> I'm not happy with [ca0fd02](https://github.com/Kozea/WeasyPrint/commit/ca0fd0239fddf9227de7e428278f6d6831ea4d63). Though it obviously fixes #943.

:cry:

I agree, it's weird. But I'd really love to keep the code with this hack removed, even if we have to fix bugs elsewhere.

> It modifies `_remove_decoration` of `ParentBox`, but what about `InlineLevelBox`?

Yes, we have to remove the copy hack too for `InlineLevelBox`. (It doesn't break tests!)

But about border-radius, you're right. I have to fix this.

> there must be a reason why these functions exist and zero out the whole `style` decoration, not only the box's properties.
> 
> But maybe that reason has become obsolete and editing the style dict isn't required anymore?

I think that it's obsolete, or at least that use cases where it's still needed can be easily fixed. But of course it's hard to find these use cases.

> I whish there were more **comments in the source**! It's hard to find reasons for source code in the commit history when there are almost no hints but the limited commit titles.

I wish there were **more comments** and **more tests**. I'm trying harder and harder to add long commit messages, tests and comments, but I've had my weakness moments too.

> BTW: During my research for #943 I was wondering why the box on the second page had no problems with its _bottom decoration_...

It's probably because the second page doesn't have to be repaginated.

Tontyna:
> It's probably because the second page doesn't have to be repaginated.

Of course I forced repagination for page nr 2, too -- but since after the firs pass the **style** for the box is **manifested** to the top-less version, no problem. 

What I don't understand: What prevents the bottom-less style (for page 1, first pass) from being manifested? Does that mean that the box on page 1 is not bottom-less? Only we dont see the padding and margin and border because they are cut off (by the page margin)?

liZe:
I've open #950 with proposed fixes. There's still work to do, at least a lot of tests.

> Of course I forced repagination for page nr 2, too -- but since after the firs pass the **style** for the box is **manifested** to the top-less version, no problem.

OK.

> What I don't understand: What prevents the bottom-less style (for page 1, first pass) from being manifested? Does that mean that the box on page 1 is not bottom-less? Only we dont see the padding and margin and border because they are cut off (by the page margin)?

For the second part of the block (on page 2, first pass), `_remove_decoration` is called twice (I don't know why). The second time, the style didn't change and the dictionary is not copied. It is then reused during the second pass, when `_remove_decoration` is called again for the first part of the block, with the missing top decoration. The bottom decoration is removed during this call, the dict is modified and thus copied. The second part gets the same style values as the ones given to the first part (the top decoration is already missing).

Here's some dirty debugging. Each time, the first line is printed at the beginning of `_remove_decoration` and the second one at the end. `True` and `False` are for `start, end`, and the dimensions are the top and bottom margins.

```
INFO: Step 1 - Fetching and parsing HTML - /tmp/break.html
INFO: Step 2 - Fetching and parsing CSS - CSS string
INFO: Step 3 - Applying CSS
INFO: Step 4 - Creating formatting structure
INFO: Step 5 - Creating layout - Page 1
<BlockBox div> Dimension(value=16.0, unit='px') Dimension(value=16.0, unit='px')
copy False True Dimension(value=16.0, unit='px') Dimension(value=0, unit='px')
INFO: Step 5 - Creating layout - Page 2
<BlockBox div> Dimension(value=16.0, unit='px') Dimension(value=16.0, unit='px')
copy True False Dimension(value=0, unit='px') Dimension(value=16.0, unit='px')
<BlockBox div> Dimension(value=0, unit='px') Dimension(value=16.0, unit='px')
do not copy True False Dimension(value=0, unit='px') Dimension(value=16.0, unit='px')
INFO: Step 5 - Creating layout - Repagination #1
INFO: Step 5 - Creating layout - Page 1
<BlockBox div> Dimension(value=0, unit='px') Dimension(value=16.0, unit='px')
copy False True Dimension(value=0, unit='px') Dimension(value=0, unit='px')
INFO: Step 5 - Creating layout - Page 2
<BlockBox div> Dimension(value=0, unit='px') Dimension(value=16.0, unit='px')
do not copy True False Dimension(value=0, unit='px') Dimension(value=16.0, unit='px')
<BlockBox div> Dimension(value=0, unit='px') Dimension(value=16.0, unit='px')
do not copy True False Dimension(value=0, unit='px') Dimension(value=16.0, unit='px')
INFO: Step 5 - Creating layout - Repagination #2
INFO: Step 5 - Creating layout - Page 1
<BlockBox div> Dimension(value=0, unit='px') Dimension(value=16.0, unit='px')
copy False True Dimension(value=0, unit='px') Dimension(value=0, unit='px')
INFO: Step 5 - Creating layout - Page 2 (up-to-date)
INFO: Step 6 - Drawing
INFO: Step 7 - Adding PDF metadata
```

I can try to find if it's possible to never edit styles during layout.

liZe:
> I can try to find if it's possible to never edit styles during layout.

Git grep gives me this:

```
weasyprint/css/__init__.py:                    style[name] = values, weight
weasyprint/css/__init__.py:                            style[name] = values, weight
weasyprint/css/__init__.py:                        style['padding_' + side] = computed_values.ZERO_PIXELS
weasyprint/css/__init__.py:                        style['margin_' + side] = computed_values.ZERO_PIXELS
weasyprint/css/__init__.py:        computed_styles[element, pseudo_type] = computed_from_cascaded(
weasyprint/css/__init__.py:                                style[name] = values, weight
weasyprint/css/__init__.py:                computed[name] = parent_style[name]
weasyprint/css/__init__.py:        computed['page'] = parent_style['page']
weasyprint/css/__init__.py:                computed[name] = specified[name] = parent_style[name]
weasyprint/formatting_structure/build.py:                    style['display'] = 'block'
weasyprint/formatting_structure/build.py:                    style['display'] = 'none'
weasyprint/formatting_structure/build.py:            box.style['white_space'] = 'pre-wrap'
weasyprint/formatting_structure/build.py:        marker_box.style['position'] = 'absolute'
weasyprint/formatting_structure/build.py:        marker_box.style['transform'] = (
weasyprint/formatting_structure/build.py:        wrapper.style[name] = table.style[name]
weasyprint/formatting_structure/build.py:        table.style[name] = properties.INITIAL_VALUES[name]
weasyprint/formatting_structure/build.py:            border_grid[grid_y][grid_x] = (score, (style, width, color))
weasyprint/formatting_structure/build.py:        box.style['border_%s_style' % side] = 'solid',
weasyprint/formatting_structure/build.py:        box.style['border_%s_width' % side] = twice_width / 2
weasyprint/formatting_structure/build.py:        box.style['border_%s_color' % side] = transparent
weasyprint/formatting_structure/build.py:    chosen_box.style['overflow'] = 'visible'
weasyprint/layout/flex.py:            new_child.style['width'] = 'auto'
weasyprint/layout/flex.py:            new_child.style['min_width'] = Dimension(0, 'px')
weasyprint/layout/flex.py:            new_child.style['max_width'] = Dimension(float('inf'), 'px')
weasyprint/layout/flex.py:            new_child.style['height'] = 'auto'
weasyprint/layout/flex.py:            new_child.style['min_height'] = Dimension(0, 'px')
weasyprint/layout/flex.py:            new_child.style['max_height'] = Dimension(float('inf'), 'px')
weasyprint/layout/flex.py:            child.style[axis] = 'max-content'
weasyprint/layout/flex.py:                child.style[axis] = 'auto'
weasyprint/layout/flex.py:                child.style[axis] = 'auto'
weasyprint/layout/flex.py:                        child.style[cross] = Dimension(
weasyprint/layout/inlines.py:        first_letter_style['position'] = 'static'
weasyprint/layout/pages.py:            style[propname] = ()
weasyprint/layout/pages.py:        style[propname] = tuple(justified_values)
weasyprint/layout/pages.py:        style['counter_increment'] = (
weasyprint/layout/pages.py:    style['overflow'] = root_box.viewport_overflow
weasyprint/tests/test_text.py:    new_style['font_family'] = MONO_FONTS.split(',')
```

- The ones in `css` should be OK.
- In `formatting_structure`, we're in the early steps of the rendering, we can accept modifications for now even if they can/should be removed in the future.
- The ones in flex must be removed.
- The one for the first letter is not causing troubles for now, but it should be removed.
- The ones in `pages.py` would be better handled with attributes on page and page-margin boxes.
- The one in the test can be ignored.

What do you think of that?

If it's not too slow, we could even try to use frozendicts. It should be possible to save a very large amount of memory with cache, as a lot of boxes use the same style without using the same object. [Mozilla's blog entry](https://hacks.mozilla.org/2017/08/inside-a-super-fast-css-engine-quantum-css-aka-stylo/) has really good ideas, but it's hard (impossible?) to get this kind of structures in Python.

Tontyna:
> For the second part of the block (on page 2, first pass), _remove_decoration is called twice (I don't know why).

Yep, same dirty debugging here and this second (`do not copy`) call is the one that manifests the top-less style. Thats why my clueless PR (forcing a copy) seemed to fix the bug.
Maybe we should detect, analyze and finally remove this (superfluent?) second call...

> What do you think of that?

Basically: Yes.
Not sure about `formatting_structures/build.py` -- thinking of the `parse_again` closures and the `target-collector`

liZe:
> Maybe we should detect, analyze and finally remove this (superfluent?) second call...

`_remove_decoration` is called when the parent box is copied to create the second part of the block. So it's called:
- during the first part layout,
- when the second part is created from the first part,
- during the second part layout.

I'll try to find if we really need to call it when it's copied.

Tontyna:
> So it's called:
> 1.  during the first part layout,
> 2.  when the second part is created from the first part,
> 3.  during the second part layout.

N° 1 is ok. 
N° 3 is obsolete (and the killer) if N° 2 already created a copy
Is N°2 required at all? Won't N°3 do the job anyway?

Tontyna:
Ah!

> I'll try to find if we really need to call it when it's copied.

is the same as

> Is N°2 required at all? Won't N°3 do the job anyway?

liZe:
> Is N°2 required at all? Won't N°3 do the job anyway?

5cb24e2c

Explicitely calling `_remove_decoration` is probably too much, but it's really easier to understand what's going on.

I don't know yet if I want to keep this, but if we do I'll remove the leading underscore.

Tontyna:
> I don't know yet if I want to keep this, but if we do I'll remove the leading underscore.

Yeah, noticed the underscore :grin: 

Wondering whether top and bottom margins/paddings/borders are the only properties requiring different manifestations when it comes to page breaks.

BTW: I like the `remove_decoration_sides` set.

Tontyna:
Had a look at the current #950 `_reset_spacing` -- I miss a `TODO: border-radius` :smirk: 

Indeed, it looks like modifying and copying the `style` never was required. The issue nagging my brain finally turned out to be #636 and wasn't related to shared modified `style`s at all.

Don't you recall what happened in August 2017 aka d4357e5?

liZe:
> Had a look at the current #950 `_reset_spacing` -- I miss a `TODO: border-radius`

You don't! It's fixed by b4ff03c, that's why `self.remove_decoration_sides` has been added.

> Indeed, it looks like modifying and copying the `style` never was required. The issue nagging my brain finally turned out to be #636 and wasn't related to shared modified `style`s at all.

Good to know.

> Don't you recall what happened in August 2017 aka [d4357e5](https://github.com/Kozea/WeasyPrint/commit/d4357e5a4a8570cc7e12be6d1e7311e78c2dc1fd)?

I really don't. Let's pretend it never happened… until next bug.

It was added after a period when a lot of things changed in the way styles were managed. WeasyPrint's code is much better now, I really think that it's not needed anymore.

The more I think about that, the more I'm convinced that styles should never be modified during layout.

> freeze StyleDicts, as it will avoid terrible future (current?) bugs, help us keeping a simple code (see the TODO in the StyleDict class), and make them hashable

That was my idea in August 2017, it's time to add tests and get rid of these "terrible bugs" for good.

Tontyna:
> The more I think about that, the more I'm convinced that styles should never be modified during layout

I agree. 

The style dict contains the **computed** values, calculated during "Step 3 - Applying CSS". 
In "Step 4 - Creating formatting structure" these computed values AKA 'StyleFor' are connected to the (initial, before-layout-) boxes.

The **used** values are determined during "Step 5 - Creating layout" and should be stored in the corresponding property of the (paginated) boxes.

Tontyna:
Finally I understand why `_remove_decoration_` is called twice. It's a result of the double trees -- the **master tree** created in "Step 4" and the **page tree** created for each page in "Step 5".

Each page is started with a `resume_at` pointing into the **master tree**. This `resume_at` enters `block_container_layout` (via `make_page` -> `block_level_layout` -> `block_level_layout_switch` -> `block_box_layout`) and its decoration is removed in an appropriate way. The **used values** are calculated and eventually a `copy_with_children` of the block  (paginated copies of the children inclusive, decoration explicitly removed in #950) is appended to the **page tree**.

What I don't like about this is the fact that the **used values**, apart from some cosmetics concerning the height, are calculated with and stored in the **master** BlockBoxes. But that's probably nitpicking, unavoidable and doesn't hurt. Not touching the **comupted values** is the point.

Tontyna:
Closing this PR due to its uselessness

PR #950, opened 2019-09-24
Don't edit style when splitting boxes

### LINKED ISSUE #943: Footer appears to be bumping up list items on a `<ul>` that's split across pages

liZe:
Fix #943


liZe:
Tests added for `box-decoration-break: clone` work, and they are pretty complicated.

I’ll have to add a test for #943.

PR #956, opened 2019-10-03
Sanitize simplified line breaking

### LINKED ISSUE #953: IndexError: tuple index out of range with weasyprint 50
After upgrading to 50 from 49 I got the following error when generating a PDF:

```
IndexError: tuple index out of range
[...]
  File "weasyprint/layout/blocks.py", line 63, in block_level_layout
    page_is_empty, absolute_boxes, fixed_boxes, adjoining_margins)
  File "weasyprint/layout/blocks.py", line 77, in block_level_layout_switch
    page_is_empty, absolute_boxes, fixed_boxes, adjoining_margins)
  File "weasyprint/layout/blocks.py", line 130, in block_box_layout
    absolute_boxes, fixed_boxes, adjoining_margins)
  File "weasyprint/layout/blocks.py", line 376, in block_container_layout
    for line, resume_at in lines_iterator:
  File "weasyprint/layout/inlines.py", line 53, in iter_line_boxes
    absolute_boxes, fixed_boxes, first_letter_style)
  File "weasyprint/layout/inlines.py", line 70, in get_next_linebox
    skip_stack = skip_first_whitespace(linebox, skip_stack)
  File "weasyprint/layout/inlines.py", line 210, in skip_first_whitespace
    result = skip_first_whitespace(box.children[index], next_skip_stack)

```

Downgrading to 49 prevents this error.
Possibly similar to #783? I will try to come up with a test case when I find some time.

Tontyna:
The special case that an already broken child must be split again confronts us with two skip stacks: The initial (absolute) and the adjusted (partial) stack. Often the new stack is a relative one and (carefully) adding both stacks is ok. But the following must be observed:

- The first number of the combined stack has to be the index of the child  we broke twice.
- A child index bigger than the starting index of the initial stack denotes  that the new skip stack is an absolute stack. Don't add nothing.

Should fix #953

### DIFF
--- a/weasyprint/layout/inlines.py
+++ b/weasyprint/layout/inlines.py
@@ -858,11 +858,27 @@ def split_inline_box(context, box, position_x, max_x, skip_stack,
                             # add the original skip stack to the partial
                             # skip stack we get after the new rendering.
 
-                            # We have to do:
-                            # resume_at + initial_skip_stack
-                            # but adding skip stacks is a bit complicated
-                            current_skip_stack = initial_skip_stack
-                            current_resume_at = (child_index, child_resume_at)
+                            # Combining skip stacks is a bit complicated
+                            # We have to:
+                            # - set `child_index` as the first number
+                            # - append the new stack if it's an absolute one
+                            # - otherwise append the combined stacks
+                            #   (resume_at + initial_skip_stack)
+
+                            # extract the initial index
+                            if initial_skip_stack is None:
+                                current_skip_stack = None
+                                initial_index = 0
+                            else:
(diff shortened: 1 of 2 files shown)


Tontyna:
That's what I came up after a lot of dirty debugging -- eventually I gained a grasp of how those  recursive skip stacks (skipping back and forth) are meant to work. 

What confused me a lot is the mixture of child indices with text positions. And of course there are the two trees of boxes. Skip stack pointing to the original unsplit raw formatted tree, children being copies containing adjusted text fragments. It occurred to me that I wanted them to have a pointer to their origin, or to their parent box...

PR #959, opened 2019-10-04
Line breaking tests

grewn0uille:


Tontyna:
@grewn0uille the regression tests never enter the critical code part -- the width of `10.5ch` is appropriate with `monospace` font-family. For `ahem` the width should be `10em`.

Tontyna:
@grewn0uille thanks for taking the trouble of writing the tests :1st_place_medal:

Tontyna:
Merged your branch into #956 and changed the units... whats up with `ahem`? On Windows `test_breaking_linebox_regression_12()` passes, but not on Linux...

Tontyna:
@grewn0uille what's the use of checking the `position_y` in `test_breaking_linebox_regression_12()`? Reveals that without a `line-height: 1` on Windows an extra pixel is added while on Linux the line height defaults to 1em/16 pixel. And why did you expect 17?

grewn0uille:
> @grewn0uille what's the use of checking the position_y in test_breaking_linebox_regression_12()? Reveals that without a line-height: 1 on Windows an extra pixel is added while on Linux the line height defaults to 1em/16 pixel. And why did you expect 17?

That was totally useless, I removed it (and also the line-height as the position_y is not checked anymore).
It was like for checking the number of lines was correct, but it's checked when unpacking p children. So that was stupid.

Tontyna:
Done.

Issue #970, opened 2019-10-15
API-Question: Removing/suppressing elements from render output?


JohannesMunk:
Hey Weasyprint-Crew and Community,

I needed to hide/remove elements from the generated output depending wheter or not they are on an even or odd page. Selecting elements with the @page :left selector is explicitely excluded by the specs.... so I hacked together a workaround.

It only works for static / non-layout relevant changes. In my case some (formatted) text box had to be on the left on left pages and on the right side on right pages. What I do now, is just insert it twice, and let the code below hide the unwanted ones.

I adapted the main.py to render and output separately. In between all elements that are "tagged" (by @string_set = "onlyonright" or "onlyonleft") and do not meet their page modulo are set to zero opacity.

There must be nicer ways to do this!

* Easier tagging than string_set? I didnt know which attributes Weasyprint would let pass through the render step, so I took smth that was there. Another hack would be to use "magic" z_values. Or smth else?

* Can the target box be removed from the childrens tuple in order to really remove the box? My first try (commented) didnt work, and as my python is non-existent I stopped there. The opacity trick of course means that the text is still contained in the document. Other ways to achieve this?

Here comes my code.

```
___main__.py:
.....

def findAndApplyOnlyOn(pageNo, pbox):
    if not hasattr(pbox, "children"): return
    for i in range(len(pbox.children) - 1, -1, -1):
        box = pbox.children[i]
        if isinstance(box.string_set, list):
            if len(box.string_set) > 0:
                if isinstance(box.string_set[0], tuple):
                    if (box.string_set[0][0] == 'onlyonright' and pageNo % 2 == 1) or (box.string_set[0][0] == 'onlyonleft' and pageNo % 2 == 0):
                        # pbox.children = pbox.children[ : i ] + pbox.children[i+1 : ]
                        box.style['visibility'] = 'hidden'
                        box.style['opacity'] = 0
                        continue

        findAndApplyOnlyOn(pageNo, box)

def main(argv=None, stdout=None, stdin=None):
    """The ``weasyprint`` program takes at least two arguments:
.....

    if args.output == '-':
        if stdout is None:
            stdout = sys.stdout
        # stdout.buffer on Py3, stdout on Py2
        output = getattr(stdout, 'buffer', stdout)
    else:
        output = args.output

    kwargsRender = {
        'stylesheets': args.stylesheet,
        'presentational_hints': args.presentational_hints}
    kwargs = {}
    if args.resolution:
        if format_ == 'png':
[...]

Issue #434, opened 2017-02-21
Find a way to easily reach the internal values of the rendered tree


G-E-N-E:
I'm not sure if this is a bug or intended behavior but the height of a box is 'auto' after rendering unlike the width of the box which appears to be being calculated based on the content.

How can I get the calculated height of the element id=header from the rendered document, i.e. 
doc = HTML(string=html_below).render(stylesheet=[CSS(css_below)])?

It seems like the width of the elements is calculated but the height remains set to 'auto'.
I'd like to resolve this to a unit so unfixed elements (not shown) can be positions relative to the dynamic height.

# HTML

```html
<div id="header">
  <div id="header-right">
    <div>Row 1, Column 2</div>
    <div>Row 2, Column 2</div>
  </div>

  <div>
    <div>Row 1, Column 1</div>
    <div>Row 2, Column 1</div>
  </div>
</div>
```

# CSS

```css
#header {
  position: fixed;
  top: 0;
  left: 0;
  right: 0;
}

#header-right {
  float: right;
  text-align: right;
}
```

liZe:
There's no "official" and reliable way to get the boxes and their attributes. But with the current version, you can try to get what you want:

```
>>> doc.pages[0]._page_box.children[0].children[0].children[0]
<Placeholder <BlockBox div None>>
>>> doc.pages[0]._page_box.children[0].children[0].children[0].style['position']
'fixed'
>>> doc.pages[0]._page_box.children[0].children[0].children[0].height
37.28125
```

AFAIK, there's currently no easy way to automatically get what you want (for example, get the blocks corresponding to your tags using their id), but it may be possible to override some classes and functions if you really need it.

G-E-N-E:
This is the same conclusion I've come too.
I'm in the progress of defining some access functions around id's and element types; I'll post what I come up with in case anyone is interested.

liZe:
> I'm in the progress of defining some access functions around id's and element types; I'll post what I come up with in case anyone is interested.

Cool!

liZe:
I think that 4365b19 gives a nice way to enable post-processing operations. Feel free to reopen if needed!

PR #972, opened 2019-10-17
Add html element to Box class

Tontyna:
Unless the Box is a PageBox or a MarginBox the new attribute `element` points to the html element (xml.etree.ElementTree.Element) where the Box stems from.
This enables e.g. post-processing that depends on Element attributes, like 'class' or 'id'.

Caution: Elements with no subelements will test as False! Test the element's existence with `if box.element is not None:`.

Related to #434, #970.

### DIFF
--- a/weasyprint/formatting_structure/boxes.py
+++ b/weasyprint/formatting_structure/boxes.py
@@ -89,8 +89,9 @@ class Box(object):
     def all_children(self):
         return ()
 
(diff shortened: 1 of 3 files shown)


liZe:
> Related to #434, #970.

It gives a lot of post-processing possibilities, it doesn't change anything for memory consumption, let's merge!

PR #975, opened 2019-10-22
Return appropriate tuple for string-set:none

Tontyna:
This snippet used to crash WeasyPrint with `UnboundLocalError: local variable 'computed_value' referenced before assignment`:

```html
<style>
div {
  string-set: none;
}
</style>
<div>OMG</div>
```

Reason being the tuple-constructor in `computed_values.string_set()` splitting `'none'` instead of a tuple consiting of a string-name and its string-value. 

BTW: WeasyPrint's unconventional way of handling `string-set` enables declarations like `string-set: none, stringname "value";`.


liZe:
Of course, thank you

Issue #974, opened 2019-10-21
Improper css var() causes crash


sureshvv:
Getting this error when I process this html file
[yy1.txt](https://github.com/Kozea/WeasyPrint/files/3752440/yy1.txt)

Tontyna:
That's a crash with announcement -- do you see the WARNING ``Unsupported computed value `calc(var(--line-height) / var(--line-height-ratio))` set in variable `--font-size` for property `font-size`.``? 

That's because `calc()` isn't implemented in WeasyPrint. Although this shouldn't crash WeasyPrint. 

And indeed, it's not the unsupported `calc`, that crashes WeasyPrint, but the fallback value -- WeasyPrint's `compute` function detects that it cannot replace the `var(--fontsize)` and falls back to the font-size of the parent. Which happens to be a unit-less simple integer or float.
And crashes in the `font_size()` function. We could make `font_size` aware for simple numbers, but then the `length()` function will crash.

@liZe - this fallback to a unit-less parent property is fatal for **all** css properties that designate a length, didn't count the `COMPUTER_FUNCTIONS` that call the `length()` method and the css properties that directly call `length()`...

@sureshvv - even if WeasyPrint was able to calculate your `--line-height` value it would crash because the `calc()` in your css doesn't supply a unit, just plain numbers.

sureshvv:
Browsers seem to handle it well. The html file displays perfectly in all the browsers that I have tested on. wkhtmltopdf also handles the same html file but not as well ;)

Tontyna:
> Browsers seem to handle it well.

Yes, they dont crash. My Browser **ignores** the calculated dimension-less values and falls back to the initial or inherited value.

Tontyna:
> this fallback to a unit-less parent property is fatal for all css properties ...

That's not true. Almost all of the 'lengthy' properties get a sane value from `INITIAL_VALUES` and/or the responsible `@register_computer` function handles the case when this value is an integer.

Looks like only `font-size` and `word-spacing` are affected by the unit-less fallback crash.

Tontyna:
> Looks like only font-size and word-spacing are affected by the unit-less fallback crash.

Minimal example to crash WeasyPrint by `var()`:

```html
<style>
div {
  --value: invalidValue;
  border-spacing: var(--value);
}
</style>
<div>crash</div>
```

The complete list of properties that kill WeasyPrint when defined via an invalid `var()` value is:

 - word-spacing
 - font-size
[...]

PR #977, opened 2019-10-23
Dont crash due to invalid var() values

Tontyna:
When a css property is defined via the var() function and the variable contains an invalid value a WARNING is generated and either an initial or the parent-value is assigned. 

In most cases the fallback is an appropriate value and computation succeeds without further intervention, but for some properties this is not the case. We could adjust `css.properties.INITIAL_VALUES`, but I suppose there are reasons for the *improper* values, probably they are required elsewhere.

That means we must take care for those special fallback values in the corresponding @register_computer-function. Like in `border_width()` -- this function pays attention since fadb60c14fc3fe7c7cb790ca505488ff8a8467b3, 8 years ago.

### DIFF
--- a/weasyprint/css/computed_values.py
+++ b/weasyprint/css/computed_values.py
@@ -737,6 +737,10 @@ def word_spacing(computer, name, value):
     """Compute the ``word-spacing`` property."""
     if value == 'normal':
         return 0
+    elif isinstance(value, int):
+        # The initial value can get here, but length() would fail as
+        # it does not have a 'unit' attribute.
+        return value
     else:
         return length(computer, name, value, pixels_only=True)
 
--- a/weasyprint/css/computed_values.py
+++ b/weasyprint/css/computed_values.py
@@ -20,8 +20,8 @@ from ..logger import LOGGER
 from ..urls import get_link_attribute
 from .properties import INHERITED, INITIAL_VALUES, Dimension
 from .utils import (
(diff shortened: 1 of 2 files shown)


Tontyna:
af7817163a073236294c6d7b139899d49d324163 is related to #974: @sureshvv's cutestrap style sheet won't crash any more.
Though it won't render as expected since the css is calc()ulating a lot and WeasyPrint doesn't (yet) support calc.

Tontyna:
Now all previously crashing fallback values -- see the list in https://github.com/Kozea/WeasyPrint/issues/974#issuecomment-544907390 -- don't crash anymore.

Tontyna:
Was tempted to pass those *insane* fallback values through the proven validation functions of Step 2 and apply the @register_computer-function afterwards. But my first candidate was `font-size` ... and I had to abandon the ~~good~~ lazy idea.

Nevertheless it's nagging my brain that the validation must be done twofold. Yes, I know: specified and computed are different entities, but nevertheless...

Tontyna:
WOW! My first unit-test! There you go :smirk:

Since all `KNOWN_PROPERTIES` are tested with an improper value, this should be future-proof.

Be aware that if the INITIAL_VALUES are altered (by adding new properties or changing a default value) you mustn't only adapt the basic validation functions for Step 2 but also prepare the corresponding @register_computer-function for those 'raw' default values.

liZe:
Shamefully forgotten by my tired eyes, half cherry-picked and half replaced by a1df1de, largely discussed in #974, delicately enlightened by @Tontyna’s first unit test…

This pull request has had an incredible life.

Issue #61, opened 2013-03-22
Form/Form Fields


schacki:
Hi all,
it would be really helpful to me, if any form input fields would be transformed to fillable input fields in the PDF document. To me it seems like this feature would be missing. Is that correct? If so, are there any plans to add this feature?
Regards,
Juergen

MarZab:
And while someone is doing that I would like to have Digital Signature Fields too. :)

djit:
+1000 for Digital Signature: this will be the killer feature for me.

liZe:
> Does anyone feel like reading the PDF spec to figure this out?

Anyone? :wink:

hicham-elmansouri:
Hi All,
Any update on this? I can try if it's strill pending.

liZe:
Hi @hicham-elmansouri,

Sorry, there's no update on this. But it's probably no that hard to fix it, even if it may be a bit long and painful.

Here are the steps if you're interested in fixing this issue:
- Add a method gathering the form fields, just like it's done for anchors in `Document.resolve_links` (in `document.py`).
- Call this method in `pdf.py` to build the list of the fields positions (in `prepare_metadata`).
- Read the [PDF documentation](https://www.adobe.com/content/dam/Adobe/en/devnet/acrobat/pdfs/pdf_reference_1-7.pdf) and find how we can create form fields (probably in 8.6).
- Add the form fields in the PDF (in `pdf.py`, in the `write_pdf_metadata` function).

Good luck!

liZe:
If we use [pdfrw](https://github.com/Kozea/WeasyPrint/tree/pdfrw), the 4th point that was the hard part is no really easier.

Anyone interested?

rdwrt:
any updates?

liZe:
> any updates?

No. If anyone is interested, [my previous comment](https://github.com/Kozea/WeasyPrint/issues/61#issuecomment-301051989) is the way to go!

liZe:
> I might give it a shot. Should I still follow [this](https://github.com/Kozea/WeasyPrint/issues/61#issuecomment-301051989)?

Here are the new steps:
- Add a method gathering the form fields, just like it's done for anchors in `Document.resolve_links` (in `document.py`).
- Call this method at the end of `Document.write_pdf` to build the list of the fields positions.
[...]

PR #985, opened 2019-10-30
Implement form fields

Afoucaul:
This is a proposal to implement support for form fields in PDF generated by WeasyPrint, which would solve #61.
Here is how this implementation works.

# Input markup
*The `wf` prefix used thoughout the implementation originally stands for 'weasyform', that is the first name I came up with for this feature*

Form fields must be described by `<div>` tags with specific classes.
As for now, this PR includes two basic classes: `wf-text` for simple single-line text entries, and `wf-textarea` for mult-iline text areas.
Additionally, the internal mechanism requires such `<div>` tags to have a `wfid` attribute, that is unique throughout the document.
There is no requirement to wrap these tags in a `<form>` parent tag as is usually done in HTML.

In essence, nothing more than this is required to create a text entry:

```html
<div class='wf-text' wfid='input'></div>
```

This will output the following document; the black rectangle is indeed an editable field.

2019-10-30-114929_642x906_scrot

# Internal mechanism

The workflow is inspired by what @liZe describes in #61.

## Markup augmentation
The markup is searched for tags with class `wf-text` or `wf-textarea`, and adds anchors into these.
The markup after augmentation looks like this:

```html
<div class='wf-text' wfid='input'>
    <div class='wf-top-left' id='WF-wf-text-input-topleft'></div>
    <div class='wf-bottom-right' id='WF-wf-text-input-bottomright'></div>
</div>
```

These inner anchors are positioned absolutely within the field, one at the top left corner and the other at the bottom right corner.
This is used later on to fetch the expected position of the field and insert a PDF form field object there.
To achieve this position, some CSS is needed, which is described in a [`html5_wf.css`](https://github.com/Afoucaul/WeasyPrint/blob/armand/forms/weasyprint/css/html5_wf.css) stylesheet I created next to other already existing stylesheets.

The markup augmentation is performed in the [`HTML.__init__`](https://github.com/Afoucaul/WeasyPrint/blob/armand/forms/weasyprint/__init__.py#L123) method, that calls a function named [`augment_markup`](https://github.com/Afoucaul/WeasyPrint/blob/armand/forms/weasyprint/forms.py#L95) defined in the `forms` module.

## PDF manipulation

The next step is to parse the PDF generated, and insert into it the form fields described by the markup.
First, the rectangle of the input fields is read from the PDF source, by reading PDF anchors created from the `id='WF-...'` attribute of the augmented markup.
Based on the coordinates of the `-topleft` and the `-bottomright` anchors, the rectangle can be easily retrieved.

Then, using the utilities from the `weasyprint.pdf` module, PDF objects are created for the fields and the acroform that will contain them.
These objects are created from templates defined in `forms.py`:

```python
ACROFORM = """
<<
    /DA (/Times 10 Tf 0 g)
    /DR <<
      /Font <<
        /Times <<
          /BaseFont /Times-Roman
          /Subtype /Type1
          /Type /Font
        >>
      >>
    >>
    /Fields [{fields}]
>>
"""

WF_CLASSES = {
    'wf-text': """
<<
    /BS <<
        /S /S
        /W 1
    >>
    /FT /Tx
    /Rect [{rect}]
    /Subtype /Widget
    /Type /Annot
    /V ()
>>
""",
    'wf-textarea': """
<<
    /BS <<
        /S /S
        /W 1
    >>
    /Ff 4096
    /FT /Tx
    /Rect [{rect}]
    /Subtype /Widget
    /Type /Annot
    /V ()
>>
"""
}
```

Lastly, the xref table and trailer are updated.

All these manipulation is performed in a new function named [`pdf.write_pdf_form`](https://github.com/Afoucaul/WeasyPrint/blob/armand/forms/weasyprint/pdf.py#L611).

# Example

Here is a sample input, and the PDF produced:

```HTML
Name:
<div class='wf-text' wfid='name'></div>

Comment:
<div class='wf-textarea' wfid='comment'></div>
```

Output:
2019-10-30-122902_414x304_scrot

Actual PDF file: [test.pdf](https://github.com/Kozea/WeasyPrint/files/3788477/test.pdf)

# To improve

The code currently is in a working state, and leaves rooms for further development.
There are three main points that are left to improve:

1. The form fields currently seem to only support ASCII characters
2. The font used for field content is hard-coded in the PDF object templates and cannot be changed from CSS
3. Only single-line and multi-line text fields are implemented

---

I'll be happy to hear your thoughts about the way I implemented this.
I took many design decisions that might not be the best, and I'll be glad to work toward integrating this implementation into your vision.
Also, please do ask any question regarding the implementation if you're unsure why I did things this way.


liZe:
Thanks a lot for your pull request and sorry for being so long, I'll take the time to review your code as soon as I can!

liZe:
> I'll be happy to hear your thoughts about the way I implemented this.
> I took many design decisions that might not be the best, and I'll be glad to work toward integrating this implementation into your vision.
> Also, please do ask any question regarding the implementation if you're unsure why I did things this way.

This pull request is full of very useful code and really clear explanation, thanks a lot!

I will merge this PR in a branch, and change a couple of things on it:
- Instead of relying on divs with special classes, we can render PDF inputs from real form inputs, using a global CLI/API option. Having a minimum of custom mechanisms is one of the main goals of WeasyPrint :wink:.
- I think that we can use the boxes position instead of dealing with extra CSS. That’s a bit harder, but I think that we have everything in the layout for that (and #972 may be useful).

That’s a really nice feature to have, thanks again for your work. I hope that I’ll merge it in `master` branch soon!

liZe:
Included in #1018.

Afoucaul:
Thanks @liZe for your feedbacks! I'm glad this PR becomes helpful.
I'll follow closely the new PR you opened, let me know if you'd like some help regarding any of the issues!

PR #989, opened 2019-11-07
require Pyphen >= 0.9.1

FelixSchwarz:
With older versions tests I get 19 test failures in test_text.py. I am not sure if this is peculiarity of the test suite but 0.9.1 was released almost 6 years ago (December 2013) so requiring it should be fine.

Side note: ~~Fedora still ships 0.9.0~~ Fedora 30 ships 0.9.1 but somehow that version also triggers the same test failures. Anyway 0.9.0 is definitively not good enough to get the test suite passing.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -13,7 +13,7 @@ WeasyPrint |version| depends on:
 * tinycss2_ ≥ 1.0.0
 * cssselect2_ ≥ 0.1
 * CairoSVG_ ≥ 2.4.0
-* Pyphen_ ≥ 0.8
+* Pyphen_ ≥ 0.9.1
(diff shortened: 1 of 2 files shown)


liZe:
> With older versions tests I get 19 test failures in test_text.py. I am not sure if this is peculiarity of the test suite but 0.9.1 was released almost 6 years ago (December 2013) so requiring it should be fine.

You're right. 0.9.0 is broken, that's why tests don't pass. They pass with 0.8, but 0.9.1 is old enough to merge your pull request.

FelixSchwarz:
Thank you.

Btw: I noticed that tests fail with Fedora because the maintainer decided to use system-provided dicts instead of the ones shipped with pyphen. I'll try to change that in the future.

Issue #954, opened 2019-09-27
Generated PDF differs slightly from sample on weasyprint site


espechtcfe:
When generating a PDF of the "Report" sample using Weasyprint 50, I'm seeing a difference in the PDF when compared to the sample on the weasyprint site.

I've attached the PDF that is being generated.  If you compare the typography features page (page 8) with the version on the weasyprint site, you'll notice that in the one I generated, the copy in the Ligatures and Numbers sections includes line breaks that don't exist in the version on the website.

I've installed Weasyprint 50 using this docker image - https://hub.docker.com/r/minidocks/weasyprint - which I modified by changing the Dockerfile to specify version 50.  I have verified that version 50 is running in the container.  The command I'm running from a folder above the report folder to generate the report is:
docker run --rm -v `pwd`:/app -w /app minidocks/weasyprint:latest ./report/report.html weasyprint_test.pdf

Thank you for looking into this.  I'm just getting started with Weasyprint and am very impressed with the product!
[weasyprint_test.pdf](https://github.com/Kozea/WeasyPrint/files/3663316/weasyprint_test.pdf)

liZe:
There's also a difference in the small caps text. I have to find what's wrong.

> Thank you for looking into this. I'm just getting started with Weasyprint and am very impressed with the product!

:smile:

knixeur:
Ok, that was hard to find.
There's a problem with font-variants that was introduced in v44 (v43 works fine).
The problem is that a new Pango attribute list is being generated and replacing the current attributes.

https://github.com/Kozea/WeasyPrint/blob/master/weasyprint/text.py#L769
https://github.com/Kozea/WeasyPrint/blob/master/weasyprint/text.py#L783

Replaces the attributes generated on setup https://github.com/Kozea/WeasyPrint/blob/master/weasyprint/text.py#L728

knixeur:
Ok, I made a PR but doesn't fix Ligatures and Numbers, it does fix small-caps example. If I open report.html in a browser I see break lines in Ligatures and Numbers examples.

knixeur:
Line breaks in Ligatures and Numbers "broke" between v45 and v46

liZe:
@knixeur Thanks a lot, your comments (and your PR) were really, really useful :heart:.

PR #998, opened 2019-11-26
When changing Layout's attribute in set_text, do not override them

knixeur:
Between tags v43 and v44 there was a refactor in class Layout. Settings the layout attributes was done in one place and after the refactor it's done during setup() and somethings during set_text() call.
set_text's is overriding the attributes set by setup

### DIFF
--- a/weasyprint/text.py
+++ b/weasyprint/text.py
@@ -147,6 +147,7 @@ ffi.cdef('''
     void g_type_init (void);
 
     void pango_layout_set_width (PangoLayout *layout, int width);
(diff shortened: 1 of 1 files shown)


knixeur:
Mmm maybe setup() should also update attributes instead of setting new ones since it's called from reactivate. I'm open to suggestions.
Also I didn't know how to test this without using mock/patch, let me know if there's some test I could use as base to test it.

liZe:
Thank you!

liZe:
Related to #954.

PR #1018, opened 2019-12-31
Implement form fields

### LINKED ISSUE #61: Form/Form Fields

liZe:
Includes commits from @Afoucaul’s #985, fixes #61.

We have a lot of work:
- [ ] Allow users to type something else than ASCII characters.
- [ ] Use the real font used for the input rendering.
- [ ] Allow users not to include PDF inputs, using a CLI/API option.
- [ ] Handle more input types (and don’t crash when meeting an unknown one).
- [ ] Add tests (of course!).


liZe:
The ASCII limitation and the font are probably the same problem.

liZe:
I’m not sure that I can find a simple solution without using something different from Cairo :/.

Afoucaul:
You mean, regarding the ASCII characters and the font face?

liZe:
> You mean, regarding the ASCII characters and the font face?

Yes. We probably would have to change the way fonts are included in the PDF, and I suppose that it’s really hard to do this only with a PDF post-processor.

liZe:
This feature will be added without Cairo (see #1232). A lot of code will need to be changed, I’m closing this issue, but we’ll for sure get some code from this PR :wink:.

Issue #1023, opened 2020-01-09
unexpected justification


tomek1024:
In very seldom cases (combination of margin values, text, text settings and fonts) weasyprint incorrectly breaks the line. I attach the source as well the result. I use manual hyphenation.

html:
```html
<!doctype html>
<html lang='pl'>
  <head>
    <meta charset="UTF-8">
    <link href='style.css' rel='stylesheet' type='text/css'>
  </head>
  <body>
    <p><strong>LIS&shy;BETH SALAN&shy;DER</strong> &amp; Aga wyszła spod prysz&shy;nica.</p>
  </body>
</html>
```

CSS:
```css
@font-face {
  font-family: "BookmanOld";
  src: url("https://zecer.elibri.com.pl/lte/fonts/BookmanOld/BookmanOld_R.ttf");
  font-weight: normal;
  font-style: normal;
}

@font-face {
  font-family: "BookmanOld";
  src: url("https://zecer.elibri.com.pl/lte/fonts/BookmanOld/BookmanOld_RB.ttf");
  font-weight: bold;
  font-style: normal;
}


@page {
  size: 160mm 235mm;
  margin-left: 21mm;
  margin-right: 21mm;
}


html { hyphens: manual; }

body {
  margin: 0;
  font-size: 14pt;
  font-family: BookmanOld;
}

p {
  text-align: justify;
  text-indent: 7mm;
  letter-spacing: -0.02em;
}
```

[test.pdf](https://github.com/Kozea/WeasyPrint/files/4042385/test.pdf)

liZe:
Hello, and thank you for this issue.

I can’t reproduce the bug :unamused:. Could you please give the output of `weasyprint -i`?

tomek1024:
Sure:

```
System: Linux
Machine: x86_64
Version: #40~18.04.1-Ubuntu SMP Thu Nov 14 12:06:39 UTC 2019
Release: 5.0.0-37-generic

WeasyPrint version: 50
Python version: 3.6.9
Cairo version: 11510
Pango version: 14014
CairoSVG version: 2.4.2
```

Tontyna:
@liZe too bad that `weasyprint -i` fails on Windows due to the availability of [`os.uname()`](https://docs.python.org/3/library/os.html#os.uname): recent flavors of Unix.

liZe:
[...]

PR #1044, opened 2020-01-29
Use platform.uname instead of os.uname

Tontyna:
`platform.uname()` should work on all platforms (cf. https://github.com/Kozea/WeasyPrint/issues/1023#issuecomment-573250759) and therefore `weasyprint --info`, too.

Decided to `import platform` locally to avoid unnecessary initialization when there is no `--info` argument.

It's just educated guessing that it reveals interesting facts on Windows -- I don't have no Windows computer anymore.

Have no idea how to write a unit test for `PrintInfo` action :flushed:

### DIFF
--- a/weasyprint/__main__.py
+++ b/weasyprint/__main__.py
@@ -11,7 +11,6 @@
 
 import argparse
 import logging
-import os
 import sys
 
 import cairosvg
@@ -22,8 +21,9 @@ from .text import cairo, pango
 
 class PrintInfo(argparse.Action):
     def __call__(*_, **__):
-        uname = os.uname()
-        print('System:', uname.sysname)
+        import platform
+        uname = platform.uname()
+        print('System:', uname.system)
         print('Machine:', uname.machine)
         print('Version:', uname.version)
         print('Release:', uname.release)


liZe:
Hello!

I had actually changed this in the `clean` branch too, in 10a7863. I really should have merged it sooner… And there’s also a test in this commit, at least checking that it doesn’t crash as it used to do on Windows :smile:.

Tontyna:
Ah so simple, just `_run('--info')`

liZe:
I’ve merged the `clean` branch.

> It's just educated guessing that it reveals interesting facts on Windows -- I don't have no Windows computer anymore.

Oo

Tontyna:
> Oo

I know, [Linux sucks](https://itvision.altervista.org/why.linux.is.not.ready.for.the.desktop.current.html), but [Windows is an abomination](https://itvision.altervista.org/why-windows-10-sucks.html).

liZe:
> I know, [Linux sucks](https://itvision.altervista.org/why.linux.is.not.ready.for.the.desktop.current.html), but [Windows is an abomination](https://itvision.altervista.org/why-windows-10-sucks.html).

That was interesting reading!

PR #1061, opened 2020-02-14
Update instructions for install on Windows

mPyth:
In a try to run command "python -m weasyprint http://weasyprint.org weasyprint.pdf" I have encountered on problem "Entry Point Not Found" with following text: "The procedure entry point inflateValidate could not be located in the dynamic link library C:\Program Files\GTK3-Runtime Win64\bin\libpng16-16.dll".
After 90-120 mins I found the problem - there were two "zlib1.dll" files - the second came from PostgreSQL (v10) and because it was first in the PATH it was called instead of GTK's version.
With this small change someone else could save the time and potential new tickets are prevented.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -448,6 +448,7 @@ Let's find out. Enter the following command:
 .. code-block:: console
 
     WHERE libcairo-2.dll
+    WHERE zlib1.dll
 
 This should respond with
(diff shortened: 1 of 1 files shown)


liZe:
Thank you!

This problem may virtually happen with any DLL provided by the GTK installer. This Windows let’s-install-one-version-of-each-library-for-each-application strategy is sometimes bad.

One day I’ll sell Windows and macOS binaries with everything bundled. Anyone interested? :smile:

Issue #106, opened 2013-07-04
RTL support


jamma313:
Hi,

What is the current status of weasyprint concerning rtl support ?
It looks like it has a partial support (can display text but not boxes), is this a long way to finish it ?

Thanks.

bahmany:
Hi, RTL support, Please

Thanks

doctoromer:
Hey, any rtl support is planned soon?

liZe:
> Hey, any rtl support is planned soon?

Adding RTL and BiDi support will be possible after rewriting the line breaking algorithm (#301), which is explained in [Fragmentation Module](https://www.w3.org/TR/css-break-3/#breaking-rules).

**That's huge.**

Secret breaking news: implementing the breaking rules and adding rtl/bidi support is on my short list for 2018. It'll come with dropping Python 2 support and releasing 1.0.0. Maybe.

doctoromer:
Thank you very much! I will look forward to it.

doctoromer:
thank you very much!

בתאריך יום ו׳, 23 במרץ 2018, 18:34, מאת mustafa0x ‏<notifications@github.com

liZe:
> According to some brief tests with 0.42, RTL/BIDI supports seems to be pretty decent.

RTL suppot is not that bad for simple use cases and will probably be quite good in version 43.

Bidi is a much harder goal to reach. We'll have to open a separate issue when this one is closed.

mohdnajdy:
@liZe Is there a way we can maybe chat or you can give me some guides on how to contribute to the RTL support ?

dumblob:
@mohdnajdy did you have any success in adding support for RTL to weasyprint?

mohdnajdy:
@dumblob not really. I am still lost on where to contribute at for supporting RTL

liZe:
> I am still lost on where to contribute at for supporting RTL

It’s really hard to know where to start. `git grep rtl` will give a few TODOs, but that’s a small part of the work needed.

Some parts are probably easier to fix: tables, list markers, flex blocks… The hard work is of course in inline blocks layout.

You can find failing tests related to RTL in [WeasySuite](http://test.weasyprint.org/suite-css21/results/) (this page may take a long time to render). And, of course, we’ll need to write our own tests too.

[...]

PR #1070, opened 2020-02-25
RTL: render table correctly when direction is rtl

mohdnajdy:
This pull request should fix the table rendering in RTL.which is part of the issue: #106


mohdnajdy:
@liZe Do you think this is a proper solution ?

liZe:
Included in #1079.

PR #1072, opened 2020-02-26
Drop extraneous font matching

gentooboontoo:


liZe:
I suppose that it’s harmless, thanks!

PR #1099, opened 2020-04-10
added function for port specification

eyseman:
Added a new function which allows the user to select a specific port. If no port is specified, port 5000 is used as default port.


liZe:
That’s a good idea, thank you!

I’ll fix a couple of things (including lint and indentation errors), but I’ll keep your idea :smile:.

PR #1116, opened 2020-05-16
Add missing validation package to setup.cfg

balazsdukai:
The `weasyprint.css.validation` package was missing from the setup, which was causing a `ModuleNotFoundError: No module named 'weasyprint.css.validation'` when trying to run weasyprint that was installed directly from the `master` branch.


liZe:
Thank you!

PR #1141, opened 2020-06-14
Dont clip page margins on account of body overflow

### LINKED ISSUE #35: Page margins are broken with "body { overflow: auto }" style
I have a CSS file that defines headers and footers using @top-center, @bottom-center, and @top-right. When I use weasyprint the resulting PDF does not have the headers and footers. Here's a snippet from the CSS file:

```
@page
{
    size: letter;
    margin: 27mm 16mm 27mm 16mm;
    counter-increment: page;
    @bottom-center
    {
        content: counter(page);
        white-space: pre;
        color: grey;
        }
    @top-center
    {
        content: 'Strategic Planning Concept';
        color: grey;
        border-bottom: solid 1px grey;
    }
    @top-left
    {
        content: ' \a \a \a ';
        white-space: pre;
        color: grey;
        border-bottom: solid 1px grey;
    }
    @top-right
    {
        content: ' \a \a \a February 6, 2013';
        white-space: pre;
        color: grey;
        border-bottom: solid 1px grey;
    }
}
```

Tontyna:
If we want the `viewport_overflow` restriction being applied to the `<BlockBox html>` only, the instructions for `draw_stacking_context()` are:

1. never clip when a PageBox is rendered
2. do the clip when drawing a `<BlockBox html>` whose page's overflow isn't `visible`

Good thing is: 
We can rely upon the `with stacked(context)` statement to save and restore the context state automagically.

fixes #35

### DIFF
--- a/weasyprint/draw.py
+++ b/weasyprint/draw.py
@@ -184,6 +184,14 @@ def draw_stacking_context(context, stacking_context, enable_hinting):
     # See http://www.w3.org/TR/CSS2/zindex.html
     with stacked(context):
         box = stacking_context.box
+
+        # apply the viewport_overflow to the html box, see #35
+        if box.element_tag == 'html' and (
+                stacking_context.page.style['overflow'] != 'visible'):
+            rounded_box_path(context,
+                    stacking_context.page.rounded_padding_box())
+            context.clip()
+
         if box.is_absolutely_positioned() and box.style['clip']:
             top, right, bottom, left = box.style['clip']
             if top == 'auto':
@@ -223,7 +231,9 @@ def draw_stacking_context(context, stacking_context, enable_hinting):
                 context, stacking_context.page, box, enable_hinting)
 
         with stacked(context):
-            if box.style['overflow'] != 'visible':
+            # dont clip the PageBox, see #35
+            if box.style['overflow'] != 'visible' and not isinstance(
+                    box, boxes.PageBox):
                 # Only clip the content and the children:
                 # - the background is already clipped
                 # - the border must *not* be clipped
--- a/weasyprint/draw.py
+++ b/weasyprint/draw.py
@@ -186,7 +186,7 @@ def draw_stacking_context(context, stacking_context, enable_hinting):
         box = stacking_context.box
 
         # apply the viewport_overflow to the html box, see #35
-        if box.element_tag == 'html' and (
+        if box.is_for_root_element and (
                 stacking_context.page.style['overflow'] != 'visible'):
             rounded_box_path(
                 context,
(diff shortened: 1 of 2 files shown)


Tontyna:
Any idea how a unit-test would look like? Checking pixels of a `write_png`?

weasyprint/draw.py
```diff
@@ -223,7 +232,9 @@ def draw_stacking_context(context, stacking_context, enable_hinting):
                 context, stacking_context.page, box, enable_hinting)
 
         with stacked(context):
-            if box.style['overflow'] != 'visible':
+            # dont clip the PageBox, see #35
+            if box.style['overflow'] != 'visible' and not isinstance(
+                    box, boxes.PageBox):
```
liZe, line 237: :+1:

weasyprint/draw.py
```diff
@@ -184,6 +184,15 @@ def draw_stacking_context(context, stacking_context, enable_hinting):
     # See http://www.w3.org/TR/CSS2/zindex.html
     with stacked(context):
         box = stacking_context.box
+
+        # apply the viewport_overflow to the html box, see #35
+        if box.element_tag == 'html' and (
```
liZe, line 189: Shouldn’t we use `box.is_for_root_element` instead?

liZe:
> We can rely upon the `with stacked(context)` statement to save and restore the context state automagically.

That’s good news.

> Any idea how a unit-test would look like? Checking pixels of a `write_png`?

:+1:

weasyprint/draw.py
Tontyna: Wasn't aware that there is a `box.is_for_root_element`. 

Though, I always wondered about the  `# No root element` fallback in `build_formatting_structure()`. Never ever managed to construct an HTML/element_tree without a 'html' root node.

liZe: > Though, I always wondered about the `# No root element` fallback in `build_formatting_structure()`. Never ever managed to construct an HTML/element_tree without a 'html' root node.

Good question, I don’t know if html5lib can give a document with no root `html` tag. Coverage says that this code is tested, so I suppose that it’s useful :smile:.

By the way, it’s easier to have an `html` tag nested in another `html` tag, that’s another good reason to use `is_for_root_element`.

Tontyna: > have an html tag nested in another html tag

I tried that, too. Various scaring nestings of `html` and `body`. Creating multiple and/or nested `BlockBox body` is possible. 
But always ended up with one single `BlockBox html` root element.

liZe:
Thank you, thank you, thank you…

weasyprint/draw.py
liZe: :scream: 

https://github.com/Kozea/WeasyPrint/blob/f333cefc515b15b878b6555d150e0471456ec90e/weasyprint/tests/test_boxes.py#L1324-L1334

Issue #1145, opened 2020-06-17
Bookmarks repeat on every page


arikrupnik:
I'm formatting an XML file. I want to generate an index, but for some reason I'm getting an entry for every section on each page. What am I doing wrong? Using Weasyprint 51 and Python 3.7.5 on Linux.

```xml
<doc>
  <section heading="1976">
    <section heading="Cemetery">
      <scene>EXT. CEMETERY -- NIGHT</scene>
      Rederik, now 28, lies behind a GRAVESTONE. He has a beard now, but is growing thinner on the top. He squints through a tree branch he's holding in front of his face.
```
CSS snippet:

```css
section { bookmark-level: 1; bookmark-label: attr(heading); }
section > section { bookmark-level: 2; }
section > section > section { bookmark-level: 3; }
```
![Screenshot_2020-06-17_12-18-50](https://user-images.githubusercontent.com/1756197/84928901-bb2a9580-b094-11ea-8306-0e9d55365ae8.png)

Tontyna:
Don't think you're doing anything wrong. 

The level-1-section, wrapping everything, is present on every page that WeasyPrint generates -- try to give it a border or background color and you should see that decoration on each page.

Like those 'decorative' stylings, the `bookmark-*` properties of the section are present on each page, generating the superfluous entries in the PDF navigation.

Looks like WeasyPrint should stop propagating the bookmarks when spreading a wrapping box across pages.

arikrupnik:
Thank you for your response, @Tontyna. Is there a workaround i can use?

arikrupnik:
Maybe if I put the `bookmark-level` on `section:before`? I haven't been able to make it work, but maybe I'm missing how `:before` works in this context.

Tontyna:
Nah, no chance. Though the [spec says](https://www.w3.org/TR/css-gcpm-3/) that the `bookmark-*` properties apply to **all elements**, WeasyPrint doesn't apply them.

Only workaround at the moment is inserting (empty) dummy elements.

arikrupnik:
Thanks.

> Only workaround at the moment is inserting (empty) dummy elements.

I thought this was what `:before` did ;=)

Any sense if this may change in WeasyPrint in the foreseeable future?

Tontyna:
BTW: Not only multi-page-boxes generate multiple bookmarks. Multi-Line **Inline**-Boxes do it, too:

```html
<h1 style="display:inline; white-space:pre">
A bookmark
for every
line
[...]

PR #1146, opened 2020-06-18
Bookmark label

Tontyna:
Incited by #1145 I started a research about how to implement bookmark-labels for the `:before` and `:after` pseudo-elements and stumbled upon page counters and target counters not being updated in bookmark-labels at all.

Of course, nobody ever will use page or target counters in her bookmark-labels...

### DIFF
--- a/weasyprint/formatting_structure/build.py
+++ b/weasyprint/formatting_structure/build.py
@@ -244,6 +244,15 @@ def before_after_to_box(element, pseudo_type, state, style_for,
         target_collector, counter_style))
 
     box.children = children
+
+    # calculate the bookmark-label
+    if style['bookmark_label'] == 'none':
+        box.bookmark_label = ''
+    else:
+        _quote_depth, counter_values, _counter_scopes = state
+        compute_bookmark_label(
+            element, box, style['bookmark_label'], counter_values,
+            target_collector, counter_style)
     return [box]
 
 
--- a/weasyprint/layout/__init__.py
+++ b/weasyprint/layout/__init__.py
@@ -140,7 +140,7 @@ def layout_document(html, root_box, context, max_loops=8):
         if not reloop_content and not reloop_pages:
             break
 
-    # Calculate string-sets and bookmark-label containing page based counters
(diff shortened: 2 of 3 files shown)


Tontyna:
To actually generate a bookmark-label the pseudo-element must not only have a `bookmark-level` and a `bookmark-label`, but also have `content`, e.g.:

```css
.heading:before {
  content: '';
  bookmark-level: 1;
  bookmark-label: 'A Heading';
}
```

Tontyna:
@liZe Maybe there is a better way of removing the duplicates. Best thing would be not to create them at all in the first place, but I was unable to find that first place.

Tontyna:
There is something else not feeling quite right:

```css
div{
  bookmark-level: 1;
  bookmark-label: 'the main div';
}
div:before {
 content: 'before ';
 display:block;
 bookmark-level: 1;
 bookmark-label: 'before the div';
}
```

The `:before`-box is a child of the div's main BlockBox. Thus *the main div* precedes the *before the div* in the PDF outline. Both bookmarks target the same location in the PDF, the top left corner of the `:before`-box.

liZe:
> The `:before`-box is a child of the div's main BlockBox. Thus _the main div_ precedes the _before the div_ in the PDF outline. Both bookmarks target the same location in the PDF, the top left corner of the `:before`-box.

Having a pseudo-element called `:before` when it’s actually **inside** the box is the problem, and there’s nothing we can do about this :wink:.

liZe:
> @liZe Maybe there is a better way of removing the duplicates. Best thing would be not to create them at all in the first place, but I was unable to find that first place.

We have the same problem to remove the top border of a block on the second page when it’s split. There’s a `remove_decoration` mechanism to change the computed value for the second part of the block (and even the first one to remove the bottom border, but we don’t care for bookmarks). But it doesn’t work for repeating table headers and footers.

I don’t know if it’s possible to do something like that for bookmarks. Otherwise, I can just merge this pull request, it’s already really good.

Tontyna:
> Otherwise, I can just merge this pull request

Yes, please merge it (and close #1145)

Issue #898, opened 2019-07-12
The var() function doesn't work as part of border shorthand property


ham2anv:
Not sure if this applies to other properties, but I ran into this today in the border shorthand property. The following:

`border: 0.15rem solid var(--highlight, black);`

throws the error "WARNING: Ignored `border: 0.15rem solid var(--highlight, black)` at 8:5, invalid value.". Breaking the var() out into its own border-color property fixes the issue.

mikevoets:
The same seems to happen when concatenating `var` with other text:

E.g. where `--table-label: 'Table'`, the following declaration:

`content: 'var(--table-label) ' '. ';`

results in "Table", and not "Table 1."

mikevoets:
The bug described in my comment is unrelated to the OP.

@ham2anv @liZe the issue here is that a variable token is yet undetermined at `weasyprint/css/validation/expanders.py:241` so it will raise `InvalidValues` because it doesn't recognise what kind of declaration it is.

liZe:
Now covered by #1219.

Issue #357, opened 2016-09-13
Css function calc not work


Akay7:
Function calc of css not working :(

`/* property: calc(expression) */`
`width: calc(100% - 80px);`

https://developer.mozilla.org/en/docs/Web/CSS/calc

dennisvang:
For those interested, here's the nitty gritty: https://drafts.csswg.org/css-values-3/#calc-notation

whinee:
Please mention that `calc()` is yet to be implemented in the documentation. This might save devs that uses this project hours or even weeks of pain, debugging what might be wrong with their HTML/CSS. Thank you!

liZe:
Hi!

> This might save devs that uses this project hours or even weeks of pain, debugging what might be wrong with their HTML/CSS.

If you want to avoid "weeks of pain", you can read the logs: warnings are logged each time a property or a value is not supported.

> Please mention that `calc()` is yet to be implemented in the documentation.

You’re right, that would be very useful. Interested in opening a pull request? That would be a quite simple way to become a contributor!

whinee:
Hallo!

> If you want to avoid "weeks of pain", you can read the logs: warnings are logged each time a property or a value is not supported.

Turns out that not reading docs properly also causes this. Thanks for reminding me. Time to turn on logging.

> You’re right, that would be very useful. Interested in opening a pull request? That would be a quite simple way to become a contributor!

Contrary to my comment above, I love reading docs! Though I am not confident that I will be able to write it. I mean I can, just not properly. I guess I'll try???

Anyways, I have just re-read my comment, and I think it might have sounded rude, and I feel bad for it now, sorryy... I tried to make it short 'cause every issue / comment I write sounds childish...

liZe:
> Anyways, I have just re-read my comment, and I think it might have sounded rude, and I feel bad for it now, sorryy... I tried to make it short 'cause every issue / comment I write sounds childish...

No problem!

> Contrary to my comment above, I love reading docs! Though I am not confident that I will be able to write it. I mean I can, just not properly. I guess I'll try???

Adding a sentence here is sufficient: https://github.com/Kozea/WeasyPrint/blob/dc7c2cb18c174b39dfb869fa5820b13cdf4da7b2/docs/api_reference.rst?plain=1#L647

You can copy another line where a property is not supported. And don’t worry, it doesn’t have to be perfect, we can always enhance it afterwards!

grewn0uille:
Hello!

[...]

PR #1165, opened 2020-07-16
Add support for concatenating var() functions in 'content' declarations

mikevoets:
Concatenation of `var()` functions in content declarations does not work. Currently, a variable in content declaration gets resolved correctly, however it disregards the remaining value of a content declaration. E.g.:

```
body { --var: 'Paragraph'; counter-reset: p }
p { counter-increment: p }
p:before { content: var(--var) ': ' counter(p); }
```
resolves `p:before` to `Paragraph`, and not to as you would expect; `Paragraph: 1`.

This bug is fixed in the changes in this PR.

Linked to my comment in #898.

### DIFF
--- a/weasyprint/css/computed_values.py
+++ b/weasyprint/css/computed_values.py
@@ -176,26 +176,11 @@ def register_computer(name):
     return decorator
 
 
-def compute(element, pseudo_type, specified, computed, parent_style,
-            root_style, base_url, target_collector):
-    """Create a dict of computed values.
-
-    :param element: The HTML element these style apply to
-    :param pseudo_type: The type of pseudo-element, eg 'before', None
-    :param specified: A dict of specified values. Should contain
-                      values for all properties.
-    :param computed: A dict of already known computed values.
-                     Only contains some properties (or none).
-    :param parent_style: A dict of computed values of the parent
-                         element (should contain values for all properties),
-                         or ``None`` if ``element`` is the root element.
-    :param base_url: The base URL used to resolve relative URLs.
-    :param target_collector: A target collector used to get computed targets.
-
-    """
+def compute_variable(value, name, computed, base_url, parent_style):
     from .validation.properties import PROPERTIES
+    already_computed_value = False
 
-    def resolve_var(value, name, computed):
+    if value and isinstance(value, tuple) and value[0] == 'var()':
         variable_name, default = value[1]
         computed_value = _resolve_var(computed, variable_name, default)
         if computed_value is None:
@@ -220,11 +205,34 @@ def compute(element, pseudo_type, specified, computed, parent_style,
                 variable_name.replace('_', '-'), name.replace('_', '-'))
             if name in INHERITED and parent_style:
                 already_computed_value = True
-                return parent_style[name], already_computed_value
+                value = parent_style[name]
             else:
                 already_computed_value = name not in INITIAL_NOT_COMPUTED
-                return INITIAL_VALUES[name], already_computed_value
-        return new_value, False
+                value = INITIAL_VALUES[name]
+        elif isinstance(new_value, list):
+            value, = new_value
+        else:
+            value = new_value
+    return value, already_computed_value
+
+
+def compute(element, pseudo_type, specified, computed, parent_style,
+            root_style, base_url, target_collector):
+    """Create a dict of computed values.
+
+    :param element: The HTML element these style apply to
+    :param pseudo_type: The type of pseudo-element, eg 'before', None
+    :param specified: A dict of specified values. Should contain
+                      values for all properties.
+    :param computed: A dict of already known computed values.
+                     Only contains some properties (or none).
+    :param parent_style: A dict of computed values of the parent
+                         element (should contain values for all properties),
+                         or ``None`` if ``element`` is the root element.
+    :param base_url: The base URL used to resolve relative URLs.
+    :param target_collector: A target collector used to get computed targets.
+
+    """
 
     computer = {
         'is_root_element': parent_style is None,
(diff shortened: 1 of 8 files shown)


mikevoets:
Checks fail in Python 3.5 due to some unexpected syntax in `.egg` files. Any idea how these can be fixed?

weasyprint/css/validation/properties.py
```diff
@@ -89,10 +89,11 @@ def validate_non_shorthand(base_url, name, tokens, required=False):
     if not required and name not in PROPERTIES:
         raise InvalidValues('property not supported yet')
 
-    for token in tokens:
-        var_function = check_var_function(token)
-        if var_function:
-            return ((name, var_function),)
+    if name != 'content':
```
liZe, line 92: Why is this change needed? I’d really like to avoid special cases based on property name…

weasyprint/css/computed_values.py
```diff
@@ -195,6 +195,37 @@ def compute(element, pseudo_type, specified, computed, parent_style,
     """
     from .validation.properties import PROPERTIES
 
+    def resolve_var(value, name, computed):
```
liZe, line 198: Would it be possible to put the function outside of the `compute` function? Redefining the function each time `compute` is called takes useless time.

weasyprint/css/computed_values.py
```diff
@@ -218,37 +249,21 @@ def compute(element, pseudo_type, specified, computed, parent_style,
...
-                    value = INITIAL_VALUES[name]
-            else:
-                value = new_value
+        if value and isinstance(value, list):
+            for i, val in enumerate(value):
+                if not isinstance(val, tuple) or val[0] != 'var()':
+                    continue
+                new_value, already_computed_value = \
+                    resolve_var(val, name, computed)
+                if name == 'content':
```
liZe, line 258: Hmmmm… There’s definitely something wrong with this. `content` shouldn’t be different from other properties. I’m not sure to understand how this PR really works, but having these special cases doesn’t look like a good solution for me :unamused:.

liZe:
> Checks fail in Python 3.5 due to some unexpected syntax in `.egg` files. Any idea how these can be fixed?

Isort doesn’t support Python 3.5 anymore, and WeasyPrint could follow it for the next release. We can safely ignore this error for now.

weasyprint/css/validation/properties.py
```diff
@@ -89,10 +92,11 @@ def validate_non_shorthand(base_url, name, tokens, required=False):
     if not required and name not in PROPERTIES:
         raise InvalidValues('property not supported yet')
 
-    for token in tokens:
-        var_function = check_var_function(token)
-        if var_function:
-            return ((name, var_function),)
+    if name not in MULTIVAL_PROPERTIES:
```
mikevoets, line 95: This is necessary because a `content` declaration is a special case since its property handler returns a list, as opposed to every other property. Currently if one of the rules in `content` is a variable function, it would only parse that variable function and return it from the function here, without processing the other rules. This is why I have added a special case condition here.

I cleaned it up by declaring the content property as a multi value property.

weasyprint/css/utils.py
```diff
@@ -699,6 +699,11 @@ def get_content_list_token(token, base_url):
     if string is not None:
         return string
 
+    # <var>
```
mikevoets, line 702: Here a variable function in a content declaration is acknowledged and returned accordingly.

weasyprint/css/computed_values.py
```diff
@@ -218,37 +257,20 @@ def compute(element, pseudo_type, specified, computed, parent_style,
...
-                    already_computed_value = name not in INITIAL_NOT_COMPUTED
-                    value = INITIAL_VALUES[name]
-            else:
-                value = new_value
+        if value:
+            converted_to_list = False
+
+            if not isinstance(value, list):
+                converted_to_list = True
+                value = [value]
```
mikevoets, line 265: Multi values properties (for now only content property classifies as that) return a list that possibly includes variable functions among their declaration values. Therefore, doing `isinstance(value, tuple) and value[0] == 'var()'` would never compute a variable for that situation.

I fixed this by looping over all the values and compute the variables when there are variable functions. I temporarily convert non-list declaration values to a list to do this elegantly.

weasyprint/css/computed_values.py
```diff
@@ -176,6 +176,46 @@ def decorator(function):
...
+                'for property `%s`.', computed_value,
+                variable_name.replace('_', '-'), name.replace('_', '-'))
+            if name in INHERITED and parent_style:
+                already_computed_value = True
+                value = parent_style[name]
+            else:
+                already_computed_value = name not in INITIAL_NOT_COMPUTED
+                value = INITIAL_VALUES[name]
+        elif isinstance(new_value, list):
+            value, = new_value
```
mikevoets, line 213: The `content` property handler always returns a list regardless of input. In this case, the handler will only ever get a computed variable declaration, and thus I unpack the computed value again after getting it back from the handler.

mikevoets:
@liZe I've replied to your comments in my newest commit. I have also cleaned up a whole bit. Thanks for your review and feedback!

mikevoets:
@liZe You're probably busy, but did you have time to look at this?

liZe:
> @liZe You're probably busy, but did you have time to look at this?

I’m busy, as I’m… on holidays :smile:. Thanks a lot for your changes, I’ll have the time to review them this week or next week.

mikevoets:
Hope you had a good holiday @liZe. If there's anything I can do to improve the PR, let me know :-)

liZe:
> Hope you had a good holiday @liZe. If there's anything I can do to improve the PR, let me know :-)

I’m late, I’m really sorry… It’s currently complicated for me to find the time for a good review. I’ll do my best…

mikevoets:
> > Hope you had a good holiday @liZe. If there's anything I can do to improve the PR, let me know :-)
> 
> I’m late, I’m really sorry… It’s currently complicated for me to find the time for a good review. I’ll do my best…

All good! Hope you'll be able to look at it soon. Weasyprint has been extremely useful in my projects and I am happy to continuously contribute to improve it.

mikevoets:
Hi @liZe, just checking in to see if you had any chance to reviewing this PR any time soon? I'm depending on this fix currently, and I need to use a fork of this repository until the PR has been merged. Hope I can revert to upstream soon!

liZe:
> Hi @liZe, just checking in to see if you had any chance to review this PR any time soon? I'm depending on this fix currently, and I need to use a fork of this repository until the PR has been merged. Hope I can revert to upstream soon!

Thanks again for your patience.

I’ve found the time to read your pull request, and I have mixed feelings about it. Your code definitely goes in the right direction, I will merge it, and I’ll add a couple of minor changes (for example, to add all the properties returning iterables).

The main problem with this PR is not your code 😄. The main problem is that this PR shows how much the current implementation is wrong. I already knew that a lot of CSS values are stored in random Python structures, and passed from function to function where they’re randomly detected and randomly computed. It’s been bad for a long time, but it now reaches dangerous limits, and I don’t want to go beyond.

Let’s say this PR is the last one before a large rewrite. If we want to have `var`, `attr` and `calc` (hello #357!) working correctly, we have to clean this quickly before it turns into a monster (I’m looking at you, `split_first_line`).

mikevoets:
Thanks so much. If there's anything I can do to help you cleaning up, let me know.

liZe:
Thanks again, it’s finally merged.

> I will merge it, and I’ll add a couple of minor changes (for example, to add all the properties returning iterables).

Most of the properties returning iterables use the `comma_separated_list` decorator and return a tuple. Having a single way to handle `var` for `content` and these properties would require a lot of work.

We’ll rewrite this sooner or later, to get a clean support of `var`, `calc` and `attr`.

mikevoets:
@liZe thanks!

PR #1231, opened 2020-10-11
Kozea/WeasyPrint#106: handle rtl ol counter

malnajdi:
This pull request is related to the issue, #106 

This is an early attempt to fixing the issue of ordered list counter not respecting the direction when it's rtl.

### DIFF
--- a/weasyprint/css/counters.py
+++ b/weasyprint/css/counters.py
@@ -280,12 +280,16 @@ class CounterStyle(dict):
         # Step 6
         return initial
 
-    def render_marker(self, counter_name, counter_value):
+    def render_marker(self, counter_name, counter_value, parent_direction):
         """Generate the content of a ::marker pseudo-element."""
         counter = self.resolve_counter(counter_name)
         if counter is None:
             if 'decimal' in self:
-                return self.render_marker('decimal', counter_value)
+                return self.render_marker(
+                    'decimal',
+                    counter_value,
+                    parent_direction
+                )
             else:
                 # Could happen if the UA stylesheet is not used
                 return ''
@@ -295,6 +299,11 @@ class CounterStyle(dict):
 
         value = self.render_value(counter_value, counter_name=counter_name)
         assert value is not None
+
+        if parent_direction == 'rtl':
+            suffix = symbol(counter['suffix'] or ('string', ' .'))
+            return suffix + value + prefix
+
         return prefix + value + suffix
 
     def copy(self):
--- a/weasyprint/formatting_structure/build.py
+++ b/weasyprint/formatting_structure/build.py
@@ -145,6 +145,15 @@ def element_to_box(element, style_for, get_image_from_uri, base_url,
 
     marker_boxes = []
     if style['display'] == 'list-item':
+        # This should handle rtl direction
+        if style['direction'] == 'rtl':
+            translate_x = properties.Dimension(-5, '%')
+            translate_y = computed_values.ZERO_PIXELS
+            box.style['transform'] = (
+                    ('translate', (translate_x, translate_y)),
+                )
+            box.style['padding_right'] = properties.Dimension(10, 'px')
+
         marker_boxes = list(marker_to_box(
             element, state, style, style_for, get_image_from_uri,
             target_collector, counter_style))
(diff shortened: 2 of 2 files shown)


malnajdi:
I am not sure how can i add test to this pull request

weasyprint/formatting_structure/build.py
```diff
@@ -311,9 +319,11 @@ def marker_to_box(element, state, parent_style, style_for, get_image_from_uri,
         # See https://drafts.csswg.org/css-pseudo-4/#marker-pseudo
         marker_box.style['position'] = 'absolute'
         if parent_style['direction'] == 'ltr':
+            marker_box.style['left'] = 'auto'
```
liZe, line 322: Why is this line needed?

weasyprint/formatting_structure/build.py
```diff
@@ -311,9 +319,11 @@ def marker_to_box(element, state, parent_style, style_for, get_image_from_uri,
...
             translate_x = properties.Dimension(-100, '%')
         else:
-            translate_x = properties.Dimension(100, '%')
+            marker_box.style['right'] = properties.Dimension(0, 'px')
+            translate_x = properties.Dimension(20, '%')
```
liZe, line 326: This 20% value is a bit strange… Why doesn’t the 100% translation work?

weasyprint/css/counters.py
```diff
@@ -280,12 +280,16 @@ def render_value(self, counter_value, counter_name=None, counter=None,
         # Step 6
         return initial
 
-    def render_marker(self, counter_name, counter_value):
+    def render_marker(self, counter_name, counter_value, parent_direction):
```
liZe, line 283: 👍🏼

weasyprint/css/counters.py
```diff
@@ -295,6 +299,11 @@ def render_marker(self, counter_name, counter_value):
 
         value = self.render_value(counter_value, counter_name=counter_name)
         assert value is not None
+
```
liZe, line 302: You can put the `suffix = …` from line 298 + the return in an `else` block, that would improve readability in my opinion.

weasyprint/formatting_structure/build.py
```diff
@@ -145,6 +145,15 @@ def element_to_box(element, style_for, get_image_from_uri, base_url,
 
     marker_boxes = []
     if style['display'] == 'list-item':
+        # This should handle rtl direction
```
liZe, line 148: Why is this code needed?

weasyprint/css/counters.py
```diff
@@ -295,6 +299,11 @@ def render_marker(self, counter_name, counter_value):
...
+        if parent_direction == 'rtl':
+            suffix = symbol(counter['suffix'] or ('string', ' .'))
```
liZe, line 304: Having ' .' instead of '. ' is an easy fix, but I’m not sure about how the specification handles this. We should read it, and maybe implement [`marker-side`](https://drafts.csswg.org/css-lists-3/#marker-side) to have a clean solution.

liZe (changes requested):
The overall idea is good, and it could be merged (with the translation hack removed :wink:). But I’d really like to know the specification handles the order of prefix / value / suffix and the space before / after the full stop…

weasyprint/css/counters.py
malnajdi: I did not know marker-side specification was there. I am back with the specification but, thank you for pointing that out. I will try to implement it.

weasyprint/formatting_structure/build.py
malnajdi: It's not needed, i will remove it.

weasyprint/formatting_structure/build.py
malnajdi: I am not sure why 100% did not work. I know 20 is really strange and i will investigate more.

weasyprint/formatting_structure/build.py
malnajdi: After applying this patch it seems the only way for `direction: ltr` to work is to set left to auto.

weasyprint/css/counters.py
malnajdi: Will do.

liZe:
I’ll close this PR for now, feel free to reopen if you find the time to work on this!

Issue #1247, opened 2020-11-08
incomplete pypi tarballs: test_draw, test_layout missing


FelixSchwarz:
`WeasyPrint-52.1.tar.gz` seems to be incomplete. `test_draw` and `test_layout` are not present in the pypi tarball leading to test failures.

liZe:
Fixed by #1248.

eli-schwartz:
Note that using `packages =  find:` is probably less fragile...

PR #1248, opened 2020-11-08
include missing test data (#1247)

FelixSchwarz:
pypi tarballs are currently missing some resources required to run the test suite.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -45,6 +45,8 @@ packages =
   weasyprint.formatting_structure
   weasyprint.layout
   weasyprint.tests
+  weasyprint.tests.test_draw
+  weasyprint.tests.test_layout
   weasyprint.tools
 zip_safe = false
 setup_requires = pytest-runner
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -71,6 +73,7 @@ console-scripts = weasyprint = weasyprint.__main__:main
 [options.package_data]
 weasyprint = VERSION
 weasyprint.tests = resources/*.*, resources/*/*
+weasyprint.tests.test_draw = results/*.*
```
liZe, line 76: I think that this folder shouldn’t be included, does it?

liZe (changes requested):
Thanks a lot for this pull request!

setup.cfg
liZe: I’ve merged your PR and removed this line. Thank you!

FelixSchwarz: Sorry for answering so late. I thought the `options.package_data` line `setup.cfg` was necessary but I was probably mistaken so I'm glad you fixed my PR.

PR #1268, opened 2020-12-14
Add new dll names

naveen521kk:
When building Pango with meson I get the following `DLL` files as output.
So adding it.
If interested I will add [this](https://github.com/naveen521kk/pango-build) to docs for installing on Windows, at least would help for someone. Or I myself, would make a [`chocolatey`](https://chocolatey.org/) package if needed, but that need's to be there in docs, for choco package to be done.


liZe:
Thank you!

liZe:
> If interested I will add [this](https://github.com/naveen521kk/pango-build) to docs for installing on Windows, at least would help for someone. Or I myself, would make a [`chocolatey`](https://chocolatey.org/) package if needed, but that need's to be there in docs, for choco package to be done.

Windows packaging is something we’ll talk about when docs will be refreshed. The way WeasyPrint is currently installed in GitHub Actions is a pretty simple way to get something quite reliable.

Issue #1241, opened 2020-10-24
Building Wheels


naveen521kk:
One good thing about wheels is that it allow us to embed DLL files. So simply adding all DLL files and making the wheel platform specific would significantly help users by not installing system dependencies. I recently made a PR in cairocffi repo https://github.com/Kozea/cairocffi/pull/164 providing a way to do so. I got a nice repo which builds cairo as a single DLL file so I could do that. But here I couldn't find the one for Pango ( I don't think anything else is needed). So, finding a way to build that would be good or rather it can also be taken from msys2 in case of windows and brew in case of MacOS. I don't about linux. 

What I did there was not that correct I think but I don't know. Possibly there should be a better solution and I started searching for similar libraries.
I found `freetype-py` which had a script to build freetype same thing can be done here also I think. 
Next, I found  this issue https://github.com/pypa/wheel/issues/128 which is more related and from that thread I found `pypandoc` where their implementation was very good IMHO, See [`setup.py`](https://github.com/bebraw/pypandoc/blob/master/setup.py).
Next, I find another interesting thing which is used with `cffi` is [`milksnake`](https://github.com/getsentry/milksnake) which simply suited this one but I haven't tried using it.

Sorry, for a braindump I thought I would organize it a little better but I didn't find time. 

Finally, the conclusion is there must be platform specific wheel so that weasyprint can simply be installed using `pip` without any system dependencies.

naveen521kk:
This [commit](https://github.com/Kozea/WeasyPrint/commit/4f82a4a34eab9442db6d3b1d5c87d974dafb09f5#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711) breaks the possibility of doing this. For doing this a `setup.py` is must and without which it can't be done.

naveen521kk:
@liZe I plan to help here. I think needed dependencies are Pango and Gdk-Pixbuf right? 
For Pango, one of my repo, [pango-build](https://github.com/naveen521kk/pango-build), can be used(actually made for another project), there the DLL files are name as `CORE_MANIM_*.dll` to avoid DLL hell. I need to look into `Gdk-Pixbuf` though, it also uses meson so it should be easy to compile it using meson. Thoughts?

liZe:
> @liZe I plan to help here.

Thank you!

> I think needed dependencies are Pango and Gdk-Pixbuf right?

[...]

PR #1302, opened 2021-03-02
feat: build windows wheels

naveen521kk:
fixes https://github.com/Kozea/WeasyPrint/issues/1241
TODO:
- [x] the license file isn't included in the wheel
- [x] Set up Github Actions to Test them and Build them
- [x] Also, https://github.com/Kozea/WeasyPrint/issues/1241#issuecomment-789090182

If anyone wants to try I had already built a wheel. 
[WeasyPrint-53.0-py3-none-win_amd64.zip](https://github.com/Kozea/WeasyPrint/files/6070806/WeasyPrint-53.0-py3-none-win_amd64.zip)


naveen521kk:
@liZe is there any smoke tests that can be run to test those wheels?

naveen521kk:
Any chance for reviewing it?

liZe:
> Any chance for reviewing it?

Hello @naveen521kk.

Thanks for your hard work on this pull request, we’re sorry we took so much time to answer.

It’s a bit complicated right now for us to review this pull request, because there’s a lot of things going on with version 53: new packaging, new dependencies, etc. Moreover, providing wheels with embedded libraries is a really important decision, because it’s almost mandatory to provide them forever. It adds code that we’ll have to maintain and to test, knowing that we don’t have Windows installed on our computers.

The ideal solution would include:
- no extra Python code,
- a whole automation of packaging,
- packaging tests for all supported platforms.

Of course, this dream is a lot of work (if it’s ever possible :wink:, that’s sometimes the problem with dreams). This PR is definitely a huge step in the right direction, and we’re very grateful for that, but both fixing the technical details and accepting the [little cute puppy](https://youtu.be/tzFWz5fiVKU?t=1h3m21s) will take some time 🐕.

naveen521kk:
> It adds code that we’ll have to maintain and to test, knowing that we don’t have Windows installed on our computers.

You can always ping me, I should be able to help if things don't work :)
I have documented the code I wrote so that it doesn't create problems and is easy for anybody to understand.

> no extra Python code,

I thought it should be better to use python rather than batch or PowerShell? Anyway, this shouldn't create problems, I believe.

> a whole automation of packaging,

As you can see it is automated. Both the [pango-build](https://github.com/naveen521kk/pango-build)(which I will any maintain for another project) and Github Actions here automatically builds things. It builds and also runs a simple command as a test see the GitHub workflow I added in this PR.

> packaging tests for all supported platforms.

I think it makes sense to pack only for Windows because for Linux it is there, by default, or can easily be installed through the package manager. The same is for macOS, they have Brew.

naveen521kk:
Is it possible to review this at least now?

PR #1328, opened 2021-04-02
add ISO and JIS paper sizes

Crissov:
- some were missing from the ones already required by [css-page](https://drafts.csswg.org/css-page-3/#typedef-page-size-page-size)
- the rest has been proposed for addition in https://github.com/w3c/csswg-drafts/issues/4722
- complete ISO 216 A, B and C series
- JIS P 0202 B series


grewn0uille:
Thank you!

PR #1346, opened 2021-04-29
SVG: Allow stroke-dasharray:none

pbregener:
Fixes a crash when a SVG contains `style="stroke-dasharray:none"`.
Also see SVG spec on this: https://www.w3.org/TR/SVG11/painting.html#StrokeDasharrayProperty

Not sure this is the most elegant way of fixing it, but I guess you get the idea 😄

### DIFF
--- a/weasyprint/svg/__init__.py
+++ b/weasyprint/svg/__init__.py
@@ -573,7 +573,7 @@ class SVG:
 
         # Apply dash array
         dash_array = tuple(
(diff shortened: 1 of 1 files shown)


liZe:
> Not sure this is the most elegant way of fixing it, but I guess you get the idea

Thanks a lot for the fix! I’ll see if I can find a more elegant way :wink:.

Issue #1352, opened 2021-05-10
Out-of-order internal link target IDs break in some viewers


aschmitz:
A document with internal links generally works fine, but if the link target IDs aren't in lexicographic(? ASCIIbetical?) order, some PDF viewers (notably at least Chrome's PDFium and macOS' Preview.app) will not treat the links as if they exist.

For example, rendering the following HTML fragment (no CSS necessary) with the current `master` branch will produce [this PDF](https://github.com/Kozea/WeasyPrint/files/6454580/1352.pdf), for which the first link works in all tested PDF viewers (PDF.js, PDFium, Poppler, Preview.app), but the second link fails to work in PDFium and Preview.app:

~~~html
<p>Use <a href="#zzz">link one</a> or <a href="#aaa">link two</a>.</p>
<div id="zzz"><h1>One</h1></div>
<div id="aaa"><h1>Two</h1></div>
<div style="break-before: always;">Second page for scrolling.</div>
~~~

However, swapping the `zzz` and `aaa` IDs (while leaving everything else as-is) appears to make both links work correctly.

I did a bit of digging into this, and it looks like the "Name Trees" part of the PDF reference (3.8.4) says the keys in name trees must be ordered, but it's not clear what the correct ordering there is. (They are currently in the order the targets appear in the document: perhaps some software is treating that "tree" as ordered by some key sort order and assuming the link targets don't exist?)

Hopefully the fix is simply just sorting the name tree, but you'll likely know that better than I do. (If I get some spare time, I may also attempt a fix, but again I suspect you'll be faster with it. :slightly_smiling_face:)

liZe:
Thanks a lot for the bug report, with the very useful HTML snippet.

PR #1358, opened 2021-05-19
Sort anchors across pages

aschmitz:
Previously we had sorted anchors within one page, but if anchors happened to be out of order in subsequent pages, they wouldn't be written in order, which caused lingering issues in some PDF viewers.

Related to #1352 and the change in d4561b13af89e89acb66586b1e2da66dfaba1fac.

I'm not a huge fan of using lists to store different types of data indexed by position, but this was a quick way to do it that gets handled well by `sorted`. I won't be offended if you want to rewrite it using dicts or something else that might be more Pythonic. :slightly_smiling_face:


aschmitz:
Apologies, I thought I had removed all of my debugging before committing. This should work better.

PR #1373, opened 2021-06-02
SVG: Support fill-opacity and stroke-opacity

### LINKED ISSUE #1371: SVG: Support for fill-opacity and stroke-opacity
The old cairo-based weasyprint supported `fill-opacity` and `stroke-opacity`.  
The new cairo-less version only seems to support `opacity` so far.

Do you have plans to bring this functionality back?
I have tons of automatically generated SVGs that use those attributes.

<details>
<summary>Simple example:</summary>

`rect.svg`:

```xml
<svg width="100" height="100">
  <rect x="10" y="10" width="80" height="80"
    style="fill:red;fill-opacity:0.3;stroke-width:10;stroke:black;stroke-opacity:0.2"/>
</svg>
```

and `test.html`

```html
<html>
<body>
<img src="rect.svg" style="width:100%;">
</body>
</html>
```
</details>

pbregener:
I'm not too familiar with the WeasyPrint codebase so far, but this 2-line change seems to work for the examples I tested!

Closes #1371

### DIFF
--- a/weasyprint/svg/__init__.py
+++ b/weasyprint/svg/__init__.py
@@ -561,8 +561,9 @@ class SVG:
             self, node, fill_source, font_size, stroke=False)
         if fill_color and not fill_drawn:
             red, green, blue, alpha = color(fill_color)
+            alpha = node.get('fill-opacity', alpha)
             self.stream.set_color_rgb(red, green, blue)
-            self.stream.set_alpha(node.get('fill-opacity', alpha))
+            self.stream.set_alpha(alpha)
         fill = fill_color or fill_drawn
 
         # Get stroke data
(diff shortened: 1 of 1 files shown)


liZe:
Hello!

Thanks **a lot** for your pull request, but that’s actually more complicated than this:
- we have to take care of the color’s opacity,
- we have to handle `fill-opacity` and `stroke-opacity` for gradients and patterns.

I’m on my way for a complete pull request, but I’ve found a dirty bug with PDF alpha groups (currently we can’t have the same alpha value for both fill and stroke).

pbregener:
Hi! I was afraid that there were 1-100 things that I missed 😄 
This solution just seemed too easy, even though it worked for the basic things I tested...

And thanks for picking this up!

PR #1389, opened 2021-07-06
Add text-align-last

ColinKinloch:
I compared a few tests with firefox, as far as I can tell this works as expected.

[CSS Text Module Level 3 draft](https://drafts.csswg.org/css-text/#text-align-last-property)
[MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/text-align-last)


liZe:
Thanks a lot! We’ll clean it and merge it after 53.0 is released.

liZe:
Thanks a lot for the pull request. It’s now merged, and improved to handle `text-align-all` and `text-align-last`, transforming `text-align` into a shorthand.

PR #1401, opened 2021-07-20
More robust handling of SVG use tags

### LINKED ISSUE #1399: SVG: Crash when referencing (use) non-existing elements
Rendering a SVG that `use`s non-existing elements results in a crash.

This has been a problem in all v53 betas, including current `master` . The cairo-based version simply ignored those `use` statements.

There are various potentially trivial fixes, let me know if you'd prefer a PR over fixing this yourself @liZe :)

Example SVG:
```svg
<svg viewBox="0 0 16 10" xmlns="http://www.w3.org/2000/svg">
  <circle id="valid" cx="5" cy="5" r="4" stroke="green"/>
  <use href="#valid" x="6" fill="green"/>
  <use href="#notvalid" x="7" fill="red"/>
</svg>
```

Short backtrace:
```
  File "[...]\weasyprint\svg\defs.py", line 30, in use
    tree = svg.tree.get_child(parsed_url.fragment).copy()
AttributeError: 'NoneType' object has no attribute 'copy'
```
https://github.com/Kozea/WeasyPrint/blob/56910a84e7913342ef79f2a9894ee557bdb146de/weasyprint/svg/defs.py#L30

pbregener:
Ignore `use` tags in SVGs that have invalid targets.

Instead of calling `svg.stream.push_state()` early, only call it when we are sure that the stream should indeed be modified.
Also, from what I could see the error handling calling `svg.stream.restore()` was bound to fail as well (should be `pop_state()` instead?).

Feel free to ignore and close if I missed something and things are not as easy as I thought they were 🙈 

Fixes #1399

### DIFF
--- a/weasyprint/svg/defs.py
+++ b/weasyprint/svg/defs.py
@@ -19,22 +19,22 @@ def use(svg, node, font_size):
     """Draw use tags."""
     from . import NOT_INHERITED_ATTRIBUTES, SVG
 
-    svg.stream.push_state()
-
     for attribute in ('x', 'y', 'viewBox', 'mask'):
         if attribute in node.attrib:
             del node.attrib[attribute]
 
     parsed_url = parse_url(node.get_href())
     if parsed_url.fragment and not parsed_url.path:
-        tree = svg.tree.get_child(parsed_url.fragment).copy()
+        try:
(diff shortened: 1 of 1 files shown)


liZe:
Perfect, thank you!

PR #1447, opened 2021-09-17
Support parallel flows

### LINKED ISSUE #36: Allow page breaks in floats, absolute blocks, table-cells
Floated elements that don't fit on the current page simply fall off the bottom, rather than being placed on the next page.

Here's a handy long list of floated elements to demonstrate the problem: http://www.stripey.com/demo/weasyprint/float_off_bottom.html

Look at it in Firefox and do ‘Print Preview’. You should see that there's a page break, with the list being continued on page 2. Similarly if you print from Chromium.

But WeasyPrint generates this file, where the elements simply run off the bottom of the first page: http://www.stripey.com/demo/weasyprint/float_off_bottom.pdf

liZe:
Will fix #36.

### DIFF
--- a/tests/test_float.py
+++ b/tests/test_float.py
@@ -510,7 +510,7 @@ def test_preferred_widths_5():
 
 
 @assert_no_logs
-def test_float_in_inline():
+def test_float_in_inline_1():
     page, = render_pages('''
       <style>
         @font-face { src: url(weasyprint.otf); font-family: weasyprint }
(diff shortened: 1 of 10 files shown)


liZe:
The current master branch includes parallel flows support for table cells and floats. But it doesn’t include code from this branch 😱. Why? It’s time to close this draft pull request and explain why it failed.

We changed the way `resume_at`/`skip_stack` is stored in 3aca9f1f. We tried to use this structure in this PR to store the different parallel flows, but that wasn’t a good idea: we assume everywhere that when the main flow is not split, `resume_at` is `None`. That’s not the case anymore when we split floats for example. And that’s a nightmare to handle correctly.

Hopefully, using the new structure is not useless: it’s perfect when the main flow has parallel flows, in tables for example. It will be useful for flex and grid too. But the new structure is useless for out-of-flow elements (floats, absolutes…). We finally [decided](https://github.com/Kozea/WeasyPrint/commit/dbec957f5c51bef67d11966e58cc87a99e1ab2cb) to store the broken out-of-flow elements in the context.

We also tried in another branch to store the layout progress (skip stacks, positions…) in a common structure, to avoid long lists of parameters. That’s a good idea, isn’t it? No, it’s not. The code we got was actually longer and broken: using global mutable structures is dangerous and always introduces side effects.

So… It’s time to say goodbye to these branches, and to this code. We’ve spent countless hours to write it, it’s sad for sure, but it’s important to know when to let things go… We’ve learned a lot of things, we’ve found better solutions, there’s no need to cry. It was important to write this comment so that we don’t forget the errors we’ve done.

It was also important to write this comment to remember that one smart line of code in master sometimes requires to write dozens of invisible lines before. To all these lines that will be discarded: you can rest in peace, you’ll be proudly remembered.

💜🍲🪦

Issue #1475, opened 2021-10-20
Generated PDF size jumped from 2.2 -> 8MB after upgrade to 53.3


gnyers:
Last week I've updated my VENV to the latest WeasyPrint (v53.3) with `pip`. Suddenly the same document that used to be 2.2MB is now built as a 8.2. 

I'm using:

```bash
$ weasyprint --version
WeasyPrint version 53.3

$ python --version
Python 3.6.12

$ head -2 /etc/os-release 
NAME="openSUSE Leap"
VERSION="15.2"
```

Unfortunately I don't exactly know the previous version that I've used, but I thing it was a v52.?

This example document is moderately heavy on PNG images.

Issue #1462, opened 2021-10-06
Embedding base64-encoded image on 53.0 takes 30× more space


knyttl:
Consider following html:

```html
<!DOCTYPE html>
<html>
<head>
  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
</head>
<body>
      <img src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAZAAAAGQAQAAAACoxAthAAABiElEQVR4Xu3YPXLCMBCGYTEpKHMEjuKj4aPlKD5CSooMQvsjWQqeWDIUKd6vAa/22U6zmBCHE34X9gMZDWQ0kNEcI7fQ5FPrd3uY4tIcpgIE0gTSQT60S7PU5FoVNDMEAjlEtCOkgnTc0rdT7siFFAgE8jaiHRIIBPJuEuOPPT7NgEAgx0hJJmnx1TM8EAjkCGkil1CvqXUs7SkEAqnTQzaii8/IcyAbgUB2iXTZJTzXBYvPsLp/QiCQbuJ7LoRLtE148vM0wzahZpYZEAkEMkRKYQlyK8vr28YMyBoIpIdIIditnLRWXUIjvvgkEwQCOU4ktufuuUcKTmwOBKKBeMc+eUpDLlrSTShDJ0jVmgOB/EX8VuZUt/Kaupb1xCdCIHUgHcR/SkqWUPacEUkzAwKB5PQTLYS01oSUGbMWZUb5m3KCQCCvk/Kf45fVLVaCQCAvkpI5FSy+CaUDAoGMk5KGWL7DOW4svhLI+mCBREhLmlREO1ZSv75BIJBeMhrIaCCjgYzm35IH8TLVbS2jtlUAAAAASUVORK5CYII=">
</body>
</html>
```

When compiled with 52.5, the PDF takes 1.5 kb. When compiled with 53.0, it takes 40 kb.

- [ticket52.5.pdf](https://github.com/Kozea/WeasyPrint/files/7291318/ticket52.5.pdf)
- [ticket53.0.pdf](https://github.com/Kozea/WeasyPrint/files/7291319/ticket53.0.pdf)

I would expect the PDF not to grow so enormously.

Issue #1444, opened 2021-09-10
Raster image rendering performance dropped


SvenBecker:
The rendering speed of raster images dropped since WeasyPrint version 53, see benchmarks in #1439.

liZe:
The problem is particularly important with images other than JPEGs, as we have to convert them into JPEG2000 to embed them in the PDF.

Maybe there’s a better solution. I’ve read that some PNG files can also be embedded with no conversion.

aschmitz:
I'll also mention that Firefox's PDF.js is also very slow at decoding JPEG2000 images, so this affects both rendering and viewing speed in some cases.

I believe you're right that certain kinds of PNG can be embedded without significant image conversion (assuming a compatible filter, color choices, etc.), but if we're trying to load raster graphics (other than JPEG or JPEG2000), it might be fastest to convert to that subset of PNG that can be directly embedded, rather than JPEG2000 (which is a fairly complicated and slow format to encode).

liZe:
> I'll also mention that Firefox's PDF.js is also very slow at decoding JPEG2000 images, so this affects both rendering and viewing speed in some cases.

That’s sad :cry:.

> it might be fastest to convert to that subset of PNG that can be directly embedded, rather than JPEG2000

That would be the best solution. I’ve spent some time playing with the PIL to get the information needed to include the PNG files, but I’m not sure that it’s the right tool for this job. Could a PNG-in-PDF guru help me for this issue :smile:?

liZe:
#1481 should help a lot. Tests with the current `master` are welcome!

mpth:
Just ran a test with a real-world document. It had ~31 images pointing to PNG files and also 515 PNG images embedded via data-uri (`src="data:image/png..."`)

Size before #1481 
`37.5 MB`

Size after #1481
`4.7 MB`

Looks like an improvement to me 🎉 

Thanks @liZe @aschmitz ❤️

PR #1481, opened 2021-10-29
change: encode non-JPEG images as PNGs instead of JPEG2000 images

aschmitz:
This uses Pillow to re-encode any non-JPEG image as a PNG, then inline that
image's IDAT chunks as a FlateDecode value, which allows us to reuse the work
from the PNG encoder. This means we'll reencode some PNGs we could have passed
through directly, but that could be changed later. Alpha layers continue to be
handled separately, as appears to be required by the PDF spec.

Provides a potential fix for #1444, which works in local tests (and appears to pass the existing test suite, at least locally).

There's a lingering optimization where we could directly pass through PNGs that meet certain criteria:

* Must have no alpha transparency (it has to be broken out into a separate layer).
* May need to not be indexed / using a palette (I don't remember if PDF supports this directly, but if it does, it would require more work).
* We would need to do more work to handle gamma correction (which PDF does support), color spaces (maybe a PDF feature), rotation, etc. if relevant for the image - although I don't actually know if Pillow is covering those for us now or not.
* If we're particularly suspicious of the encoding of the PNG, `optimize_size` could lead us to push Pillow to re-encode the image, although I suspect this would rarely show a benefit and could increase sizes in already-extra-compressed images.

Because that requires more investigation and work, I haven't done it here, this is mostly just the minimum effort to get the change out the door. It could be worth filing an issue to investigate "PNG passthrough" for the future, but this solves my immediate concerns of JPEG2000 being slow. I expect - but haven't thoroughly tested - that it may also resolve issues like #1475. (For the example in #1462, the output PDF with this patch is 6.06 KiB, which is bigger than version 52, but much smaller than version 53.0.)

### DIFF
--- a/weasyprint/document.py
+++ b/weasyprint/document.py
@@ -10,6 +10,7 @@ import hashlib
 import io
 import math
 import shutil
+import struct
 import zlib
 from os.path import basename
 from urllib.parse import unquote, urlsplit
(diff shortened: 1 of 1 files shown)


liZe:
💜 Thanks a lot for this pull request! 💜

> There's a lingering optimization where we could directly pass through PNGs that meet certain criteria:

Sure. As you said, it can be done later, this solution is probably already faster than using JPEG2000.

> I expect - but haven't thoroughly tested - that it may also resolve issues like #1475. (For the example in #1462, the output PDF with this patch is 6.06 KiB, which is bigger than version 52, but much smaller than version 53.0.)

#1475 is probably solved (@gnyers and @mpth, feel free to test when it’s merged!). #1462 uses a 1-bit image as far as I can remember, so it could be improved if we can use a palette in the future.

We’ll probably have to get a lot of different image formats for tests. We should have done this in version 53 already, but it would be a good topic for [version 55](https://github.com/Kozea/WeasyPrint/milestone/32).

> I see the failing test, but I don't understand what it's saying. Do you understand what the issue is? I suspect there's probably something up with the way we're lightly abusing BytesIO (possibly the `.seek(0)`, although that seems to work fine), but I'll admit that Python is not a native language for me.

I suppose that the linter doesn’t like `document.py` (maybe a useless import), and that a bug in a pytest plugin gives this error. Nothing to worry about, I’ll fix it!

liZe:
(For the record: the bug is https://github.com/tholo/pytest-flake8/issues/81.)

mpth:
Do I have to clone the repo and build the binary myself or can I get it via pip or from elsewhere?

liZe:
> Do I have to clone the repo and build the binary myself or can I get it via pip or from elsewhere?

You can use `pip install git+https://github.com/Kozea/WeasyPrint`.

PR #1493, opened 2021-11-12
Make sure url is relative before joining base_url

ColinKinloch:
I have a nuxt project that results in a pdf missing fonts.

The project that produces css like this:
```
@font-face {
  font-family: ForkAwesome;
  src: url(/_nuxt/fonts/forkawesome-webfont.2668138.eot);
  src: url(/_nuxt/fonts/forkawesome-webfont.2668138.eot?#iefix&v=1.2.0)
      format("embedded-opentype"),
    url(/_nuxt/fonts/forkawesome-webfont.fafb863.woff2) format("woff2"),
    url(/_nuxt/fonts/forkawesome-webfont.2350c61.woff) format("woff"),
    url(/_nuxt/fonts/forkawesome-webfont.94d78a2.ttf) format("truetype"),
    url(/_nuxt/img/forkawesome-webfont.86fe294.svg#forkawesomeregular)
      format("svg");
  font-weight: 400;
  font-style: normal;
  font-display: block;
}
```

The URLs are absolute with root being the `./dist` directory generated by nuxt.

```
weasyprint.CSS(filename='./dist/_nuxt/css/25a755f.css', base_url='./dist', font_config=font_config)
```

would result in 
```
urljoin('/path/to/dist/', '/_nuxt/fonts/forkawesome-webfont.fafb863.woff2')
```
which would return `/_nuxt/fonts/forkawesome-webfont.fafb863.woff2'` rather than joining with the base_url.

This MR converts absolute path URLs to relative: `_nuxt/fonts/forkawesome-webfont.fafb863.woff2` before calling joinurl.

### DIFF
--- a/weasyprint/css/utils.py
+++ b/weasyprint/css/utils.py
@@ -10,6 +10,7 @@
 import functools
 import math
 from urllib.parse import unquote, urljoin
+from pathlib import Path
 
 from tinycss2.color3 import parse_color
 
@@ -145,6 +146,10 @@ def safe_urljoin(base_url, url):
     if url_is_absolute(url):
         return iri_to_uri(url)
     elif base_url:
+        # Make sure url is relative before joining base_url
+        url_path = Path(url)
+        if url_path.is_absolute():
+            url = str(url_path.relative_to(url_path.root))
         return iri_to_uri(urljoin(base_url, url))
     else:
         raise InvalidValues(


liZe:
Hello!

Thanks for your pull request. But I’m not sure that there’s a bug in the current behavior.

Opening an HTML file with `` in a browser tries to open a the file in the root of the filesystem, not in the current folder (that’s the default base URL).

ColinKinloch:
I suppose my problem is that I want to use web development tools that assume that root is the root of the domain, but I don't want to have to spin up a http server to produce a the pdf.

Perhaps I should be using chroot?

liZe:
> Perhaps I should be using chroot?

It’s a possible solution. If you can also use relative URLs, it’s probably much easier.

ColinKinloch:
Or perhaps a boolean called `base_url_as_root` or something that enables this code?

liZe:
> Or perhaps a boolean called `base_url_as_root` or something that enables this code?

We really try to follow the specifications, so we try hard not to add non-standard features like this one. Sorry…

If you really can’t use relative URLs, you can write your own [`url_fetcher`](https://doc.courtbouillon.org/weasyprint/stable/api_reference.html#weasyprint.default_url_fetcher). In a couple of lines, you’ll then be able to handle remove the leading slash from the URL and call the original `url_fetcher` with the relative URL.

ColinKinloch:
Seems like a `url_fetcher` that makes the URL relative isn't enough, by the time it gets to `safe_urljoin` it's back to being absolute.

If I edit the CSS files to have relative URLs they're absolute to the root of the filesystem by the time they get to `url_fetcher`.

However, prefixing the absolute path of `dist` in `url_fetcher` has solved my issue.

```
dist = Path('./dist')
def absolute_url_fetcher (url, timeout=10, ssl_context=None):
  parsed_url = urlparse(url)
  url_path = Path(parsed_url.path)
  if url_path.is_absolute():
    url_path = url_path.relative_to(url_path.root)
    parsed_url = parsed_url._replace(path = str((dist / url_path).resolve()))
  url = urlunparse(parsed_url)
  return weasyprint.default_url_fetcher(url, timeout=timeout, ssl_context=ssl_context)
```

Thanks for the help.

PR #1500, opened 2021-11-18
Skip font processing based on hash of source

ColinKinloch:
A lot of my runtime was taken up by decompressing `woff2` files that had previously been loaded.

This change writes the result of `fetch(url_fetcher, url)` to a file named by its `sha256` hash, skipping based on whether the file already exists.

It does the same with the fontconfig xml, I don't know enough about fontconfig to know if this is completely unnecessary.

Replaces #1495


liZe:
Hello, and thank you for this pull request.

Could you please try to find the reason why tests are failing, and try to fix the problem if possible?

Issue #1153, opened 2020-07-06
Add support for CSS property: word-break


qwfy:
Currently, [`word-break`](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break) is not supported, it's needed if you want to do fine-grained text layout, especially for CJK text.

liZe:
Hello!

Strange, `word-break` is documented as supported, but it’s definitely not. It should be possible to add the feature with the possibilities given by Pango.

aschmitz:
It seems as though this is a particularly acute problem in tables, as it can't be worked around with `overflow-wrap` in tables the way it can be in body text:

~~~html
<style>
html {
  overflow-wrap: break-word;
}
</style>

<table><tr><td>LONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONG</td></tr></table>

<p>LONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONGLONG</p>
~~~

![Sample screenshot](https://user-images.githubusercontent.com/29508/143977175-cb8ea2ea-9151-4339-9800-f2accf8b6fb9.png)

I suspect this is probably about to cause some issues for my uses. If you have a few minutes, can you point me towards where in the code I might try to poke to get either `overflow-wrap` to work in tables or `word-break` to work generally?

Bolik:
Hi, have any plans about this feature?

liZe:
> It seems as though this is a particularly acute problem in tables

That’s right, it would be really useful.

> Hi, have you any plans about this feature?

We have plans to rewrite the text management in the future, mostly to handle right-to-left and bidirectional text. But there’s nothing about that in the milestones yet.

> If you have a few minutes, can you point me towards where in the code I might try to poke to get either `overflow-wrap` to work in tables or `word-break` to work generally?

As far as I can remember, `overflow-wrap` should work but we had some problems with [Pango’s wrap mode](https://docs.gtk.org/Pango/enum.WrapMode.html) that wasn’t working as expected. You can look at this code, that’s supposed to break lines at characters boundaries when needed:

[...]

Issue #1514, opened 2021-12-04
test_tab_size fail with Pango 1.50


grewn0uille:

liZe:
That’s probably a bug in Pango: https://gitlab.gnome.org/GNOME/pango/-/issues/635

grewn0uille:
Bug fixed upstream

PR #1520, opened 2021-12-13
Support "overflow-wrap: anywhere"

aschmitz:
This supports the `overflow-wrap` value `anywhere`: `anywhere` is like `break-word`, but the soft breaks it allows *are* considered when calculating min-content intrinsic sizes.

This can help alleviate the problems identified in https://github.com/Kozea/WeasyPrint/issues/1153#issuecomment-982235394, but obviously doesn't close that issue.

I'm a little unsure on the test: if you have preferred ways to test this behavior, feel free to remove these tests. (Additionally, the `break-word` and `normal` tests would *seem* as though they should be 120px wide, not 100, but I can't figure out why they have their current values: I'd be interested in thoughts there.

### DIFF
--- a/tests/test_text.py
+++ b/tests/test_text.py
@@ -891,6 +891,7 @@ def test_hyphenate_limit_chars_punctuation(css):
 
 @assert_no_logs
 @pytest.mark.parametrize('wrap, text, test, full_text', (
+    ('anywhere', 'aaaaaaaa', lambda a: a > 1, 'aaaaaaaa'),
     ('break-word', 'aaaaaaaa', lambda a: a > 1, 'aaaaaaaa'),
     ('normal', 'aaaaaaaa', lambda a: a == 1, 'aaaaaaaa'),
     ('break-word', 'hyphenations', lambda a: a > 3,
@@ -918,6 +919,32 @@ def test_overflow_wrap(wrap, text, test, full_text):
     lines_full_text = ''.join(line for line in lines)
     assert test(len(lines))
     assert full_text == lines_full_text
+
+
+@assert_no_logs
+@pytest.mark.parametrize('wrap, text, body_width, expected_width', (
+    ('anywhere', 'aaaaaa', 10, 20),
+    ('anywhere', 'aaaaaa', 40, 40),
+    ('break-word', 'aaaaaa', 40, 100),
+    ('normal', 'aaaaaa', 40, 100),
+))
+def test_overflow_wrap_2(wrap, text, body_width, expected_width):
+    page, = render_pages('''
+      <style>
+        @font-face {src: url(weasyprint.otf); font-family: weasyprint}
+        body {width: %dpx; font-family: weasyprint; font-size: 20px; }
+        table {overflow-wrap: %s; white-space: normal; }
+      </style>
+      <table><tr><td>%s
+    ''' % (body_width, wrap, text))
+    html, = page.children
+    body, = html.children
+    table_wrapper, = body.children
+    table, = table_wrapper.children
+    row_group, = table.children
+    tr, = row_group.children
+    td, = tr.children
+    assert td.width == expected_width
 
 
 def white_space_lines(width, space):
--- a/weasyprint/css/validation/properties.py
+++ b/weasyprint/css/validation/properties.py
@@ -1176,7 +1176,7 @@ def white_space(keyword):
 @single_keyword
 def overflow_wrap(keyword):
     """``overflow-wrap`` property validation."""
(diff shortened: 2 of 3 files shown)


liZe:
Hello @aschmitz,

Thanks a lot for this pull request, it’s just perfect 💜.

> Additionally, the `break-word` and `normal` tests would _seem_ as though they should be 120px wide, not 100, but I can't figure out why they have their current values: I'd be interested in thoughts there.

That’s a good question. We’ll check that and find why we get 100.

liZe:
The problem is caused by the trailing space whose size is removed from the minimum width. Looks like it’s removed twice :/.

I’ll remove the trailing space from the test, and add another test to check what’s wrong with it.

liZe:
Moreover, the first test [doesn’t pass with Pango 1.50.x](https://github.com/Kozea/WeasyPrint/runs/4503444577?check_suite_focus=true#step:10:143). We already had to report [a bug with tabs](#1514), let’s hope that this bug is also easy to reproduce and to fix…

liZe:
> We already had to report [a bug with tabs](#1514), let’s hope that this bug is also easy to reproduce and to fix…

That’s actually a bug in Pango 1.48.x, fixed in 1.50.x. The problem is caused by hyphens, automatically added when words are broken. With Pango 1.48.x, when the `aaaaa` string was rendered with no width available, we got:

```
a
a-
a-
a-
a
```

Notice the missing hyphen on the first line.

As, in WeasyPrint, we render lines one by one, the hyphen was always missing. That’s why we got a width of 1 character.

Now, in Pango 1.50.x, we get:

```
a-
a-
a-
a-
a
```

That’s better! But the minimum width is then 2 characters.

The solution is to simply disable hyphenation when we break words.

liZe:
The trailing space bug is now fixed.

As far as I can tell, `break-word` actually works too, doesn’t it?

aschmitz:
`break-word` isn't _really_ supported yet, although in many cases `overflow-wrap` could be used instead.

There are two cases where this is a concern, and I mostly have to handle in advance without knowing which one I'm going to be hitting:

1. Sometimes, there's a very long string that doesn't wrap naturally in a table cell (think a URL or enterprise Java class name). Right now, this requires `overflow-wrap: anywhere` because even with `overflow-wrap: break-word`, that string will count towards the table's minimum width, which overrides even an explicitly set width(!).
2. Other times, I have a nominally fairly narrow column and a column with a lot of text. (Think "product descriptions and prices".) If I apply `overflow-wrap: anywhere`, the column width algorithm will calculate that the minimum width of each column is about one letter wide, and feed the remaining width of the page to columns in proportion to their _maximum_ widths. That's very long for the product description, but very short for the price. As a result, the price column gets squeezed and all of the prices are forced to wrap, even though they wouldn't without `overflow-wrap: anywhere`.

Unfortunately I have no real way to know which is which upfront, since my users provide arbitrary content. I could set a minimum width for columns across the board, but that feels like a bit of a hack as well.

However, I can detect a long string that wouldn't otherwise wrap, and automatically enclose it with something like `<span style="word-break: break-all">abcdefghijklmnopqrstuvwxyz</span>`. That would let the layout engine wrap the _long_ words anywhere it needs to, and still respect shorter things.

(Of course the "real" solution is having people specify column widths if necessary, which we also support, but it would be very nice to not require everyone to do that in common cases.)

liZe:
> `break-word` isn't _really_ supported yet, although in many cases `overflow-wrap` could be used instead.
> 
> There are two cases where this is a concern, and I mostly have to handle in advance without knowing which one I'm going to be hitting: […]

I agree, `overflow-wrap: break-word` isn’t really useful for these use cases, and that’s much better with `overflow-wrap: anywhere`.

But, by "`overflow-wrap: break-word` actually works", I mean "`overflow-wrap: break-word` behaves accordingly to the specification". I think that it works correctly, now that Pango hyphens have been removed.

> However, I can detect a long string that wouldn't otherwise wrap, and automatically enclose it with something like `<span style="word-break: break-all">abcdefghijklmnopqrstuvwxyz</span>`. That would let the layout engine wrap the _long_ words anywhere it needs to, and still respect shorter things.

That would be the easiest way to handle long strings. Hyphenation is a nice solution for real words, but it’s quite useless for URLs for example.

aschmitz:
> But, by "`overflow-wrap: break-word` actually works", I mean "`overflow-wrap: break-word` behaves accordingly to the specification".

Yep, you're right. I was misinterpreting that as `word-break` the CSS property, not `break-word` the `overflow-wrap` value. `break-word` should be working correctly now as well.

Issue #1512, opened 2021-11-30
issue with UL or OL in rtl direction


engragy:
i can convert an HTML to pdf with this awesome solution (Weasyprint) as easy with direction "LTR", yet i notice there is limited support for "RTL".
the style orientation of ul/ol list items is left even though the ul and its li has the dir="rtl".
so is there a way around this limitation or can you tell me which file/part of the python code that may need a fix, may be i can help.

liZe:
Hello!

The position of the marker is set here:
https://github.com/Kozea/WeasyPrint/blob/8ae09175eb39d17a613d241b5a727eb40a11f2f7/weasyprint/formatting_structure/build.py#L317-L328

The code is supposed to handle rtl direction: the marker is translated to the right, by 100% of its own width, and that’s OK. The problem is that its original position is on the left of the parent, and it should be on the right instead.

To understand what’s wrong, you can render this code using a browser and WeasyPrint:

```html
<html dir="rtl">
  <div style="position: absolute; padding: 1em; background: green">abc</div>
</html>
```

Right-to-left layouts should put absolute boxes on the right. In WeasyPrint, it’s put on the left. As markers are absolute boxes, we have exactly the same problem.

The missing code to handle this is here (there’s even a TODO!):

https://github.com/Kozea/WeasyPrint/blob/8ae09175eb39d17a613d241b5a727eb40a11f2f7/weasyprint/layout/absolute.py#L53-L124

The direction is in the box style, you can see an example in another function:

https://github.com/Kozea/WeasyPrint/blob/8ae09175eb39d17a613d241b5a727eb40a11f2f7/weasyprint/layout/absolute.py#L307-L315

To fix the bug, we have to set `translate_box_width` and `translate_x` according to the direction. It requires some logic, but it should be possible in less than 10 lines of code. We’ll need tests too, but I can definitely write them later.

It’s actually a good issue to discover WeasyPrint’s code, don’t hesitate to ask if you have any questions!

liZe:
And for ordered lists, there’s another problem to fix:
https://github.com/Kozea/WeasyPrint/blob/8ae09175eb39d17a613d241b5a727eb40a11f2f7/weasyprint/formatting_structure/build.py#L307-L309

(There’s no need to panic, it can be fixed later! 😁)

Issue #1376, opened 2021-06-04
support for svg mix-blend-mode: multiply


danfitz36:
it's now supported in chrome, but it's new. any chance it can be supported here?

liZe:
Hello!

Thank you for this feature request. As many blend modes are supported in PDF (including multiply), we can imagine that this feature could be implemented. Of course, having the feature in PDF doesn’t mean that it’s easy to implement :smile:, but at least it’s a good start!

danfitz36:
I'm mostly a designer who codes some. is there anything else I can do to be helpful on something like this?

liZe:
> I'm mostly a designer who codes some. is there anything else I can do to be helpful on something like this?

Thanks a lot for your help!

The first steps are not really fun and require some ability to focus on a specification, but that where I’d start:
- download the [PDF specification](https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf),
- be sure that Blend modes (chapter 11.3.5) are what we want,
- find where the blend mode key has to be inserted in the PDF,
- find where the key has to be inserted in WeasyPrint (maybe somewhere [here](https://github.com/Kozea/WeasyPrint/blob/b24198df471e004469e4865ee958706068932586/weasyprint/document.py#L215)?),
- insert the `Multiply` key in the code and check that it works (no need to get the `mix-blend-mode` value from the SVG, that’s just a test),
- get the real `mix-blend-mode` value from the SVG and make it find its way to the place where it has to be inserted in WeasyPrint.

TL;DR: find a PDF sample in the spec where the `Multiply` (or any other) blend mode is set, try it in WeasyPrint’s code. If it works, make WeasyPrint use the real `mix-blend-mode` value instead of forcing `Multiply`. The first step could be complicated, but doesn’t require any coding abilities.

liZe:
Note that SVG’s `feBlend` is supported. Having `mix-blend-mode` supported in SVG (and actually even [in CSS](https://drafts.fxtf.org/compositing-1/#mix-blend-mode)) could actually be quite easy. If anyone is interested in contributing, just ask, we’ll help you!

PR #1522, opened 2021-12-13
fix #1007 Make 'inherit' value work correctly for shorthand properties

### LINKED ISSUE #1007: Make 'inherit' value work correctly for shorthand properties
From [Cascading and Inheritance](https://www.w3.org/TR/css-cascade-3/#shorthand):

> If a shorthand is specified as one of the CSS-wide keywords, it sets all of its sub-properties to that keyword […].

It should be probably handled by the `expander` decorator.

ckepper:
* skip shorthand expansion for 'inherit' tokens, hereby applying parent styles

This is my first contribution to WeasyPrint - I am not sure if this actually fixes the issue or just makes the error message go away.

### DIFF
--- a/tests/test_css_validation.py
+++ b/tests/test_css_validation.py
@@ -132,6 +132,11 @@ def test_spacing_invalid(rule):
     ('text-decoration-line: overline', {'text_decoration_line': {'overline'}}),
     ('text-decoration-line: overline blink line-through', {
         'text_decoration_line': {'blink', 'line-through', 'overline'}}),
+    ('text-decoration: none', {
+        'text_decoration_line': 'none',
+        'text_decoration_color': 'currentColor',
+        'text_decoration_style': 'solid'}),
+    ('text-decoration: inherit', { 'text_decoration': 'inherit'}),
 ))
 def test_decoration_line(rule, result):
     assert expand_to_dict(rule) == result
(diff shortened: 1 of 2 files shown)


liZe:
Hello!

> This is my first contribution to WeasyPrint

Thanks a lot for this PR!

> I am not sure if this actually fixes the issue or just makes the error message go away.

It makes the error message go away, but unfortunately it doesn’t fix the problem :/.

But there’s actually a decorator that already does what we want: `@generic_expander` is able to do what we want, and we have to use it everywhere. Unfortunately, there are lots of little details to take care about, so that’s not an easy fix for a first contribution. I can try to write a nice PR so that you can learn more about the validation system, and maybe open another PR in the future!

ckepper:
Understood. Thank you for responding so quickly. I will gladly analyze your PR when it's ready to learn more about how WeasyPrint works.

Are there other issues you would recommend to work on? I checked the [good first issue](https://github.com/Kozea/WeasyPrint/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) list but nothing else jumped at me right away.

liZe:
The easiest issue is probably to add the support of `inherit` to other properties. `flex-flow`, `page-break-inside` and `word-wrap` are probably the easiest ones.

Just like I did for `text-decoration`, you can add the `@generic_expander` decorator and replace the yielded values (often strings or numbers) by a list of real tokens. You can then test with a real document, and add a test if it works.

If it’s too complicated or too abstract (diving into WeasyPrint is a bit hard, that’s totally normal), I can give more information and/or add an example with more details in #1524.

#1512 and #1376 are also good issues to start, quite interesting with a small amount of code to write and visible results. They probably require some details and hints, but don’t hesitate to add a comment in these tickets if the topic is interesting for you!

PR #1524, opened 2021-12-14
Handle 'inherit' value for shorthand properties

### LINKED ISSUE #1007: Make 'inherit' value work correctly for shorthand properties

liZe:
Fix #1007.

The first commit b18b67a fixes `inherit` for `text-decoration`.

Using `@generic_expander` is a good idea, because it handles a lot of things, including `inherit` and `initial` automatically, and could handle other values like [`unset`](https://www.w3.org/TR/css-cascade-4/#inherit-initial) and [`revert`](https://www.w3.org/TR/css-cascade-4/#default). Using `@generic_expander` requires to yield tokens, not computed values, but that’s a fair trade-off and we should use it everywhere it’s possible.

We can then test the `inherit` value for all the expanders, as it’s now done for `text-decoration`.

If anyone (@ckepper?) is interested in updating this PR to handle other expanders, don’t hesitate to add a comment!

Here are the properties that aren’t / weren’t supported:

- [x] `text-decoration` (b18b67a)
- [x] `word-wrap` (#1525)
- [x] `page-break-after/before`
- [x] `border-radius`
- [x] `page-break-inside`
- [x] `line-clamp`
- [x] `flex-flow`
- [x] `flex`
- [x] `text-align`

### DIFF
--- a/tests/test_css_validation.py
+++ b/tests/test_css_validation.py
@@ -1104,6 +1104,7 @@ def test_page_break_invalid(rule):
 @assert_no_logs
 @pytest.mark.parametrize('rule, result', (
     ('page-break-inside: avoid', {'break_inside': 'avoid'}),
+    ('page-break-inside: inherit', {'break_inside': 'inherit'}),
 ))
 def test_page_break_inside(rule, result):
     assert expand_to_dict(rule) == result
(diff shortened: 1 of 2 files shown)


liZe:
@ckepper Thanks for your contribution, it will be merged soon 💚.

If you’re interested in these features, don’t hesitate to review this PR and give some feedback! If there’s anything you don’t understand, just ask, I’ll be happy to help.

PR #1525, opened 2021-12-14
#1007 Make 'inherit' value work correctly for `word-wrap`

ckepper:
and `overflow-wrap` properties


liZe:
That’s great, thank you!

Related to #1007.

PR #1534, opened 2022-01-05
Fix `overflow-wrap`, support `word-break: break-all`

aschmitz:
Previously, if `overflow-wrap: anywhere` (or `break-word`) was set on an inline element (like an `<a>` or `<span>`) whose first word occurred towards the end of a line, it would have broken that word, even if doing so wasn't necessary (and wouldn't have happened were it not for the wrapping element).

This pipes through a `is_line_start` argument to `split_text_box` that only allows the break on a word if it's the first word in a line.

Additionally, this adds support for the `break-all` value for `word-break`. The `keep-all` and `break-word` properties are not yet supported: the former seems like it is likely to require a Pango setting that I haven't yet dug into, and the latter has to set a value in another CSS property and I haven't dug into the CSS settings portion of WeasyPrint too much yet.

I'm not thrilled with the updated if statement in line_break.py, so feel free to modify it if you'd prefer different formatting.

This references #1153, but as above, doesn't fully close it yet. (There may not be a ton of work remaining, though!)

### DIFF
--- a/tests/test_text.py
+++ b/tests/test_text.py
@@ -929,6 +929,10 @@ def test_overflow_wrap(wrap, text, test, full_text):
     # This can be done accidentally if it is in its own inline element.
     ('overflow-wrap: anywhere', ['aaa', 'bbb']),
     ('overflow-wrap: break-word', ['aaa', 'bbb']),
+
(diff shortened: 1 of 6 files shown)


liZe:
Thanks for the pull request! 💜

PR #1541, opened 2022-01-17
fix: space out no-repeat patterns more

### LINKED ISSUE #1539: Tight background-repeat tolerances lead to erroneous display in some viewers
When providing a `background-repeat` value that indicates some repeating is allowed, WeasyPrint uses a PDF pattern, which must repeat in both directions.

This is handled by setting the repeat width to the bigger of the background image or the painting area:
https://github.com/Kozea/WeasyPrint/blob/8d2c549a82dfa0bef4891c78b18ffc447e0519d5/weasyprint/draw.py#L379-L382

However, because this means the pattern repeats *right* at the edge of the painting area, and some PDF viewers are very aggressive about their antialiasing(?) and not very picky about boundaries, this can lead to a blurred line of the background repeating at the edge of the containing box.

A small reproduction:
~~~html
<style>
  div {
    background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100'%3E%3Cpath d='M0 0h50v50H0z'/%3E%3C/svg%3E");
    background-repeat: repeat-y;
  }
</style>
<div>Hello world</div>
~~~

This produces [repeat-test.pdf](https://github.com/Kozea/WeasyPrint/files/7857933/repeat-test.pdf). A screenshot of the PDF in Firefox's PDF.js at an arbitrary zoom level:

Screenshot showing a very faint line at the right edge

It's very faint to see, but at the right edge of the page, there is a light grey line (zooming in with an image editor may be useful). This is dependent on the viewer: it appears that PDF.js and PDFium (Firefox + Chrome) will show a grey line at some zoom levels, Pango may not show it at any zoom levels(?), and I'm told that Adobe Acrobat Reader DC on Windows will display it at most/all zoom levels.

Most likely, the fix is to lift the code from later in the file up a bit:
https://github.com/Kozea/WeasyPrint/blob/8d2c549a82dfa0bef4891c78b18ffc447e0519d5/weasyprint/draw.py#L415-L418

I believe that block can be eliminated, and moved into lines 382 and 402 instead, and should resolve the issue. (It's not clear to me if there's a need to ensure that the `repeat_width` is at least as big as `image_width`, but if so, we should keep the `max()` call.)

If you'd like me to submit a PR for this, let me know. I have no idea how to add a test for it, unfortunately.

aschmitz:
This ensures the edges of a repeating pattern don't show up on the edge of the painting area in some PDF viewers. Closes #1539.

### DIFF
--- a/weasyprint/draw.py
+++ b/weasyprint/draw.py
@@ -378,8 +378,10 @@ def draw_background_image(stream, layer, image_rendering):
 
     if repeat_x == 'no-repeat':
         # We want at least the whole image_width drawn on sub_surface, but we
-        # want to be sure it will not be repeated on the painting_width.
-        repeat_width = max(image_width, painting_width)
+        # want to be sure it will not be repeated on the painting_width. We
+        # double the painting width to ensure viewers don't incorrectly bleed
+        # the edge of the pattern into the painting area. (See #1539.)
+        repeat_width = max(image_width, 2 * painting_width)
     elif repeat_x in ('repeat', 'round'):
         # We repeat the image each image_width.
         repeat_width = image_width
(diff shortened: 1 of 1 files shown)


liZe:
Thank you!

PR #1559, opened 2022-02-02
fix: handle float rounding when extending boxes over page breaks

aschmitz:
When a box would break over the edge of a page, its height is extended to the bottom of that page (per https://www.w3.org/TR/css-break-3/#box-splitting , primarily to allow backgrounds and borders to continue to the end of the page).

When this happened, sometimes the values that would be calculated for the height of the extended element would be rounded *over* the calculated height that remained on the page, forcing the entire containing box to wrap to the next page.

Rather than trying to carefully manage the order of operations to try to be safe in IEEE floats for directions, we apply a small "fudge factor": if an element fits very nearly (within a thousandth of a pixel) into the remaining space, it is still accepted.

### DIFF
--- a/tests/layout/test_page.py
+++ b/tests/layout/test_page.py
@@ -142,6 +142,40 @@ def test_page_breaks(html):
 
 
 @assert_no_logs
+def test_page_breaks_box_split():
+    # If floats round the wrong way, a block that gets filled to the end of a
+    # page due to breaking over the page may be forced onto the next page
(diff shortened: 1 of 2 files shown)


liZe:
Thanks for the pull request!

We’ll just change a bit the workaround to match the ones we already have:

https://github.com/Kozea/WeasyPrint/blob/21314bc726887b73469a5ac467e82feb009d0c52/weasyprint/layout/inline.py#L680-L686

PR #1566, opened 2022-02-10
fix: don't output footnotes before their call sites

### LINKED ISSUE #1564: Footnotes are sometimes placed on the page before the corresponding footnote call
In some cases, which appears to be "when a footnote call would normally be set on one page, and there would be room for the footnote, but a later rendering check pushes the footnote to the next page", a footnote can appear on the page before its call. (This violates the GCPM spec [which says](https://www.w3.org/TR/css-gcpm-3/#footnote-policy) "A footnote body must never be placed on a page before the footnote reference.")

A moderately minimal test case:

~~~html
<style>
  @page { @footnote { background-color: #afa; } }
  html {
    font-size: 10pt;
  }
  .footnote {
    float: footnote;
  }
  ::footnote-call {
    background-color: #aff;
  }
  div {
    height: 65em;
    background-color: #aaf;
  }
  p {
    break-inside: avoid;
    background-color: #faa;
  }
</style>
  
<div></div>

<p>
  foo bar baz foo bar baz foo bar baz foo bar baz foo bar baz foo bar baz foo
  bar baz foo bar baz foo bar baz foo bar baz foo bar baz
  <span class="footnote">footnote</span>
  foo bar baz foo bar baz foo bar baz foo bar baz foo bar baz foo bar baz foo
  bar baz foo bar baz foo bar baz foo bar baz foo bar baz foo bar baz foo bar
  baz
</p>
~~~

That produces [footnote-wrap.pdf](https://github.com/Kozea/WeasyPrint/files/8030099/footnote-wrap.pdf), which has a page break that looks like the following:

Screenshot of the PDF

The blue div is simply used to push the red paragraph to approximately the correct position. The issue is that the (light blue) footnote call occurs on the page after the (green) footnote.

My speculation is that this happens because the footnote call site *could* fit on the first page if it were allowed to break (removing the `break-inside` rule renders "correctly"). It does not appear to be related to the fact that the entire paragraph may fit on the first page if the footnote area did not exist, as making the paragraph much longer still results in the same behavior. However, other ways of forcing the paragraph to wrap to the next page (such as setting `orphans` or `widows` to 3) produce the same result: this isn't just an issue with `break-inside`.

I have not taken too much time to try to dig into why this occurs, and am mostly filing the issue in case you have ideas as to where to look. Otherwise I'll likely dig into this as time allows, possibly later this week.

aschmitz:
If a footnote has been output but we then decide to cancel the line it was output on, we should cancel the output of the footnote itself.

This requires a bit of extra bookkeeping, so we know which footnotes are relevant, but otherwise largely works using the existing remove_placeholders code.

Includes a minor refactor of footnote area management in LayoutContext, and a new testing utility: `tree_position`.

Closes #1564.

### DIFF
--- a/tests/layout/test_footnotes.py
+++ b/tests/layout/test_footnotes.py
@@ -8,7 +8,7 @@
 
 import pytest
 
-from ..testing_utils import assert_no_logs, render_pages
+from ..testing_utils import assert_no_logs, render_pages, tree_position
 
 
 @assert_no_logs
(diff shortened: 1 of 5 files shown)


aschmitz:
I'm certainly conscious of the fact that this PR is one of the more wide-reaching ones I've submitted. Definitely feel free to push back on any parts of it if you'd like them done different ways: I'm trying to fit things into the conventions I've seen elsewhere in the code, but clearly I haven't seen everything. :sweat_smile:

liZe:
Thanks **a lot** for this pull request. It’s a very good idea to use `remove_placeholders` to remove footnotes. Your code is always clean, always tested… 💜

PR #1570, opened 2022-02-14
fix: more robustly handle unlayout'ing a footnote

aschmitz:
This fixes a bug accidentally introduced in #1566, where we would try to unlayout a footnote that had not yet been laid out, if the algorithm decided that there would be insufficient space on a page before laying out a further footnote.

(The fix is to check to see whether we've actually laid it out yet - or, equivalently, put things back as if we haven't. Otherwise we get a ValueError from trying to remove from `self.reported_footnotes` and a duplicated entry in `self.footnotes`.)


liZe:
Thank you!

PR #1577, opened 2022-02-24
Add support for GitHub Codespaces / Devcontainers

mormahr:
I discovered a bug and wanted to contribute a fix, but found it difficult to setup a working local dev environment. That motivated me to use Codespaces instead, but the default devcontainer lacks the native dependencies and doesn't automatically install the required pip dependencies. I created a new devcontainer configuration based on the Python 3 template and added the native dependencies `libpango-1.0-0 libharfbuzz0b libpangoft2-1.0-0 fonts-dejavu ghostscript` to the image as well as add a post-create command that installs the Python dependencies.

With this image a new contributor can simply press "New Codespace", implement the change, run tests and open a PR.

In addition to the devcontainer, I'd also suggest adding a Visual Studio Code settings file that includes configuration for pytest, so that tests can be run from the UI. I can do that either in this PR or a follow up, if wanted.

Caveat:
I'm not that familiar with Python and Devcontainers specifically to know why it creates the `venv` in `.venv`, so I added that as an additional ignored path.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,21 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.222.0/containers/python-3/.devcontainer/base.Dockerfile
+
+# [Choice] Python version (use -bullseye variants on local arm64/Apple Silicon): 3, 3.10, 3.9, 3.8, 3.7, 3.6, 3-bullseye, 3.10-bullseye, 3.9-bullseye, 3.8-bullseye, 3.7-bullseye, 3.6-bullseye, 3-buster, 3.10-buster, 3.9-buster, 3.8-buster, 3.7-buster, 3.6-buster
+ARG VARIANT="3-bullseye"
+FROM mcr.microsoft.com/vscode/devcontainers/python:0-${VARIANT}
+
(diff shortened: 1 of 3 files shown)


liZe:
Hello!

First of all, thanks a lot for this pull request.

> I discovered a bug and wanted to contribute a fix, but found it difficult to setup a working local dev environment.

Creating a working dev environment is supposed to be painless, at least on Linux. What went wrong for you with the [documentation steps for contributing](https://doc.courtbouillon.org/weasyprint/stable/contribute.html)? Maybe we can improve the documentation on this point.

> With this image a new contributor can simply press "New Codespace", implement the change, run tests and open a PR.

This would be definitely useful for users who want to use Codespaces.

However, we won’t merge this pull request. Having configuration files for various tools in the repository looks like a good idea in the beginning, but it’s really hard to maintain for us. We like to keep the code with a minimum amount of files.

> In addition to the devcontainer, I'd also suggest adding a Visual Studio Code settings file that includes configuration for pytest, so that tests can be run from the UI. I can do that either in this PR or a follow up, if wanted.

There’s already some configuration for pytest in `pyproject.toml`. Doesn’t VSCode follow the configuration provided there?

liZe:
Feel free to add comments, even if it’s closed!

PR #1587, opened 2022-03-03
perf: cache calculations for 'ch' and 'ex' units

aschmitz:
I get a ~30% speedup from caching this calculation on a relatively large (~300 page) test document. We use a fair number of ch-based lengths, but not an absurd amount. Regardless, this seems to provide a significant speedup as measured by hyperfine:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Original | 67.992 ± 1.279 | 66.966 | 71.106 | 1.46 ± 0.04 |
| Cached | 46.707 ± 1.056 | 45.688 | 49.402 | 1.00 |

It wasn't immediately clear to me where a cache for these would be best, but this has it anchored at any style root rather than globally, as arguably the fonts might change between documents. (That said, actually handling `@font-face` fonts seems to be a TODO here at the moment anyway.)

### DIFF
--- a/weasyprint/css/__init__.py
+++ b/weasyprint/css/__init__.py
@@ -34,8 +34,8 @@ PSEUDO_ELEMENTS = (
     'footnote-call', 'footnote-marker')
 
 DEFAULT_CACHE = {
-    'length_ch': {},
-    'length_ex': {},
+    'ratio_ch': {},
+    'ratio_ex': {},
 }
 
 
--- a/weasyprint/css/__init__.py
+++ b/weasyprint/css/__init__.py
@@ -33,6 +33,11 @@ PSEUDO_ELEMENTS = (
     None, 'before', 'after', 'marker', 'first-line', 'first-letter',
     'footnote-call', 'footnote-marker')
 
+DEFAULT_CACHE = {
+    'length_ch': {},
+    'length_ex': {},
+}
+
 
 PageType = namedtuple('PageType', ['side', 'blank', 'first', 'index', 'name'])
 
@@ -600,6 +605,10 @@ class AnonymousStyle(dict):
         })
         self.parent_style = parent_style
         self.specified = self
+        if parent_style:
+            self.cache = parent_style.cache
+        else:
+            self.cache = DEFAULT_CACHE.copy()
 
     def copy(self):
         copy = AnonymousStyle(self.parent_style)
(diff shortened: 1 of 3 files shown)


liZe:
Thanks (again!) for this pull request.

> It wasn't immediately clear to me where a cache for these would be best, but this has it anchored at any style root rather than globally, as arguably the fonts might change between documents. (That said, actually handling `@font-face` fonts seems to be a TODO here at the moment anyway.)

Yes, I suppose that the cache should be stored in the `FontConfiguration` class. It would avoid one more "memory leak", and give the user the possibility to flush it if needed.

Moreover, the 1ch / 1em and 1ex / 1em ratios could probably be calculated once for each font face, and maybe without even using Pango (that’s needlessly expensive for what we want to achieve).

But it’s not easy to do. What do you think?

aschmitz:
Attaching the configuration to a `FontConfiguration` is fair, I'll investigate doing that when I get some time.

It wasn't immediately clear to me what could potentially affect the ratios for each type: certainly the features could (tabular figures could affect the 1ch size), style will, stretch will, weight will. I know that OpenType _hinting_ can change with font size, but I can't think of anything that would actually modify the glyphs themselves. (But, truthfully, since I couldn't rule it out, I included it in the cache.)

If we're fairly sure that the ratio doesn't change for a given font configuration across different sizes, it may be easy enough to simply cache that instead, as this is already doing it for 1ex/1em.

Doing the calculation without Pango seems like it could potentially be nice from a philosophical point of view, but with the caching it would be unlikely to be a huge time savings, so I'd probably wait until more things are shifted over to HarfBuzz rather than spend a bunch more effort on it here, since we already have nearly-working code, if that seems reasonable. (I could simply modify the 1ex/1em calculation to also handle 1ch/1em.)

aschmitz:
Digging a bit further, I'm not sure I can actually attach this to a FontConfiguration object the way styles are handled right now. I *think* the current style doesn't actually have a `context` object accessible to it, which is also the reason for all the "TODO: use context to use `@font-face` fonts" notes in computed_values.py.

Do you see a good way to get one in? If not, I can push the refactor to just cache a ratio for the ex and ch sizes for now, and get to the other refactor later

Presumably eventually you'll want to plumb through the context or at least a FontConfiguration object to the style, but it's not clear that this PR is necessary the right place to do that unless you see an easy way.

aschmitz:
This now has the ratio calculation code that lets our cache be independent of font size.

@liZe: does this seem like something you'd be willing to merge as-is, or do we need to plumb FontConfiguration / context in to style handling first?

liZe:
> This now has the ratio calculation code that lets our cache be independent of font size.
> 
> @liZe: does this seem like something you'd be willing to merge as-is, or do we need to plumb FontConfiguration / context in to style handling first?

It’s OK, thanks for your second commit. We don’t need `FontConfiguration` for now, we can keep the attribute on `style`. I’ll take the time to merge it soon, and maybe to change a couple of minor things to please the problems I have in my head 😁.

PR #1607, opened 2022-03-22
fix: calculate ex ratio correctly

aschmitz:
We had lost a negative sign accidentally in 4833bf69c8d9d278d66aef9e1e856f02122d42c8. Sorry about that!

I couldn't come up with a good way to test this off the top of my head especially given that it seems we can't necessarily rely on the testing font being available for use since we don't yet support `@font-face` in these computed values.


liZe:
Thank you.

> I couldn't come up with a good way to test this off the top of my head especially given that it seems we can't necessarily rely on the testing font being available for use since we don't yet support `@font-face` in these computed values.

Yes, that’s a problem with no perfect solution right now. We could at least test that the ex and ch units give "suitable" values to avoid obvious problems like this one. I’ll add these tests.

PR #1611, opened 2022-03-22
feat: add optional PDF metadata producer from HTML metadata

fgsalomon:


liZe:
Hi!

First of all, thank you for this pull request.

Choosing what to put in these tags required more time than one would imagine 😃, but the current situation seems to be the right one for us. The PDF specification gives this definition of the "Producer" field:

> If the document was converted to PDF from another format, the name of the conforming product that converted it to PDF.

That’s a good definition of WeasyPrint! There’s no need to lie and change this in our opinion.

If you want to put the name of the company in the PDF, the "Author" field is a good solution. If you want to include the name of the software that created your HTML document, the "Creator" PDF field (coming from the "generator" HTML field) is what you want:

> If the document was converted to PDF from another format, the name of the conforming product that created the original document from which it was converted.

And if you have some time to spend, interesting arguments are also given [in this Cairo issue](https://gitlab.freedesktop.org/cairo/cairo/-/issues/109).

Do you have a specific use case that would require to change this field?

fgsalomon:
Hi @liZe !

I forgot to add a message to the PR, sorry. First of all, thank you for your work in WeasyPrint. It's a great piece of software ❤️ 
I totally agree with your points, it's just in my specific use case I'm serving the generated PDFs and I just like to hide this information because I don't want to give clues about my server and its dependencies to possible attackers.

liZe:
> I forgot to add a message to the PR, sorry. First of all, thank you for your work in WeasyPrint. It's a great piece of software

No problem!
 
> I totally agree with your points, it's just in my specific use case I'm serving the generated PDFs and I just like to hide this information because I don't want to give clues about my server and its dependencies to possible attackers. I didn't know you could tell the PDF producer from its structure but I agree with what Paolo said in the Cairo issue, that security is not black and white.

Security is not black and white, for sure, but encouraging security through obscurity is not something we’d like to do. Moreover, having "WeasyPrint" in the metadata of a document generated by WeasyPrint seems to be a fair amount of advertising for a tool available for free 😀. (Of course, there are other ways to [support us](https://opencollective.com/courtbouillon)…)

So, we won’t merge this PR.

But, to be honest and transparent, there’s already a way to change the metadata of the PDF using the public API: you can give [a `finisher` function to `write_pdf`](https://doc.courtbouillon.org/weasyprint/stable/api_reference.html#weasyprint.HTML.write_pdf), and change whatever you want in the metadata including the "producer".

I bet you won’t need much time to find exactly how 😀.

Have fun with WeasyPrint!

TomaszGasior:
@liZe 

> But, to be honest and transparent, there’s already a way to change the metadata of the PDF using the public API: you can give [a finisher function to write_pdf](https://doc.courtbouillon.org/weasyprint/stable/api_reference.html#weasyprint.HTML.write_pdf), and change whatever you want in the metadata including the "producer".

As far as I understand, it's not possible for people who work with WeasyPrint as binary and not Python library – for example in apps developed in other languages, like PHP, when WeasyPrint is invoked through binary. 

While you may say security by obscurity is not good thing, the final decision about the shape of generated PDF document should be given to the one who develops end user's software. WeasyPrint should not force its name here.

liZe:
> As far as I understand, it's not possible for people who work with WeasyPrint as binary and not Python library – for example in apps developed in other languages, like PHP, when WeasyPrint is invoked through binary.

No, it’s not possible without a small Python script.

> While you may say security by obscurity is not good thing, the final decision about the shape of generated PDF document should be given to the one who develops end user's software.

If you really want security by obscurity, you really should use a third-party library to parse and rewrite the whole PDF. There are other ways to know that WeasyPrint generated the PDF, for example the :heart: gives you a simple way (`head document.pdf -n 2`).

> WeasyPrint should not force its name here.

That’s your opinion. As explained above:

> having "WeasyPrint" in the metadata of a document generated by WeasyPrint seems to be a fair amount of advertising for a tool available for free

Issue #1620, opened 2022-04-04
rtl numbered lists should have dot on the left


grewn0uille:
See https://github.com/Kozea/WeasyPrint/issues/1512#issuecomment-988262114

Thomasb81:
With following code
```
<!DOCTYPE html>
<html lang="en" dir="rtl">
<header>issue1620</header>
<body>
<ul>
    <li>one</li>
    <li>two</li>
    <li>three</li>
</ul>
</body>
</html>
```
Chrome, firefox and weasyprint, print dot on the right. Is Chrome and firefox are also wrong ?

grewn0uille:
Hello!

With unordered lists, the dot is on the right (with rtl), and it’s right.

When said the dot should be on the left, I meant the dot next to the number (for numbered lists). Like we should have "item .1" instead of "item 1."

Thomasb81:
> When said the dot should be on the left, I meant the dot next to the number (for numbered lists). Like we should have "item .1" instead of "item 1."

ok

so correct testcase:
```
<!DOCTYPE html>
<html lang="en" dir="rtl">
<header>issue1620</header>
<body>
<ol>
    <li>one</li>
    <li>ywo</li>
    <li>three</li>
</ol>
</body>
</html>
```

Thomasb81:
Hello

I understand that in context of above example following statement 
https://github.com/Kozea/WeasyPrint/blob/635f1818b5db7aeaaef196fdd286eb6f1281c882/weasyprint/css/counters.py#L292

should be : 
`    return prefix + suffix + value`

but I have difficulties to understand on what I should test against `style['direction'] == 'rtl'`. Because direction style seems not to be cascaded down to `<li> `box element.
I naively expect style to be inherit from parent.

liZe:
> should be :
> ` return prefix + suffix + value`

That’s more complicated than this (as explained in #1639, for the people interested in this issue.)

mahdisn76:
Any update on this?

liZe:
> Any update on this?

No. Would you be interested in working on this?

liZe:
Fixed as a duplicate of #2298.

PR #1639, opened 2022-05-07
Fix issue  #1620

Thomasb81:
Fix the issue
Add tests

### DIFF
--- a/tests/layout/test_list.py
+++ b/tests/layout/test_list.py
@@ -1,6 +1,7 @@
 """Tests for lists layout."""
 
 import pytest
+from weasyprint import HTML
 
 from ..testing_utils import assert_no_logs, render_pages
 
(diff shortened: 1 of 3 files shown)


tests/layout/test_list.py
```diff
@@ -103,3 +104,70 @@ def test_lists_page_break():
...
+def test_order_lists_rtl():
+    # Regression test for https://github.com/Kozea/WeasyPrint/issues/1620
+
+    def render_pages(html_content):
+        """ replace test render_page because style over-ride dir=rtl"""
+        return [
+            page._page_box for page in
+            HTML(string=html_content).render().pages]
+
+    check = [[". 1", "one"],
```
Thomasb81, line 148: Is this expected ? It seems firefox rending does not add space between dot and number...

liZe:
Hi!

First of all, thanks a lot for your pull request.

Issues related to right-to-left text are always complicated to solve. And as you have noticed, the rendering you get is not what we should get: we want " .1", not ". 1".

And unfortunately, this problem is not just a problem of reverting the order of the characters. It’s a real problem of [bidirectional text](https://www.unicode.org/reports/tr9/). Here is an example showing how complicated it is to handle things correctly:

```css
@counter-style wow {
  system: extends decimal;
  prefix: 'b- ';
  suffix: 'a. ';
}
```

This example gives:
68747470733a2f2f66696c65732e6769747465722e696d2f3565333832336132643733343038636534666438386465372f434c537a2f7468756d622f53637265656e73686f742d66726f6d2d323032322d30352d30382d30302d33362d30382e706e67

So… We’ll need to spend more time to understand deeply how rtl and bidi work, in order to fix problems like this (and many other ones).

PR #1695, opened 2022-07-22
Add language information to PDF

weddige:
This PR adds the language of the document to the catalog, if `<html lang="THE-LANGUAGE">` is set.

image

### DIFF
--- a/weasyprint/document.py
+++ b/weasyprint/document.py
@@ -104,7 +104,7 @@ class DocumentMetadata:
     """
     def __init__(self, title=None, authors=None, description=None,
                  keywords=None, generator=None, created=None, modified=None,
-                 attachments=None, custom=None):
+                 attachments=None, lang=None, custom=None):
         #: The title of the document, as a string or :obj:`None`.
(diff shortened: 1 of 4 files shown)


weasyprint/document.py
```diff
@@ -145,6 +145,9 @@ def __init__(self, title=None, authors=None, description=None,
         #: Extracted from the ``<link rel=attachment>`` elements in HTML
         #: and written to the ``/EmbeddedFiles`` dictionary in PDF.
         self.attachments = attachments or []
+        #: Document language as BCP 47 language tags.
+        #: Extracted from ``<html lang=lang>`` in HTML.
```
weddige, line 149:
```suggestion
        #: Document language, as a string or :obj:`None`.
        #: Extracted from ``<html lang=lang>`` in HTML
        #: and written as Lang to the document catalog.
```

liZe:
Thanks a lot for your contribution ❤️. Funny to see that we have worked on an equivalent feature for PDF/UA support!

PR #1703, opened 2022-08-07
Update the text-transform "capitalize" behaviour

VeteraNovis:
Transform the first typographic letter unit of each word to uppercase
without modifying other characters

See issue Kozea#1612

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,7 @@ dependencies = [
   'Pyphen >=0.9.1',
   'Pillow >=4.0.0',
   'fonttools[woff] >=4.0.0',
+  'regex >= 2022.7.25',
 ]
 classifiers = [
   'Development Status :: 5 - Production/Stable',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,6 @@ dependencies = [
   'Pyphen >=0.9.1',
   'Pillow >=4.0.0',
   'fonttools[woff] >=4.0.0',
-  'regex >= 2022.7.25',
 ]
 classifiers = [
   'Development Status :: 5 - Production/Stable',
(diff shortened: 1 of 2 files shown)


liZe:
Hi!

Thanks a lot for your contribution, it’s nice and seems to fix the original bug.

The only concern we have is adding a new external dependency. We’re trying hard to avoid new external libraries, moreover when they are written and C (like `regex`), because they can cause hard times to users.

It’s probably possible to get the same results using Pango. There’s no obvious solution in Pango’s API (ie. we’ll have to write some code 😀), but we already have the code getting text boundaries in [`get_next_word_boundaries`](https://github.com/Kozea/WeasyPrint/blob/f6a13ed636dd2a04ead8f0cb0211a2f4a14f1c92/weasyprint/text/line_break.py#L569).

VeteraNovis:
Hey mate. That's a good point.

I'll see what I can do to do away with the external dependency, and update the PR.

I've been using WeasyPrint for a while now, so it feels good to be able to contribute.

VeteraNovis:
> Hi!
> 
> Thanks a lot for your contribution, it’s nice and seems to fix the original bug.
> 
> The only concern we have is adding a new external dependency. We’re trying hard to avoid new external libraries, moreover when they are written and C (like `regex`), because they can cause hard times to users.
> 
> It’s probably possible to get the same results using Pango. There’s no obvious solution in Pango’s API (ie. we’ll have to write some code grinning), but we already have the code getting text boundaries in [`get_next_word_boundaries`](https://github.com/Kozea/WeasyPrint/blob/f6a13ed636dd2a04ead8f0cb0211a2f4a14f1c92/weasyprint/text/line_break.py#L569).

After doing more research, I don't think implementing this in regex will be feasible without implementing the regex module. The way that the default "re" module handles word boundaries is insufficient to deal with Unicode characters (outside of standard Latin characters).

For that reason, I have opted to utilise the already present `unicodedata` module to determine the Unicode general category of each character. The first typographic letter unit (any character in the Letter or Number category) in each word can now easily be determined, splitting based on the presence of a character from the Separator category.

liZe:
Thanks a lot!

We’ll merge your PR, clean a couple of things and add tests soon.

PR #1754, opened 2022-10-26
bump minimum version of Pillow to actual required minimum

eli-schwartz:
In commit a4fc7a11a90727195fd3edd781f2265aaf386a66 handling of Image.Transpose enum members was added, which is introduced in Pillow 9.1.0 as the superior way to handle constants:

https://pillow.readthedocs.io/en/latest/deprecations.html#constants

Previously, the minimum version was rather low. Make sure people cannot accidentally install WeasyPrint with too old Pillow.


eli-schwartz:
Done.

liZe:
Let’s merge, the bug in GitHub Actions is unrelated.

PR #1820, opened 2023-02-25
Adds a fuzz testing harness

jesslatimer:
This PR adds a basic fuzz test that targets HTML, CSS and write_pdf. It appears to have found at least two non-critical bugs so far, so I hope this will be of help to you. I will be uploading the inputs created by the fuzzer that caused the crashes. 
I have opened an issue that explains the benefits of fuzz testing here: https://github.com/Kozea/WeasyPrint/issues/1821


liZe:
See #1821.

PR #1867, opened 2023-04-24
Remove duplicate compression of attachments

timoramsauer:
Attachments are compressed twice on the current master branch. This results in the attachment not opening properly. At least Evince only uncompresses once. 
The function at fault is `write_pdf_attachment` in [anchors.py](https://github.com/Kozea/WeasyPrint/tree/master/weasyprint/pdf/anchors.py).

The function already compresses the attachment internally and then passes it to `Stream`, along with the `compress` argument set to `compress`. This is evaluated as `True` and the attachment is compressed again by `Stream`.

This pull request removes compression inside `write_pdf_attachment` and leaves compression to `Stream`

### DIFF
--- a/weasyprint/pdf/anchors.py
+++ b/weasyprint/pdf/anchors.py
@@ -2,7 +2,6 @@
 
 import hashlib
 import io
-import zlib
 from os.path import basename
 from urllib.parse import unquote, urlsplit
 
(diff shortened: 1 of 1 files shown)


timoramsauer:
There doesn't currently seem to be an option to store attachments uncompressed. Currently `zlib.compressobj()` is always passed and evaluated to `True`, which seems unintended.
Do you want compression of attachments to be an additional argument for `optimize_size` or should it rely on the `uncompressed_pdf` option?

liZe:
Thanks a lot!

PR #1874, opened 2023-05-09
Drawn underline and overline behind text

pointlessone:
Browsers draw underline/overline behind text and line-through on top of text. This is not very apparent in most cases because browsers also break underline to not collide with glyphs. WeasyPrint doesn't do that at the moment.

image

The screen shot above is made in Safari. [Codepen](https://codepen.io/cheba/pen/xxyWjoZ) with demonstration.

### DIFF
--- a/tests/draw/test_text.py
+++ b/tests/draw/test_text.py
@@ -520,9 +520,9 @@ def test_text_underline(assert_pixels):
     assert_pixels('''
         _____________
         _zzzzzzzzzzz_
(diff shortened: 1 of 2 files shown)


liZe:
> Fixed tests but I'm not sure if that's the best way to go about it. Given that wasyprint font is solid blocks I don't know how else to reveal decorations that fully covered by text other than using transparency.

Thanks a lot for this PR! 💜

We’ll check the code and the tests soon!

Issue #1278, opened 2020-12-29
Collapsed borders not drawn for tables in margin boxes


marcosn41:
I made a complex document in it and I'm putting header and footer in @ top-center and @ top-bottom. The problem is that I use a simple 1-row table. On the first page everything is ok, on the second onwards ONLY the border does not work is borderless the rest everything is ok. This only occurs in the header and footer in the rest of the document is ok

First page:
![image](https://user-images.githubusercontent.com/14874674/103299575-9f0c9700-49db-11eb-9e2a-36da1640e57f.png)

Second one:
![image](https://user-images.githubusercontent.com/14874674/103299613-afbd0d00-49db-11eb-8849-d2e2f325b23b.png)

Code:

 ```
<div class="header">
         <table style="border-collapse: collapse; width: 100%; height: 84px;">
             <tbody class="bodyHeader">
                 <tr style="height: 21px;">
                     <td style="width: 12.1754%; height: 21px; background-color: cornflowerblue;" rowspan="4">&nbsp;</td>
                     <td style="width: 64.5104%; height: 21px;" rowspan="2">&nbsp;</td>
                     <td style="width: 19.905%; height: 21px;">&nbsp;</td>
                 </tr>
                 <tr style="height: 21px;">
                     <td style="width: 19.905%; height: 21px;">&nbsp;</td>
                 </tr>
                 <tr style="height: 21px;">
                     <td style="width: 64.5104%; height: 21px;" rowspan="2">&nbsp;</td>
                     <td style="width: 19.905%; height: 21px;">&nbsp;</td>
                 </tr>
                 <tr style="height: 21px;">
                     <td style="width: 19.905%; height: 21px;">&nbsp;</td>
                 </tr>
             </tbody>
         </table>
         
     </div>
```

```
div.header {
            /* display: block; */
            width: 100%;
            position: running(header);
            margin: 0mm;
            padding: 0mm;
        }
        @page iretrato{
            size: A4;
            margin: 2cm;
            margin-top: 3.5cm;

            @top-center {
                vertical-align: top;
                margin: 0mm;
                padding-top: 1cm;
                content: element(header);
                width: 100%;
            }

            @bottom-center {
                content: element(footer);
            }    
        }
```

marcosn41:
Ok, i found why this happening, its border-collapse: collapse. I dont know if this is a bug because only shows border on first page but my problem is solved removing this param.

liZe:
[...]

PR #1884, opened 2023-05-21
Propose fix for issue #1278

kygoh:
Before these properties are overwritten in the `set_transparent_border` function:
- `border_{side}_style`
- `border_{side}_width`
- `border_{side}_color`

a copy of the original value is stored in the `style` dictionary using property name prefixed with double underscore `__`:
- `__border_{side}_style`
- `__border_{side}_width`
- `__border_{side}_color`

When `collapse_table_border` is called each time the boxes of the table are built, the original property values are restored first, if exists.


liZe:
Hi!

Thanks a lot for your proposal. 💜

> a copy of the original value is stored in the `style` dictionary using property name prefixed with double underscore `__`

That’s something I would really like to avoid 😢. Adding "ghost" style properties is a workaround that often leads to other problems in the future, and even whet it seems harmless it’s often a trap set for future developers.

As explained in [this comment](https://github.com/Kozea/WeasyPrint/issues/1278#issuecomment-1188767790) and the next one, the solution is to keep the real style and to change both the `layout.table` module to handle correct border widths and the `draw` module to draw correct colors. Even if it may be really appealing, any other solution is a workaround on top of a buggy workaround, and that’s the kind of things that often give maintenance nightmares.

Would you be interested in trying to find such a solution?

kygoh:
> Adding "ghost" style properties is a workaround that often leads to other problems in the future

I am aware of the technical debt and the reason I only _propose_ this PR rather than _submit_ it.

Since I've created the test case to simulate the issue, the "ghost" style properties workaround was used to pass the test until a proper solution can be implemented.

> Would you be interested in trying to find such a solution?

I would be interested to try but this library is still new to me and I am still unclear of its inner workings.

kygoh:
> `layout.table` module to handle correct border widths

Do you mean that these lines should be removed:

https://github.com/Kozea/WeasyPrint/blob/0381529724d1c8972d1b71bda3e61a45a795dba7/weasyprint/formatting_structure/build.py#L979-L981

and the `collapse_table_borders` border conflicts resolution logic be handled in `layout.table` module?

Then, `grid_width` and `grid_height` values need to be passed to `layout.table`?

I think the `table.collapsed_border_grid` tuple is only used in 2 places:

https://github.com/Kozea/WeasyPrint/blob/0381529724d1c8972d1b71bda3e61a45a795dba7/weasyprint/draw.py#L816

https://github.com/Kozea/WeasyPrint/blob/0381529724d1c8972d1b71bda3e61a45a795dba7/weasyprint/layout/table.py#L67

and for test cases:

https://github.com/Kozea/WeasyPrint/blob/0381529724d1c8972d1b71bda3e61a45a795dba7/tests/test_boxes.py#L22

> `draw` module to draw correct colors

If the structure of `(vertical_borders, horizontal_borders)` can be preserved, then the following won't need to change?

https://github.com/Kozea/WeasyPrint/blob/0381529724d1c8972d1b71bda3e61a45a795dba7/weasyprint/draw.py#L862

https://github.com/Kozea/WeasyPrint/blob/0381529724d1c8972d1b71bda3e61a45a795dba7/weasyprint/draw.py#L884

liZe:
> Do you mean that these lines should be removed and the `collapse_table_borders` border conflicts resolution logic be handled in `layout.table` module?

Yes, exactly.

> Then, `grid_width` and `grid_height` values need to be passed to `layout.table`?

Yes.

> If the structure of `(vertical_borders, horizontal_borders)` can be preserved, then the following won't need to change?

That would be great!

More changes are probably required for the layout than for the drawing, because the drawing part is already different for collapsed and separate cells.

kygoh:
Please take a look at the latest changes that omitted `border_{side}_style` and `border_{side}_color` properties from `set_transparent_border` yet passed all test cases.

At this point is it safe to assume only the change of `border_{side}_width` in the box style dictionary causes issue #1278?

From your https://github.com/Kozea/WeasyPrint/issues/1278#issuecomment-1188773452:

> we also probably need to change the `layout.table` module to use the right value of the border width

I think the right value will be:

https://github.com/Kozea/WeasyPrint/blob/0381529724d1c8972d1b71bda3e61a45a795dba7/weasyprint/formatting_structure/build.py#L1102

If the `border_{side}_width` is omitted from `set_transparent_border`, among the tests that failed are:
```
FAILED tests/layout/test_table.py::test_layout_table_fixed_5 - assert 10.0 == 5
FAILED tests/layout/test_table.py::test_layout_table_auto_17 - assert 10.0 == 5
```
due to:
```
assert table.border_left_width == 5  # half of the collapsed 10px border
```

Can you point out where `table.border_left_width` value is set?

kygoh:
> These attributes are [used values](https://www.w3.org/TR/css-cascade-5/#used):
>
> > The used value is the result of taking the [computed value](https://www.w3.org/TR/css-cascade-5/#computed-value) and completing any remaining calculations to make it the absolute theoretical value used in the formatting of the document.

https://github.com/Kozea/WeasyPrint/blob/0381529724d1c8972d1b71bda3e61a45a795dba7/docs/going_further.rst?plain=1#L225-L226

Based on the above, is the following correct?

- `table.style['border-left-width']` contains computed value
- **used value** `table.border_left_width` can be "_set_" from **computed value** `table.style['border-left-width']`

> the solution is to change these attributes instead of the style dictionary (that are computed values)

meaning:

https://github.com/Kozea/WeasyPrint/blob/0381529724d1c8972d1b71bda3e61a45a795dba7/weasyprint/formatting_structure/build.py#L1102

should instead be:

```py
    setattr(box, f'border_{side}_width', twice_width / 2)
```

liZe:
> Based on the above, is the following correct?

Yes, it is.

> At the following point?

`resolve_percentage` is called for each box (table, row, cell…), in different parts of `layout.table`. Here, it’s just for the table, but it’s also done for rows, cells, etc.

> meaning `box.style[f'border_{side}_width'] = twice_width / 2` should instead be `setattr(box, f'border_{side}_width', twice_width / 2)`

That’s the idea, but it may not be that simple. The current bad solution replaces the style during the build step (that builds the boxes), so that the layout step (that sets the position and the size of each box) is sure to use the values that result from the collapsing algorithm. Now, we have to set the used values during the layout step.

We have to set the collapsed values after almost each `resolve_percentages` call in `layout.table`. It means that [this code](https://github.com/Kozea/WeasyPrint/blob/0381529724d1c8972d1b71bda3e61a45a795dba7/weasyprint/formatting_structure/build.py#L1097-L1152) can be removed, and replaced by equivalent calls after almost each call to `resolve_percentages` in `layout.table`.

Here’s what you can try:
1. Remove [this code](https://github.com/Kozea/WeasyPrint/blob/0381529724d1c8972d1b71bda3e61a45a795dba7/weasyprint/formatting_structure/build.py#L1097-L1152).
2. Write equivalent functions (`remove_borders`, and `set_collapsed_border` instead of `set_transparent_border` as we only need change the width) in `layout.table`, that will set used values (`setattr(box, f'border_{side}_width', twice_width / 2)`) instead of computed values.
3. Find the `resolve_percentages` calls in `layout.table` that are called on the same boxes as the previous `remove_borders` or `set_transparent_border`. Call `remove_borders` or `set_collapsed_border` after the corresponding `resolve_percentages` call (only when `border-collapse` is set to `collapse`, of course), so that we have the right used value as soon as possible.

If some tests don’t pass after this, it probably means that we use computed values instead of used values somewhere during the layout. But maybe you’ll be lucky and everything will work as expected!

kygoh:
The problem encountered with "experiment" using _used values_ over _computed values_ for `border-{...}-width`:
```pycon
================================================================ short test summary info ================================================================
FAILED tests/layout/test_table.py::test_layout_table_fixed_5 - assert 32.0 == 34
=========================================================== 1 failed, 112 deselected in 2.01s ===========================================================
```
at
https://github.com/Kozea/WeasyPrint/blob/21810bad84007aec00fdc240251356367e3d6da7/tests/layout/test_table.py#L239
for HTML:
https://github.com/Kozea/WeasyPrint/blob/21810bad84007aec00fdc240251356367e3d6da7/tests/layout/test_table.py#L206-L223

Tracing the source code for the failed test, I think:
1. When [`table_wrapper_width`](https://github.com/Kozea/WeasyPrint/blob/master/weasyprint/layout/table.py#L804) is called, the `border-{...}-width` _used values_ for cells have not been harmonized nor set.
2. The table width is derived from `table_min_content_width` value at:
https://github.com/Kozea/WeasyPrint/blob/21810bad84007aec00fdc240251356367e3d6da7/weasyprint/layout/table.py#L708-L712
3. Stepping through the code for how `table_min_content_width` and `table_max_content_width` are derived will eventually  lead to the [`margin_width`](https://github.com/Kozea/WeasyPrint/blob/21810bad84007aec00fdc240251356367e3d6da7/weasyprint/layout/preferred.py#L134-L159) function which is referencing _computed values_:
https://github.com/Kozea/WeasyPrint/blob/21810bad84007aec00fdc240251356367e3d6da7/weasyprint/layout/preferred.py#L150-L153
4. Putting breaks in the source shows that at
https://github.com/Kozea/WeasyPrint/blob/21810bad84007aec00fdc240251356367e3d6da7/weasyprint/layout/table.py#L777
5. `table.column_widths = [ 30, 60 ]` for _used values_ vs `table.column_widths = [ 30, 58 ]` for _computed values_ due to:
    ```
    54 (width) + 1 (border-left-width) + 5 (border-right-width) = 60
    54 (width) + 2 (border-left-width) + 2 (border-right-width) = 58
    ```

Any pointers?

kygoh:
> 1. Remove [this code](https://github.com/Kozea/WeasyPrint/blob/0381529724d1c8972d1b71bda3e61a45a795dba7/weasyprint/formatting_structure/build.py#L1097-L1152).

This code is not so straightforward to remove as the `border_{left,right}_width` _computed values_ are being used to determine preferred margin width:

https://github.com/Kozea/WeasyPrint/blob/21810bad84007aec00fdc240251356367e3d6da7/weasyprint/layout/preferred.py#L150-L153

Thus, I've made "smaller" modifications first until I can better understand the logic:
1. Instead of using `__border_{...}_width` to "backup" the original values, new _computed values_ `collapse_border_{...}_width` is used to store the harmonized collapsed border widths (commit 9e59b20f3)
2. These new computed values `collapse_border_{...}_width` are set as _used values_ `border_{...}_width` in [`resolve_percentages`](https://github.com/Kozea/WeasyPrint/blob/0381529724d1c8972d1b71bda3e61a45a795dba7/weasyprint/layout/percent.py#L51) (commit 800a683)
3. These new computed values `collapse_border_{...}_width` are also used in preferred margin width calculation when applicable (commit d87034f)

I believe _used values_ will not be available yet when [`margin_width`](https://github.com/Kozea/WeasyPrint/blob/21810bad84007aec00fdc240251356367e3d6da7/weasyprint/layout/preferred.py#L134-L159) function is called as described in earlier https://github.com/Kozea/WeasyPrint/pull/1884#issuecomment-1561390937

kygoh:
> 3\. [...] so that we have the right used value as soon as possible

Continuing from the latest commit, which have passed all tests, if I naively change `margin_width` to _used values_:
```py
    if left:
        width += box.border_left_width
    if right:
        width += box.border_right_width
```
the result for [`tests/layout/test_table.py::test_layout_table_fixed_5`](https://github.com/Kozea/WeasyPrint/blob/21810bad84007aec00fdc240251356367e3d6da7/tests/layout/test_table.py#L203-L242) :
```pycon
================================================================ short test summary info ================================================================
FAILED tests/layout/test_table.py::test_layout_table_fixed_5 - AttributeError: 'TableColumnBox' object has no attribute 'border_left_width'. Did you mean: 'border_width'?
```
with the relevant stack trace:
```console
...
weasyprint/layout/table.py:712: in auto_table_layout
    table_and_columns_preferred_widths(context, box, outer=False)
weasyprint/layout/preferred.py:448: in table_and_columns_preferred_widths
    min_content_width(context, groups[i]))
weasyprint/layout/preferred.py:47: in min_content_width
    return block_min_content_width(context, box, outer)
weasyprint/layout/preferred.py:180: in block_min_content_width
    return _block_content_width(
weasyprint/layout/preferred.py:104: in _block_content_width
    return adjust(box, outer, width)
weasyprint/layout/preferred.py:173: in adjust
    return margin_width(box, fixed, left, right)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

box = <TableColumnBox col>, width = 30.0, left = True, right = True

    def margin_width(box, width, left=True, right=True):
        """Add box paddings, borders and margins to ``width``."""
        percentages = 0
    
        for value in (
            (['margin_left', 'padding_left'] if left else []) +
            (['margin_right', 'padding_right'] if right else [])
        ):
            style_value = box.style[value]
            if style_value != 'auto':
                if style_value.unit == 'px':
                    width += style_value.value
                else:
                    assert style_value.unit == '%'
                    percentages += style_value.value
    
        collapse = box.style['border_collapse'] == 'collapse'
        if left:
>           width += box.border_left_width
E           AttributeError: 'TableColumnBox' object has no attribute 'border_left_width'. Did you mean: 'border_width'?
```

From the stack trace, it's obvious that `resolve_percentages` has not been called to set the _used values_ for `TableColumnBox`.

If we were to stick to just _used values_ for `border-{...}-width`, could this lead to a cyclic problem ie table width requires columns and cells widths (which haven't been resolved) but columns and cells widths need table width (which can't be resolved)?

kygoh:
Please review the latest commit:

1. `collapse_table_border` has been moved to `layout.table` to better reflect its functionality
2. instead of overwriting border-{...}-width _computed values_ during collapsed-borders conflict resolution, they are set as _used values_
3. modification appears to pass all tests.

While I think the code can be refactored further, I will appreciate if you can merge this fix for the next release.

kygoh:
@liZe please review the latest commit:

1. The _computed_ border styles are now left untouched so restoring original values with "ghost" style properties are not required anymore (commit 9e59b20) 
2. With the _used_ border widths being set, these are also being used to compute cell and table intrinsic offsets as stated in [3.8.2. Computing Cell Measures](https://drafts.csswg.org/css-tables-3/#computing-cell-measures) (commit bb382b5)
3. Codes and test cases related to border-collapse have been moved to `layout.table`
4. However, the draw module is still using the winning border properties `table.collapsed_border_grid` rather than _used values_ - probably to go back after gaining more familiarity.

I hope this fix can be merged while we continue improving the maintainability of border collapse codes.

liZe:
Sorry for the delay, we’re currently quite busy with other projects! We’ll review and hopefully merge this PR as soon as possible.

kygoh:
> 1. The _computed_ border styles are now left untouched so restoring original values with "ghost" style properties are not required anymore (commit [9e59b20](https://github.com/Kozea/WeasyPrint/commit/9e59b20f3ad23a55d274b000c211ed59a11fd069))

To summarize the key areas of the fix:

The original `set_transparent_border` causes border styles to be modified:

https://github.com/Kozea/WeasyPrint/blob/4e9ad95499bc5a7cf4baf0fbefd5826c533ff8ca/weasyprint/formatting_structure/build.py#L1100-L1103

so that `margin_width` uses the new collapsed width:

https://github.com/Kozea/WeasyPrint/blob/4e9ad95499bc5a7cf4baf0fbefd5826c533ff8ca/weasyprint/layout/preferred.py#L150-L153

and later store as used values in `resolve_percentages`:

https://github.com/Kozea/WeasyPrint/blob/4e9ad95499bc5a7cf4baf0fbefd5826c533ff8ca/weasyprint/layout/percent.py#L95-L98

This fix stores the collapsed border width in used values:

https://github.com/Kozea/WeasyPrint/blob/af141bac1e677f2e2863150b336278a6be52fb7b/weasyprint/layout/table.py#L1082-L1084

and modifies `margin_width` to get the collapsed width from used values, if set:

https://github.com/Kozea/WeasyPrint/blob/af141bac1e677f2e2863150b336278a6be52fb7b/weasyprint/layout/preferred.py#L154-L171

and skip storing as used values again in `resolve_percentages`:

https://github.com/Kozea/WeasyPrint/blob/af141bac1e677f2e2863150b336278a6be52fb7b/weasyprint/layout/percent.py#L95-L102

liZe:
Hi!

Thanks for your work on this issue 💜.

I’ve changed the tests to use drawing, as it’s easier to debug, and to be sure that the rendering is what we actually expect from the layout.

(Images are generated in `tests/draw/results/*` to check what’s going on.)

It looks like there are differences between what we have and what we expect:

- For `test_running_elements_table_border_collapse` and `test_running_elements_table_border_collapse_span`, there are spaces after the text. We don’t have that with the previous version, there must be something wrong.
- `test_running_elements_table_border_collapse_border_style` seems to be buggy. As it was also buggy with previous version, I’ve marked it as xfail.
- `test_running_elements_table_border_collapse_margin` shows that the margin doesn’t work when repeated. It was buggy in the previous version, it’s also marked as xfail.

So, what’s important is to find where this space comes from, other bugs can be fixed later. I’ll check the PR in details when these two bugs pass.

kygoh:
> For `test_running_elements_table_border_collapse` and `test_running_elements_table_border_collapse_span`, there are spaces after the text. We don’t have that with the previous version, there must be something wrong.

The space arises from moving the collapsed-borders conflict resolution from `formatting_structure.build.wrap_table` (called before [`make_margin_boxes`](https://github.com/Kozea/WeasyPrint/blob/ce46dbe5a070f231979cfe2449b1d7221f363ed2/weasyprint/layout/page.py#L354-L485)):

https://github.com/Kozea/WeasyPrint/blob/9e59b20f3ad23a55d274b000c211ed59a11fd069/weasyprint/formatting_structure/build.py#L979-L981

to `layout.table.table_wrapper_width` (called after [`make_margin_boxes`](https://github.com/Kozea/WeasyPrint/blob/ce46dbe5a070f231979cfe2449b1d7221f363ed2/weasyprint/layout/page.py#L354-L485)):

https://github.com/Kozea/WeasyPrint/blob/2a3747ac5245442c1d0aac8f23775ce4f457b270/weasyprint/layout/table.py#L812-L817

The used border-width values of the cell (half the winning border-width) is used by [`make_margin_boxes`](https://github.com/Kozea/WeasyPrint/blob/ce46dbe5a070f231979cfe2449b1d7221f363ed2/weasyprint/layout/page.py#L354-L485) to determine the cells and table width:

https://github.com/Kozea/WeasyPrint/blob/ce46dbe5a070f231979cfe2449b1d7221f363ed2/weasyprint/layout/page.py#L445-L447

through `margin_width`:

https://github.com/Kozea/WeasyPrint/blob/ce46dbe5a070f231979cfe2449b1d7221f363ed2/weasyprint/layout/preferred.py#L154-L171

Restoring the collapsed-borders conflict resolution to its original place seems to fix the two bugs.

tests/layout/test_table.py
```diff
@@ -2988,120 +2973,3 @@ def test_min_width_with_overflow():
...
-blue = (0, 0, 1, 1)
-yellow = (1, 1, 0, 1)
-black_3 = ('solid', 3, black)
-red_1 = ('solid', 1, red)
-yellow_5 = ('solid', 5, yellow)
-green_5 = ('solid', 5, green)
-dashed_blue_5 = ('dashed', 5, blue)
-
-
-@assert_no_logs
```
kygoh: Please consider preserving these test cases as unit test on [conflict resolution logic for collapsed borders](https://drafts.csswg.org/css-tables-3/#conflict-resolution-for-collapsed-borders)

kygoh:
> what’s important is to find where this space comes from

Explained in https://github.com/Kozea/WeasyPrint/pull/1884#issuecomment-1714731599 and fix has been pushed

> * `test_running_elements_table_border_collapse_margin` shows that the margin doesn’t work when repeated

Appears to be a "repeat" of issue #1278 but for margins and the "ugly fix" to keep the original margins style:

```diff
diff --git a/weasyprint/formatting_structure/build.py b/weasyprint/formatting_structure/build.py
index b247eb19..6663214c 100644
--- a/weasyprint/formatting_structure/build.py
+++ b/weasyprint/formatting_structure/build.py
@@ -994,6 +994,13 @@ def wrap_table(box, children):
     # of the wrapper and the table. The other get the initial value.
     # TODO: put this in a method of the table object
     for name in properties.TABLE_WRAPPER_BOX_PROPERTIES:
+        if name in ['margin_left', 'margin_right', 'margin_top', 'margin_bottom']:
+            if f'*-{name}' in table.style:
+                # restore original style
+                table.style[name] = table.style[f'*-{name}']
+            else:
+                # backup original style
+                table.style[f'*-{name}'] = table.style[name]
         wrapper.style[name] = table.style[name]
         table.style[name] = properties.INITIAL_VALUES[name]
```

kygoh:
To recap, the root cause of issues (#1278 and https://github.com/Kozea/WeasyPrint/issues/2013#issuecomment-1840915969) boil down to _computed values_ in running tables are being overwritten, `box.style` at:

https://github.com/Kozea/WeasyPrint/blob/e018bf3fbb13a05d2cfb647cb28cc4c0434558b0/weasyprint/formatting_structure/build.py#L1094-L1100

and `table.style` at:

https://github.com/Kozea/WeasyPrint/blob/e018bf3fbb13a05d2cfb647cb28cc4c0434558b0/weasyprint/formatting_structure/build.py#L988-L993

liZe:
@kygoh Thanks a lot 💜

PR #1887, opened 2023-05-24
Fix footnote-call displayed incorrectly for some fonts

### LINKED ISSUE #1876: Footnote-call displayed incorrectly for some fonts
For some fonts footnote calls are displayed the same as regular text and not as superscript. The problem is that some fonts don't support `font-variant-position: super`, which is used in the default weasyprint CSS to style footnote calls. I have tested only a few fonts, but in general `font-variant-position: super` seems work well variable fonts it seems to work correctly, but not for static fonts. `sup` element, however, are displayed as expected for all fonts.

I think weasyprint should use the same styling rules for footnote calls as for `<sup>` such that they are displayed correctly for all fonts.

This is also recommended in CSS Fonts Module Level 4 (https://drafts.csswg.org/css-fonts/#font-variant-position-prop):
> Because of these limitations, font-variant-position is not recommended for use in user agent stylesheets.

Screenshot_20230511_093746

```html
<html>
<head>
  <style>
    @import url('https://fonts.googleapis.com/css2?family=Noto+Sans:wght@200..900&family=Tinos&display=swap');
    .font-tinos {
        font-family: "Tinos";
    }
    .font-noto {
        font-family: "Noto Sans";
    }
    .footnote {
      float: footnote;
    }
  </style>
</head>
<body>
  <div class="font-tinos">
    <p>footnote<span class="footnote">footnote</span></p>
    <p>font-variant-position: super<span style="font-variant-position: super">1</span></p>
    <p>sup<sup>1</sup></p>
  </div>
  <div class="font-noto">
    <p>footnote<span class="footnote">footnote</span></p>
    <p>font-variant-position: super<span style="font-variant-position: super">2</span></p>
    <p>sup<sup>2</sup></p>
  </div>
</body>
</html>
```

MWedl:
Fix #1876

### DIFF
--- a/weasyprint/css/html5_ua.css
+++ b/weasyprint/css/html5_ua.css
@@ -616,7 +616,7 @@ var { font-style: italic; }
 video { object-fit: contain; }
 xmp { display: block; font-family: monospace; margin-bottom: 1em; margin-top: 1em; /* unicode-bidi: isolate; */ white-space: pre; }
 
-::footnote-call { content: counter(footnote); vertical-align: baseline; font-size: 100%; line-height: inherit; font-variant-position: super; }
+::footnote-call { content: counter(footnote); vertical-align: super; font-size: smaller; line-height: inherit; }
 ::footnote-marker { content: counter(footnote) '. '; }
 
 @page {


liZe:
Thanks a lot!

Issue #58, opened 2013-03-21
New page-margin boxes layout algorithm


SimonSapin:
Catch up with spec changes: http://dev.w3.org/csswg/css3-page/#margin-dimension

galbwe:
@liZe Hi, I'd like to get started contributing to this project. I'm still familiarizing myself with the codebase, and am not 100% certain about which part of the code to modify to address this issue. I think I need to edit the ```compute_variable_dimension``` function in ```layout/pages.py```, based on a search for the w3 url. Is this on the right track? Thanks.

liZe:
> Hi, I'd like to get started contributing to this project.

Thank you! Don't hesitate to ask if you have any questions, I'll do my best to help you.

> I think I need to edit the `compute_variable_dimension` function in `layout/pages.py`, based on a search for the w3 url. Is this on the right track?

Yes, it is.

In 2012, when @SimonSapin implemented the page-margins layout, the specification was this one:
https://www.w3.org/TR/2006/WD-css3-page-20061010/#margin-dimension

The algorithm was really hard to implement and didn't give good results. Simon decided to implement something else, and he proposed a replacement of the specification, based on what he implemented in WeasyPrint:
https://github.com/SimonSapin/css/blob/8c17fbe/margin-boxes-variable-dimension

This proposition was accepted, probably with some changes (I don't remember, maybe Simon does). So the first step is to find what's different between the current specification and the current implementation in `compute_variable_dimension`, and then fix these differences.

You can spend some time reading [Chapter 5.3 of the specification](https://www.w3.org/TR/css-page-3/#margin-dimension). It's totally normal to find it over-complicated at first :smile:, but it gets easier with time. [The algorithm](https://www.w3.org/TR/css-page-3/#variable-sizing) is close to the [current code](https://github.com/Kozea/WeasyPrint/blob/bb69424d811b39821697326ea0fe1d26fe540bc2/weasyprint/layout/pages.py#L182-L276), so it can be helpful to have both side by side to understand how it works.

I'll be happy to help, so feel free to ask anything.

galbwe:
@liZe Thank you for the context, and for the speedy reply! I'll start working through the specs and reach out if I have any questions.

liZe:
Closed by #1890.

PR #1890, opened 2023-05-28
Refactor compute_variable_dimension function to specifications

kygoh:
> In 2012, when @SimonSapin implemented the page-margins layout, the specification was this one:
https://www.w3.org/TR/2006/WD-css3-page-20061010/#margin-dimension
>
> [...]
>
> This proposition was accepted, probably with some changes (I don't remember, maybe Simon does). So the first step is to find what's different between the current specification and the current implementation in `compute_variable_dimension`, and then fix these differences.

_Originally posted by @liZe in https://github.com/Kozea/WeasyPrint/issues/58#issuecomment-511187829_

I think the proposition was accepted without changes(?) because all tests passed without changes even after refactoring the code to "more closely follow" the specifications layout terminologies and algorithm flow.

### DIFF
--- a/weasyprint/layout/page.py
+++ b/weasyprint/layout/page.py
@@ -27,7 +27,8 @@ class OrientedBox:
     @outer.setter
     def outer(self, new_outer_width):
         self.inner = min(
-            max(self.min_content_size, new_outer_width - self.sugar), self.max_content_size)
+            max(self.min_content_size, new_outer_width - self.sugar),
+            self.max_content_size)
 
     @property
     def outer_min_content_size(self):
@@ -210,9 +211,12 @@ def compute_variable_dimension(context, side_boxes, vertical, available_size):
             if available_size > (
                     box_a.outer_max_content_size +
                     box_c.outer_max_content_size):
-                # sum of the outer max-content widths is less than the available width
-                flex_space = (available_size -
-                    box_a.outer_max_content_size - box_c.outer_max_content_size)
+                # sum of the outer max-content widths
+                # is less than the available width
+                flex_space = (
+                    available_size -
+                    box_a.outer_max_content_size -
+                    box_c.outer_max_content_size)
                 flex_factor_a = box_a.outer_max_content_size
                 flex_factor_c = box_c.outer_max_content_size
                 flex_factor_sum = flex_factor_a + flex_factor_c
@@ -225,11 +229,12 @@ def compute_variable_dimension(context, side_boxes, vertical, available_size):
             elif available_size > (
                     box_a.outer_min_content_size +
                     box_c.outer_min_content_size):
-                # sum of the outer min-content widths is less than the available width
-                flex_space = (available_size -
-                    box_a.outer_min_content_size - box_c.outer_min_content_size)
-                box_a.inner = box_a.min_content_size
-                box_c.inner = box_c.min_content_size
+                # sum of the outer min-content widths
+                # is less than the available width
+                flex_space = (
+                    available_size -
+                    box_a.outer_min_content_size -
+                    box_c.outer_min_content_size)
                 flex_factor_a = (
                     box_a.max_content_size - box_a.min_content_size)
                 flex_factor_c = (
@@ -238,15 +243,15 @@ def compute_variable_dimension(context, side_boxes, vertical, available_size):
                 if flex_factor_sum == 0:
                     flex_factor_sum = 1
                 box_a.outer = box_a.min_content_size + (
-                   flex_space * flex_factor_a / flex_factor_sum)
+                    flex_space * flex_factor_a / flex_factor_sum)
                 box_c.outer = box_c.min_content_size + (
-                   flex_space * flex_factor_c / flex_factor_sum)
+                    flex_space * flex_factor_c / flex_factor_sum)
             else:
                 # otherwise
-                flex_space = (available_size -
-                    box_a.outer_min_content_size - box_c.outer_min_content_size)
-                box_a.inner = box_a.min_content_size
-                box_c.inner = box_c.min_content_size
+                flex_space = (
+                    available_size -
+                    box_a.outer_min_content_size -
+                    box_c.outer_min_content_size)
                 flex_factor_a = box_a.min_content_size
                 flex_factor_c = box_c.min_content_size
                 flex_factor_sum = flex_factor_a + flex_factor_c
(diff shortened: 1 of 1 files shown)


liZe:
Thanks! It seems to be safe to merge and to close #58.

One of the main differences between WeasyPrint and other tools (at least Prince) is that page margin boxes with defined widths are not resized to exactly cover the whole margin: they can overflow or leave empty spaces. I can’t find anything in the specification about this, did I miss something?

kygoh:
> page margin boxes with defined widths are not resized to exactly cover the whole margin: they can overflow or leave empty spaces. I can’t find anything in the specification about this, did I miss something?

If I understand the specification correctly, WeasyPrint is implemented according to the [specification](https://www.w3.org/TR/css-page-3/#margin-dimension):

> The rules for applying [min-height](https://www.w3.org/TR/CSS21/visudet.html#propdef-min-height), [max-height](https://www.w3.org/TR/CSS21/visudet.html#propdef-max-height), [min-width](https://www.w3.org/TR/CSS21/visudet.html#propdef-min-width), and [max-width](https://www.w3.org/TR/CSS21/visudet.html#propdef-max-width) [[CSS21]](https://www.w3.org/TR/css-page-3/#biblio-css21) do apply to page-margin boxes and _**may imply a recalculation of the [width](https://www.w3.org/TR/CSS21/visudet.html#propdef-width), [height](https://www.w3.org/TR/CSS21/visudet.html#propdef-height), and/or margins if the dimensions resulting from the specified width or height violate their constraint**_. If the UA does not support the [min-height](https://www.w3.org/TR/CSS21/visudet.html#propdef-min-height) or [min-width](https://www.w3.org/TR/CSS21/visudet.html#propdef-min-width) properties then it must behave as if [min-height](https://www.w3.org/TR/CSS21/visudet.html#propdef-min-height) and [min-width](https://www.w3.org/TR/CSS21/visudet.html#propdef-min-width) were always zero.

The statement in _**italic-bold**_ is what I was referring to and [`compute_fixed_dimension`](https://github.com/Kozea/WeasyPrint/blob/0a03e3d1cf557541a7937f8603821a868c5302fe/weasyprint/layout/page.py#L105-L173) appears to match the rules in [5.3.3. Page-Margin Box Fixed Dimension Computation Rules](https://www.w3.org/TR/css-page-3/#fixed-sizing).

I must admit I am not very familiar with the whole specification, so please point out if I've misunderstood it.

kygoh:
> What happens for example when `width: 400mm` is set to `@top-center` when the page width is 210mm?

Do you mean that even though:

> [compute_fixed_dimension](https://github.com/Kozea/WeasyPrint/blob/0a03e3d1cf557541a7937f8603821a868c5302fe/weasyprint/layout/page.py#L105-L173) appears to match the rules in [5.3.3. Page-Margin Box Fixed Dimension Computation Rules](https://www.w3.org/TR/css-page-3/#fixed-sizing)

it wasn't handled correctly or the specification is silent on how that should be handled?

liZe:
> Do you mean that even though:
> 
> > [compute_fixed_dimension](https://github.com/Kozea/WeasyPrint/blob/0a03e3d1cf557541a7937f8603821a868c5302fe/weasyprint/layout/page.py#L105-L173) appears to match the rules in [5.3.3. Page-Margin Box Fixed Dimension Computation Rules](https://www.w3.org/TR/css-page-3/#fixed-sizing)
> 
> it wasn't handled correctly or the specification is silent on how that should be handled?

This part sets the height of the `@top-center` box, but I’m talking of its width. I’m afraid there’s nothing in the specification about that, and I was wondering if you’ve seen something I haven’t.

kygoh:
> This part sets the height of the `@top-center` box, but I’m talking of its width. I’m afraid there’s nothing in the specification about that, and I was wondering if you’ve seen something I haven’t.

I think I understand what you mean. However, I still feel that WeasyPrint's implementation is better than the other tools by rendering the following two cases for the various tools in [PrintCSS Playround](https://printcss.live/):

```html
<style>
@page {
  size: 800px;
  margin: 100px;
  padding: 42px;
  border: 7px solid blue;
  @top-center { content: 'the quick brown fox jumped over the lazy dog then the quick brown fox jumped over the lazy dog then the quick brown fox jumped over the lazy dog then the quick brown fox jumped over the lazy dog'; 
    width: 400px;
  }
  @top-right { content: 'the quick brown fox jumped over the lazy dog then the quick brown fox jumped over the lazy dog then the quick brown fox jumped over the lazy dog then the quick brown fox jumped over the lazy dog';
    width: 400px;
  }
}
</style>
```
```html
<style>
@page {
  size: 800px;
  margin: 100px;
  padding: 42px;
  border: 7px solid blue;
  @top-center { content: 'the quick brown fox jumped over the lazy dog then the quick brown fox jumped over the lazy dog then the quick brown fox jumped over the lazy dog then the quick brown fox jumped over the lazy dog';
    width: 400px;
  }
  @top-right { content: 'the quick brown fox jumped over the lazy dog then the quick brown fox jumped over the lazy dog then the quick brown fox jumped over the lazy dog then the quick brown fox jumped over the lazy dog';
    width: auto;
  }
}
</style>
```

The two cases were rendered exactly the same on Prince: "beautiful" without overlapping and also "auto-cropping".

On the other hand, WeasyPrint rendered the two differently and also "uglily" with overlap as well as text running into page area.

However, I feel that the "beautiful" rendering is not "respecting" the intention of the designer😬. If the designer intends the boxes to overlap (although by mistake), then it should be rendered accordingly.

liZe:
> However, I feel that the "beautiful" rendering is not "respecting" the intention of the designer. If the designer intends the boxes to overlap (although by mistake), then it should be rendered accordingly.

I agree 😄. But the only rule we try to follow is the specification, not what the user finds useful 😄. If you too can’t find anything in the specification, then let’s do what we want (ie. keep what we have, in my opinion).

kygoh:
> If you too can’t find anything in the specification, then let’s do what we want (ie. keep what we have, in my opinion).

To complete our discussion on this issue, the note in [5.3.2.2. Resolving auto widths](https://www.w3.org/TR/css-page-3/#variable-auto-sizing) states:

> Note: The high-level goals are (in order of priority) to center the middle box (B) if it is generated, to minimize overflow and overlap, and to distribute space proportionally to the amount of content.

Though not stated, I think if these high-level goals are similarly applied to fixed dimension computation rules, WeasyPrint is staying true to the specification.

liZe:
> Though not stated, I think if these high-level goals are similarly applied to fixed dimension computation rules, WeasyPrint is staying true to the specification.

You’re right. "to minimize overflow and overlap" indicates that the boxes can actually overflow and overlap if the user wants, that’s cool!

PR #1895, opened 2023-06-07
Fix min-width ignored when computing cell measures

### LINKED ISSUE #1383: Table cell min-width is ignored when another cell has overflowing text
Two tables here to compare, note the `min-width` css rule:

```html
<head>
<style>
table td { border: 1px solid black; }
table.key-val tr td:nth-child(1) { min-width: 13em; }
</style>
</head>

<body>
<table class="key-val">
    <tbody>
        <tr>
            <td>Normal Key 1</td>
            <td>Normal Value 1</td>
        </tr>
        <tr>
            <td>Normal Key 2</td>
           <td>Normal Value 2</td>
        </tr>
    </tbody>
</table>
<br>
<table class="key-val">
    <tbody>
        <tr>
            <td>Short value</td>
            <td>Works as expected</td>
        </tr>
        <tr>
            <td>Long Value</td>
            <td>Annoyingly breaks my table layout: Sed ut perspiciatis unde omnis iste natus error sit voluptatem
                accusantium doloremque laudantium, totam rem aperiam,
                eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo.
            </td>
        </tr>
    </tbody>
</table>
</body>
```

Firefox:
image

Weasyprint:
image

I have experimented with different width units as well as tried adjusting `overflow-wrap` with no effect on Weasyprint's result.

kygoh:
Fix #1383 as specified under [min-content width of a column based on cells of span up to 1](https://www.w3.org/TR/css-tables-3/#min-content-width-of-a-column-based-on-cells-of-span-up-to-1) in [3.8.3. Computing Column Measures](https://www.w3.org/TR/css-tables-3/#computing-column-measures)

### DIFF
--- a/tests/layout/test_table.py
+++ b/tests/layout/test_table.py
@@ -2912,3 +2912,64 @@ def test_table_different_display():
         </td>
       </table>
     ''')
+
+
+@assert_no_logs
+def test_min_width_with_overflow():
+    # issue 1383
+    page, = render_pages('''
+<head>
+<style>
+table td { border: 1px solid black; }
+table.key-val tr td:nth-child(1) { min-width: 13em; }
+</style>
+</head>
+
+<body>
+<table class="key-val">
+    <tbody>
+        <tr>
+            <td>Normal Key 1</td>
+            <td>Normal Value 1</td>
+        </tr>
+        <tr>
+            <td>Normal Key 2</td>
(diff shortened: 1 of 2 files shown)


liZe:
Thank you!

PR #1899, opened 2023-06-20
Fix named pages inheritence

### LINKED ISSUE #1897: Problem with named pages and out of normal page flow elements
## Problem

When rendering following HTML with weasyprint, the CSS for `named-page` is not applied on the first page. The first page is blue (from the default `@page` rule), but it contains content of the named page.

```html
<html>
    <head>
        <style>
            #header { position: running(header); }
            #watermark { position: fixed; top: 10cm; left: 5cm; font-size: 30pt; }
            
            @page {
                size: A4;
                background-color: blue;
                @top-center { content: element(header); }
            }
            
            #named-page { page: named-page; }
            @page named-page { background-color: red; }
            
            .pagebreak { break-after: page; }
        </style>
    </head>
    <body>
        <div id="header">HEADER</div>
        <div id="watermark">WATERMARK</div>
        <div id="named-page">
            NAMED PAGE 1
            <div class="pagebreak"></div>
            NAMED PAGE 2
        </div>
    </body>
</html>
```

Screenshot_20230620_090329

You can get the expected result (both pages in red) by moving the HTML elements for the header and watermark below the named page. However, I think this is unexpected behavior.

## Suggestion

I think the problem is in calculating the start page value of the first page. Currently the start page value of the first element (`#header`) is used which results in an unnamed page. I think weasyprint should exclude elements that are out of the normal page flow from the start/end page value calculation (such as `position: running()` and `position: fixed`).

This should be compliant with the CSS specification for `page` (https://drafts.csswg.org/css-page-3/#propdef-page):

> 2. Next, a start [page](https://drafts.csswg.org/css-page-3/#propdef-page) value and end [page](https://drafts.csswg.org/css-page-3/#propdef-page) value is determined for each box as the value (if any) propagated from its first or last child box (respectively), else the used value on the box itself. A child propagates its own [start](https://drafts.csswg.org/css-page-3/#start-page-value) or [end page value](https://drafts.csswg.org/css-page-3/#end-page-value) if and only if the [page](https://drafts.csswg.org/css-page-3/#propdef-page) property applies to it.
>
> > NOTE: A first or last child _box_ is not always generated by a first or last child _element_. For example, an element could only have a previous sibling with 'display: none' which does not generate any box.

What is your opinion on that?

MWedl:
Fix #1897 

Hi!
I am not sure if this PR is sufficient to reliably fix the problem with named page inheritance, since I have not found how/where weasyprint calculates class A break points. However, these changes fix the page inheritance in my situation (and do not break other test cases).

Feel free add commits or point me in the right direction :)

### DIFF
--- a/tests/layout/test_page.py
+++ b/tests/layout/test_page.py
@@ -831,16 +831,26 @@ def test_page_names_10():
       <div id="running">running</div>
       <div id="fixed">fixed</div>
       <section>
-        text
+        <h1>text</h1>
         <div class="pagebreak"></div>
-        text
+        <article>text</article>
       </section>
     ''')
     page1, page2 = pages
 
     assert (page1.width, page1.height) == (100, 100)
+    html, runing = page1.children
+    body, = html.children
+    fixed, section, = body.children
+    h1, pagebreak = section.children
+    assert h1.element_tag == 'h1'
 
     assert (page2.width, page2.height) == (100, 100)
+    html, running = page2.children
+    fixed, body = html.children
+    section, = body.children
+    article, = section.children
+    assert article.element_tag == 'article'
 
 
 @assert_no_logs
--- a/tests/layout/test_page.py
+++ b/tests/layout/test_page.py
@@ -818,6 +818,32 @@ def test_page_names_9():
 
 
 @assert_no_logs
+def test_page_names_10():
+    pages = render_pages('''
+      <style>
+        #running { position: running(running); }
+        #fixed { position: fixed; }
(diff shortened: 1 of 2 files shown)


liZe:
> Hi!
> I am not sure if this PR is sufficient to reliably fix the problem with named page inheritance, since I have not found how/where weasyprint calculates class A break points. However, these changes fix the page inheritance in my situation (and do not break other test cases).

Thanks a lot!

I’ll add a comment about this in the code, as the way we find possible page breaks doesn’t follow the specification (yet).

PR #1908, opened 2023-07-11
Fix IndexError when rendering PDF version 1.4

### LINKED ISSUE #1907: IndexError when bulding fonts dictionary for PDF version 1.4
I'm using weasyprint version 59. I found the bug when trying to render a PDF with the PDF version 1.4. The problem is in the following code https://github.com/Kozea/WeasyPrint/blob/master/weasyprint/pdf/fonts.py#L118-L123
```python
if pdf.version <= b'1.4':
    cids = sorted(font.widths)
    padded_width = int(ceil(cids[-1] / 8))
    bits = ['0'] * padded_width * 8
    for cid in cids:
        bits[cid] = '1'
```
In my case `cids` contains `[3, 49, 58, 68, 72, 73, 74, 76, 79, 81, 82, 86, 88]`. So the `cids[-1]` is `88` which can be divided by 8 without a reminder. This causes the problems in the code after that. `padded_width` value will be `11`, `bits` length is 88, and in the loop there will be an `IndexError` when trying to set the last bit when `cid = 88`.

catcombo:
Fixes #1907

### DIFF
--- a/weasyprint/pdf/fonts.py
+++ b/weasyprint/pdf/fonts.py
@@ -117,7 +117,7 @@ def build_fonts_dictionary(pdf, fonts, compress_pdf, subset, hinting):
             })
             if pdf.version <= b'1.4':
                 cids = sorted(font.widths)
-                padded_width = int(ceil(cids[-1] / 8))
+                padded_width = int(ceil((cids[-1] + 1) / 8))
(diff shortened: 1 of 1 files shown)


chadcasey:
This still fails for me using the brew-installed version, v63.0

Issue #1871, opened 2023-04-29
Footnote is appearing on page before its call


sahilrl:
Hello @liZe! 

Please note the footnote number 14 which is being rendered on page 2 while it's footnote call site is on the next page.

![weasyprint_bug_footnotes](https://user-images.githubusercontent.com/55531359/235327850-d72b8002-fd16-4d21-b7e5-a4d78c8007a1.png)

PDF
[output.pdf](https://github.com/Kozea/WeasyPrint/files/11360215/output.pdf)

HTML
```html
<!DOCTYPE html>
<html lang="en">

<head>
    <meta charset="UTF-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <link rel="preconnect" href="https://fonts.googleapis.com">
    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@300&display=swap" rel="stylesheet">
    <link href="https://fonts.googleapis.com/css2?family=Poppins:wght@200&display=swap" rel="stylesheet">
    <link rel="preconnect" href="https://fonts.googleapis.com">
    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
    <link href="https://fonts.googleapis.com/css2?family=IBM+Plex+Sans:wght@300;400;500;600;700&family=Jura:wght@700&display=swap"
          rel="stylesheet">
    <title>Document</title>
</head>

<body>
<header>
    
</header>

<footer>

</footer>

<table class="table-main">
    <p>Lorem ipsum dolor, sit amet consectetur adipisicing elit. Enim quos, alias aspernatur dolore quisquam fuga tempora consequuntur! Nesciunt, maiores maxime, laudantium magnam distinctio error, libero fugiat nobis nihil aperiam enim?</p>
<p>Lorem ipsum dolor, sit amet consectetur adipisicing elit. Enim quos, alias aspernatur dolore quisquam fuga tempora consequuntur! Nesciunt, maiores maxime, laudantium magnam distinctio error, libero fugiat nobis nihil aperiam enim?</p>
[...]

PR #1915, opened 2023-07-16
Bug Fix: Footnote is appearing on page before its call #1871

sahilrl:
Fix:  #1871


liZe:
Thanks a lot! I’ll clean a few things, add a test, and then merge this PR.

PR #1916, opened 2023-07-17
fix:prioritize GTK3 fontconfig dll over graphviz's

YasserMohsen:
graphviz on Windows has `fontconfig` DLL file, so when it is installed, it takes the needed DLL file from it instead of getting it from GTK3 installed package. So, I have just sorted the list of expected DLL filenames to prioritize the current GTK3 fontconfig DLL file which is `libfontconfig-1.dll`. Now, no need to uninstall Graphviz from Windows to make it work.

### DIFF
--- a/weasyprint/text/ffi.py
+++ b/weasyprint/text/ffi.py
@@ -436,8 +436,8 @@ harfbuzz = _dlopen(
     'libharfbuzz.so.0', 'libharfbuzz.so.0', 'libharfbuzz.0.dylib',
     'libharfbuzz-0.dll')
 fontconfig = _dlopen(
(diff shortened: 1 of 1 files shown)


liZe:
Hi!

I’m sorry, but fixing the order for your use case can also break some other users’ cases that rely on the current order. There’s no clean solution that will work for everybody :/, that’s just how DLL hell works…

In case you have multiple versions of the same library on your system, the best solution is probably not to include the programs’ path in the global list of folders where to look for libraries, and define this list for each application, just as WeasyPrint now does.

Another possible solution: would you be interesting in testing an all-in-one executable?

PR #1922, opened 2023-07-24
Add support for textLength and lengthAdjust in SVG text elements

oshmoun:
Add support for `textLength` and `lengthAdjust`by calculating a fitting letter spacing, and a fitting X scaling factor if `lengthAdjust=spacingAndGlyphs`, that make the text expand/shrink to fit the specified size.

There is one weird behaviour that browsers do, at least Chromium, that I noticed while implementing this:
If letter-spacing is specified alongside textLength, then it functions as a negative width offset, a TODO was added for that.

### DIFF
--- a/tests/draw/svg/test_text.py
+++ b/tests/draw/svg/test_text.py
@@ -301,3 +301,73 @@ def test_text_rotate(assert_pixels):
           rotate="180" letter-spacing="2">abc</text>
       </svg>
     ''')
+
+
+@assert_no_logs
+def test_text_text_length(assert_pixels):
+    assert_pixels('''
+        __RRRRRR____________
+        __RRRRRR____________
+        __BB__BB__BB________
+        __BB__BB__BB________
+    ''', '''
+      <style>
+        @font-face { src: url(weasyprint.otf); font-family: weasyprint }
+        @page { size: 20px 4px }
+        svg { display: block }
+      </style>
+      <svg width="20px" height="4px" xmlns="http://www.w3.org/2000/svg">
+        <text x="2" y="1.5" font-family="weasyprint" font-size="2" fill="red">
+          abc
+        </text>
+        <text x="2" y="3.5" font-family="weasyprint" font-size="2" fill="blue"
+          textLength="10">abc</text>
+      </svg>
+    ''')
+
+
+@assert_no_logs
+def test_text_length_adjust_glyphs_only(assert_pixels):
+    assert_pixels('''
+        __RRRRRR____________
+        __RRRRRR____________
+        __BBBBBBBBBBBB______
+        __BBBBBBBBBBBB______
+    ''', '''
+      <style>
+        @font-face { src: url(weasyprint.otf); font-family: weasyprint }
+        @page { size: 20px 4px }
+        svg { display: block }
+      </style>
+      <svg width="20px" height="4px" xmlns="http://www.w3.org/2000/svg">
+        <text x="2" y="1.5" font-family="weasyprint" font-size="2" fill="red">
+          abc
+        </text>
+        <text x="2" y="3.5" font-family="weasyprint" font-size="2" fill="blue"
+          textLength="12" lengthAdjust="spacingAndGlyphs">abc</text>
+      </svg>
+    ''')
+
+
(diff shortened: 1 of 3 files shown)


liZe:
I’ve finished the small cleanup, here are some comments:

- the 3 first commits are just for styling purpose
- d001a23 changes the signature of `draw_first_line` to allow a matrix instead of separate properties
- cd5608c tries to simplify many small details, and lead to these questions:

1. Why did you introduce the `- 0.5` value in spaces count? I didn’t find an important change with `- 1` and it looks more "normal" :smile:.
2. I’ve always overridden the letter-spacing value when "spacing" is set. The rendering is a bit different to what browsers do: they have some extra space on the right, but there’s nothing I can find in the specification about that ("Whether letter-spacing is applied at the end of a line is undefined in this level.") and what WeasyPrint does looks better in my opinion.
3. I prefer the rendering of Firefox for rotated + scaled letters, but WeasyPrint looks more like Chrome. What do you think? It should be possible to fix that by playing with transformation matrices.

Here’s a small sample I’ve used to test manually:

```svg
<svg width="150" height="1000" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" font-family="serif">
  <rect fill="pink" width="150" height="1000"/>
  <rect fill="white" width="100" height="1000" x="10"/>
  <text textLength="100" x="10" y="100" fill="black">abcd</text>
  <text textLength="100" x="10" y="200" fill="black" rotate="30">abcd</text>
  <text textLength="100" x="10" y="300" fill="black" lengthAdjust="spacingAndGlyphs">abcd</text>
  <text textLength="100" x="10" y="400" fill="black" rotate="30" lengthAdjust="spacingAndGlyphs">abcd</text>
  <text x="10" y="500" fill="black" letter-spacing="10">abcd</text>
  <text textLength="100" x="10" y="600" fill="black" letter-spacing="10">abcd</text>
  <text textLength="100" x="10" y="700" fill="black" letter-spacing="10" lengthAdjust="spacingAndGlyphs">abcd</text>
  <text textLength="100" x="10" y="800" fill="black" letter-spacing="10" lengthAdjust="spacingAndGlyphs" rotate="30">abcd</text>
</svg>
```

oshmoun:
Many thanks for the adjustments!

>     1. Why did you introduce the `- 0.5` value in spaces count? I didn’t find an important change with `- 1` and it looks more "normal" 😄.
> 

During my testing I noticed that without that, the letter spacing would be too much and move the last glyph outside of the specified text length. Hence I assumed there is an additional `0.5` gap that should be considered in the calculation. Now upon testing your code, I'm not so sure it wasn't a mea culpa during testing. Change to `-1` appears to work as expected here :+1: 
 
>     2. I’ve always overridden the letter-spacing value when "spacing" is set. The rendering is a bit different to what browsers do: they have some extra space on the right, but there’s nothing I can find in the specification about that ("Whether letter-spacing is applied at the end of a line is undefined in this level.") and what WeasyPrint does looks better in my opinion.
> 

Yes, I also can't find much in the specs, so that browser behaviour is kinda weird :sweat_smile:
I see no issue with your suggestion, overwriting `letter-spacing` makes sense, though maybe a warning could be logged to let the user know, will keep a note of that.

>     3. I prefer the rendering of Firefox for rotated + scaled letters, but WeasyPrint looks more like Chrome. What do you think? It should be possible to fix that by playing with transformation matrices.
> 

I don't have any personal preference in that regard :sweat_smile: 
But I think it'd be more time efficient to keep it like Chrome, as that's the one with least effort, YAGNI principle and all that :smile:

oshmoun:
Sorry about the delay in implementing the tests, will try to do that over the weekend 🙂

liZe:
> Sorry about the delay in implementing the tests, will try to do that over the weekend 🙂

Thanks. No need to hurry! 😄

oshmoun:
three tests added, one for textLength with default lengthAdjust (spacing)
one for `lengthAdjust=spacingAndGlyphs` but glyphs only (no letter-spacing specified)
and one for both spacing and glyphs

liZe:
Thanks a lot!

PR #1934, opened 2023-08-06
Fix balance columns before "column-span: all"

### LINKED ISSUE #1914: Balance columns before "column-span: all"
Is the [column-span](https://www.w3.org/TR/css-multicol-1/#column-span) attribute correctly handled? In my experimentation it seems just ignored, but maybe I am missing something. I used it as a property of an img tag.

sahilrl:
Fix #1914.

### DIFF
--- a/tests/layout/test_column.py
+++ b/tests/layout/test_column.py
@@ -418,6 +418,38 @@ def test_column_span_9():
 
 
 @assert_no_logs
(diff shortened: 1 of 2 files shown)


liZe:
Thanks a lot! I’ll add a test and clean a couple of things.

Issue #1904, opened 2023-07-08
Large child of `overflow: hidden` parent causes an unexpected page break.


BadCoder1337:
I tried to create some "pyramid" document ([desired layout](https://github.com/Kozea/WeasyPrint/assets/16510197/c76adab0-10e4-4793-8e3b-c824559bd1d4), [source](https://gist.github.com/BadCoder1337/b0e5b44070a46b92520652dbfc0c5a0f), [demo](https://badcoder1337.github.io/print/index.html)) with different paper formats but got some weird PDF results in [Chrome](https://github.com/Kozea/WeasyPrint/files/11990615/chrome.pdf) and [WeasyPrint](https://github.com/Kozea/WeasyPrint/files/11990614/weasyprint.pdf).

With @liZe help we found a minimal example showing the issue
```html
<div style="overflow: hidden; height: 10cm; border: 1px solid red">
  <div>abc</div>
  <div style="height: 100cm">def</div>
</div>
```
and the workaround 
```css
img {
  display: block;
  margin-top: -100cm;
  transform: translate(0, 100cm);
}
```
Version Info:
Chrome 114.0.5735.199
WeasyPrint 59.0

liZe:
Thanks for the bug report.

Here’s the related paragraph in the [specification](https://www.w3.org/TR/css-break-3/#possible-breaks):

>In addition to any content which is not generally fragmentable, UAs may consider as [monolithic](https://www.w3.org/TR/css-break-3/#monolithic) any elements with [overflow](https://www.w3.org/TR/css-overflow-3/#propdef-overflow) set to [auto](https://www.w3.org/TR/css-overflow-3/#valdef-overflow-auto) or [scroll](https://www.w3.org/TR/css-overflow-3/#valdef-overflow-scroll) and any elements with [overflow: hidden](https://www.w3.org/TR/css-overflow-3/#propdef-overflow) and a non-[auto](https://www.w3.org/TR/css-sizing-3/#valdef-width-auto) [logical height](https://www.w3.org/TR/css3-writing-modes/#block-size) (and no specified maximum logical height).

With "normal" words: we should not break boxes with defined height and hidden overflow.

PR #1936, opened 2023-08-08
Fix large child of overflow hidden parent causes an unexpected page break

sahilrl:
FIX: #1904


liZe:
Thanks a lot and sorry for the delay, we’re quite busy right now.

PR #1952, opened 2023-08-31
 Fixed two span properties in one place

ssjkamei:
Hello,

There was an error that the span existed in two places in the Actions Workflow (ubuntu-latest - 3.11).
As far as I can see, it was a property that returned the set default value, so I deleted it.

Please check if there are any problems.


grewn0uille:
Hello,

Thanks, indeed, but I think this change is already on master https://github.com/Kozea/WeasyPrint/commit/e1f4ef8fd3810b0fe0c5b885db62bc6f4e2a9ba5 :)

ssjkamei:
@grewn0uille Thank you. I was able to confirm it.
I guess I did not understand how github works. I will take it down.

PR #1954, opened 2023-08-31
pdf: add default WeasyPrint Creator metadata

pmjdebruijn:
It's a trivial change :)


liZe:
Hi!

Thanks for the pull request. But… I’m sorry, "Creator" doesn’t mean what it seems to mean. According to the PDF specification:

> Creator: If the document was converted to PDF from another format, the name of the conforming product that created the original document from which it was converted.

So, "Creator" is already set using the [HTML meta "generator" tag](https://html.spec.whatwg.org/multipage/semantics.html#meta-generator) 😄.

pmjdebruijn:
I know, the problem is, if that tag isn't set, you end up with an empty Creator, which looks very odd.
My patch, just prevents the Creator tag from ever being empty.

Issue #1958, opened 2023-09-05
Handle y offset of glyphs


gauravsamudra:
Hello! I am creating a pdf with weasyprint's html to pdf feature. I've gotten everything to work fine except for one specific set of combination of devanagari characters - in this case a glyph made up of conjunct consonant with vowel diacritics - e.g. ङ्गु - which is made up of ङ, ग and ु. It looks like ![image](https://github.com/Kozea/WeasyPrint/assets/60246931/6a253aac-dce2-4684-9fe7-fbb7a44c6449) in `ITF Devanagari` font (in vscode or chrome or firefox browsers). The glyph is ङ, underneath that a partial ग and underneath that ु

* in the pdf generated using weasyprint, the ग and ु get overlapped and the character looks like ![image](https://github.com/Kozea/WeasyPrint/assets/60246931/fd933d48-f704-4b75-8901-4505d8908f8b) - notice the overlap between partial ग and ु
* but in a pdf generated (print to pdf) by chrome or firefox browser, it works fine - ![image](https://github.com/Kozea/WeasyPrint/assets/60246931/6a253aac-dce2-4684-9fe7-fbb7a44c6449)

I am a loss of what should I be investigating. I’ve tried playing with css’ font feature settings but that doesn’t seem to have any effect. I'll appreciate any help to point me in the right direction or forum!

liZe:
Hi!

That’s probably a bug in Harfbuzz. I don’t have the ITF Devanagari font, but I also have the bug in my browser, based on the same shaping libraries as WeasyPrint, [on this page](https://www.indiantypefoundry.com/fonts/itf-devanagari/layouts) (but not in Chrome and Firefox). 

Could you please share the font by mail (for testing purpose, of course)?

gauravsamudra:
Thanks @liZe !

I am using ITF Devanagari font that is available by default on mac. I'm not sure if I can/should export and share it.
Here is a simplified version of my css and html using which I can reproduce the issue - 

``` css
html {
  font-family: 'ITF Devanagari';
  font-size: 22pt;
}
```

``` html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <title>Test</title>
  </head>

  <body>
    <p>ङ्गु</p>
  </body>
</html>
```
and the python code I use
```py
    html_01 = HTML(filename='./app/template/01-template.html')
    css_01 = CSS(filename='./app/template/01-style.css')
[...]

PR #1970, opened 2023-09-25
Handle y offset of glyphs

gauravsamudra:
Initial attempt to [Handle y offset of glyphs](https://github.com/Kozea/WeasyPrint/issues/1958)

(This is a work in progress, I've opened the PR for to get some feedback. The calculation I have - `offset_y = glyph_info.geometry.y_offset / font_size / textbox.height` does not seem to be accurate - since if the line height is large, the glyphs show up lower then they should.)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ requires-python = '>=3.7'
 readme = {file = 'README.rst', content-type = 'text/x-rst'}
 license = {file = 'LICENSE'}
 dependencies = [
-  'pydyf >=0.6.0',
+  'pydyf >=0.8.0',
   'cffi >=0.6',
   'html5lib >=1.1',
   'tinycss2 >=1.0.0',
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -12,7 +12,7 @@ requires-python = '>=3.7'
 readme = {file = 'README.rst', content-type = 'text/x-rst'}
 license = {file = 'LICENSE'}
 dependencies = [
-  'pydyf >=0.6.0',
```
gauravsamudra, line 15: I assume that once we have a new version for pypdf, it will be 0.8.0

liZe:
Thanks a lot!

> The calculation I have - `offset_y = glyph_info.geometry.y_offset / font_size / textbox.height` does not seem to be accurate - since if the line height is large, the glyphs show up lower then they should.

That was `glyph_info.geometry.y_offset / 1000` (1000 being a constant defined by Pango.)

Issue #1909, opened 2023-07-13
Add a --timeout option to the CLI API


quique:
I'm thinking of something like
`weaseyprint --timeout 30 https://weasyprint.org /tmp/weasyprint-website.pdf`

The provided value would just have to be passed to the `default_url_fetcher(url, timeout=10, ssl_context=None)` method.

liZe:
That’s a good idea. It should be possible to pass the default URL fetcher with timeout set to given value [in `__main__.py`](https://github.com/Kozea/WeasyPrint/blob/master/weasyprint/__main__.py#L204). Anyone interested to add this feature?

MalanB:
Hey, interested in picking this issue

liZe:
> Hey, interested in picking this issue

Cool! Don’t hesitate to ask if you want some hints!

staticvoidzach:
Wondering if there should be multiple types of timeout options? When looking at `weasyprint --timeout 30 in.com out.pdf` my first though is the timeout applies to the entire process, so the command will exit prematurely if the fetch + resource gather + render process takes more than 30 seconds.

Although, I'd also like to have the ability to control the timeouts of individual stages, e.g. `--timeout-for-initial-page-fetch`  and `--timeout-for-each-individual-asset-fetch-and-load` and `--timeout-for-pdf-render`, perhaps with shorter names however...

I'd also consider what the error behaviour would be? Should that be configurable? Potentially someone doesn't want their entire pdf generation to die if 1 out of 10 images takes longer than 10 seconds to fetch. Seems like an undesirable situation, yes, I wouldn't my company logo to be missing on 1 random report out of 100, however there's always a situation for these things.

Perhaps if timeout control is introduced for url fetching, potentially control for retries also needs to be considered?

Not sure about the best way to implement this though, as adding 8 new obscure arguments may be overkill and clutter the help menu. On the other hand, all of my favourite CLI tools completely fill my screen when I run `--help`, so possibly not the worst idea?

liZe:
I think that WeasyPrint can have a `timeout` option that will be used for each request, because it’s both useful (at least to avoid waiting forever when there’s a network problem) and easy to implement (just a parameter to give to the standard library.)

An option to control the overall time is useful to, but it’s much harder to implement. Moreover, other tools already exist to handle this, like [`timeout`](https://manpages.org/timeout).

[...]

PR #1971, opened 2023-09-25
Add timeout cli option

azhar316:
Fixes [issue #1909](https://github.com/Kozea/WeasyPrint/issues/1909) by using partial function to pass on the given timeout value.

### DIFF
--- a/weasyprint/__init__.py
+++ b/weasyprint/__init__.py
@@ -58,6 +58,8 @@ VERSION = __version__ = '59.0'
 #: :param cache:
 #:     A dictionary used to cache images in memory, or a folder path where
 #:     images are temporarily stored.
+#: :param int timeout:
+#:     Timeout in seconds for HTTP requests before they are dropped.
 DEFAULT_OPTIONS = {
     'stylesheets': None,
     'media_type': 'print',
@@ -75,6 +77,7 @@ DEFAULT_OPTIONS = {
     'full_fonts': False,
     'hinting': False,
     'cache': None,
+    'timeout': None
 }
 
 __all__ = [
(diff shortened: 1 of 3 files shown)


weasyprint/__init__.py
```diff
@@ -75,6 +77,7 @@
     'full_fonts': False,
     'hinting': False,
     'cache': None,
+    'timeout': None
```
liZe, line 80: Is that required? I think that `timeout` should just be a CLI option, but not something available in the Python API.

weasyprint/__main__.py
```diff
@@ -179,6 +185,10 @@ def main(argv=None, stdout=None, stdin=None, HTML=HTML):
                 optimize_size.add(arg)
     del args.optimize_size
 
+    url_fetcher = default_url_fetcher
+    if args.timeout:
```
liZe, line 189: Adding `is not None` would allow users to set 0 if they want (I don’t know if it’s useful, but I have the intuition that some users will find this useful for some reason!)

liZe (changes requested):
Thanks! Just a couple of things to check before merging.

weasyprint/__init__.py
azhar316: I added it as it is also used by CLI parser for default options but I guess it doesn't matter as anyway the value is `None`. Will remove it.

weasyprint/__main__.py
azhar316: Makes sense. Will fix it.

azhar316:
Have completed the changes.

liZe (approved):

azhar316:
@liZe Have fixed the import order!

Issue #1983, opened 2023-10-11
RecursionError: maximum recursion depth exceeded when converting a big Document


vince-vibin:
I have a very big HTML Document (~1000 Pages) which i want to convert to a PDF. But i am getting this error
`RecursionError: maximum recursion depth exceeded` 
my guess is that the recursion depth needs to be increased like [here](https://stackoverflow.com/questions/3323001/what-is-the-maximum-recursion-depth-and-how-to-increase-it).
For the call I am using this command 
`weasyprint input.html output.pdf`

grewn0uille:
Hello!

Can you please share the related HTML/CSS files?
So we can see if it’s possible to do something about it without changing the recursion depth.

vince-vibin:
Hi,
I can't upload the file because its sensitive but i've tried @lukas-holzner's fix and it worked perfectly fine. 
The file is an auto generated HTML file with ~80000 lines.

grewn0uille:
You can send a mail if you can’t share some information publicly

liZe:
Hi!

Could you please share by mail your document? It looks like you have a depth of ~400 nested tags in your HTML, is that possible?

liZe:
Hi @lukas-holzner,

Is this problem solved for you?

vince-vibin:
We where able to fix the issue with @lukas-holzner's fix. 
I don't know if you want to merge this in the state its in but we really cant share the document.

liZe:
> We where able to fix the issue with @lukas-holzner's fix.
> I don't know if you want to merge this in the state its in but we really cant share the document.

Just to be sure: does your HTML include ~400 levels of depth? If it does, we can consider that it’s an extreme corner case that can be solved by setting the recursion limit it a Python script calling WeasyPrint’s API.

lukas-holzner:
It might have 400 levels of depth. @vince-vibin  could you please check that?

vince-vibin:
yes it has ~380

liZe:
> yes it has ~380

Thanks for checking!

Then we can consider this as a corner case, and the easiest way to solve it is to use WeasyPrint with a Python script that sets the recursion limit.

PR #1984, opened 2023-10-11
Fix Recursionlimit Error by adding arg

lukas-holzner:
Fixing Issue #1983 by adding an argument to change the recursionlimit in python


andrinziegler:
I encountered the same issue and would have greatly appreciated the convenience of the new, additional argument. This is particularly true since I use a Java wrapper for the WeasyPrint CLI, rather than working directly with Python.

liZe:
@andrinziegler Do you have a HTML file with ~300 levels of nested tags too?

andrinziegler:
No, the nesting is 14 levels at its deepest. But the Document in itself is pretty huge.

Issue #1789, opened 2023-01-08
Is initial zoom setting allowed?


macdeport:
I have just discovered WeasyPrint, then used it and already appreciate all its interest: thanks to Lucie @grewn0uille, to you and Kozea for giving us access to it.

Do the rich features already available allow the following settings, especially the initial zoom?

![weasyprint-57 2--document-properties](https://user-images.githubusercontent.com/26068595/211202917-951b72e8-3e07-4a3b-8f4f-bf1cd91ee757.jpg)

liZe:
> I have just discovered WeasyPrint, then used it and already appreciate all its interest: thanks to Lucie @grewn0uille, to you and Kozea for giving us access to it.

😊 

> Do the rich features already available allow the following settings, especially the initial zoom?

It’s not supported, but it’s probably possible. The question is: can this information be specified using HTML or CSS? If there’s a specification for that, then the feature could be implemented in WeasyPrint. If it’s not, then it’s probably better to do this during a post-processing step, with a generic PDF edition library.

macdeport:
Thank you for the openness of your response and the thoughts it suggests.

> Do the rich features already available allow the following settings, especially the initial zoom?

Maybe I can propose an answer to this question: the philosophy of HTML is not to provide global formatting specifications like the zoom factor but on the other hand CSS has directives (I'm not sure of the exact term) like `@page` for the dimensions of the pages produced: the initial zoom value has some relationship?

Then this parameter could then be passed to the PDF rendering engine `pydyf`.

An alternative way could be a new `write_pdf()` parameter.

macdeport:
Lately found in documentation:

`write_pdf(target=None, stylesheets=None, zoom=1, attachments=None, finisher=None, presentational_hints=False, optimize_size=('fonts',), font_config=None, counter_style=None, image_cache=None, identifier=None, variant=None, version=None, forms=False, custom_metadata=False)`

Will give a try...

pbregener:
I use below to achieve what you probably are looking for. Check out the available options as mentioned in the inline comment.
```python
def set_viewer_prefs(_: weasyprint.Document, pdf: pydyf.PDF):
    # For options, see Table 28 in below file
    # https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/PDF32000_2008.pdf
    pdf.catalog['PageLayout'] = '/SinglePage'
    pdf.catalog['PageMode'] = '/UseNone'
```
[...]

PR #1985, opened 2023-10-11
Add PageMode and PageLayout to PDF catalog

pmjdebruijn:
See Table 28 (page 73-75) in
https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/PDF32000_2008.pdf

Thanks to #1789 for the inspiration

### DIFF
--- a/weasyprint/__init__.py
+++ b/weasyprint/__init__.py
@@ -66,6 +66,8 @@ DEFAULT_OPTIONS = {
     'pdf_variant': None,
     'pdf_version': None,
     'pdf_forms': None,
+    'pdf_pagemode': '',
+    'pdf_pagelayout': '',
     'uncompressed_pdf': False,
     'custom_metadata': False,
     'presentational_hints': False,
--- a/weasyprint/__main__.py
+++ b/weasyprint/__main__.py
@@ -92,6 +92,12 @@ PARSER.add_argument('--pdf-version', help='PDF version number')
 PARSER.add_argument(
     '--pdf-forms', action='store_true', help='include PDF forms')
 PARSER.add_argument(
+    '--pdf-pagemode', help='PDF PageMode',
+    choices=('UseOutlines', 'UseThumbs', 'UseAttachments', 'FullScreen'))
+PARSER.add_argument(
+    '--pdf-pagelayout', help='PDF PageLayout',
+    choices=('SinglePage', 'OneColumn', 'TwoColumnLeft', 'TwoColumnRight', 'TwoPageLeft', 'TwoPageRight'))
+PARSER.add_argument(
     '--uncompressed-pdf', action='store_true',
     help='do not compress PDF content, mainly for debugging purpose')
 PARSER.add_argument(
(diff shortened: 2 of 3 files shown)


pmjdebruijn:
Scribus outputs

```
/PageLayout /TwoColumnRight
/PageMode /UseOutlines

```

I also noticed that least Atril (and therefore probably Evince) does not honor WeasyPrint's output...

pmjdebruijn:
I just realized pydyf.String() caused the difference, so this should be good now.

liZe:
Hi!

There are some problems with this kind of features, as explained in #1789:
- there’s nothing in CSS or HTML to set this property,
- nothing prevents users from doing this in a post-processor stop or a finisher function,
- adding this feature would lead to accept an endless list of PDF features that will be hard to document and maintain.

This is a real life problem :smile:. Even if you took time and effort to propose this pull request, there are some problems already:
- code style is broken in CI,
- the public API is too close to the PDF names (both for words and for style) and not close to the vocabulary actually used by PDF readers,
- some possible values listed in the specification are not listed in the PR,
- there’s no test,
- there’s no documentation…

It doesn’t mean at all that your pull request is bad, it just means that adding maintainable and solid code is really hard :smile:.

Giving a general solution, as proposed in #1789, is a great way to solve this problem: users can do more, and are less limited to WeasyPrint’s API. For example, there’s no way to add an `OpenAction` field with your PR. We can bet that some people would propose a crazy command-line option allowing plain PDF dictionaries to "solve" that!

So, as explained in #1789, we’ll continue to propose post-processing steps instead of adding many other features.

We really do appreciate your effort to propose pull requests, and we hope that we’ll merge more of them. There are many open issues for bugs and feature requests, let’s please focus on them!

pmjdebruijn:
The problem is that setting very basic metadata shouldn't require a custom script.

If you give me some guidance how to translate the PDF names into something in Weasyprint, I could possibly rework the PR...

liZe:
> The problem is that setting very basic metadata shouldn't require a custom script.

I can understand your frustration, and I don’t like to refuse pull requests. But we clearly explained in #1789 that we weren’t interested in this feature, and our arguments are still valid.

If that’s very basic metadata, then it should be specified by HTML or CSS, just as document title, page size, outlines, creation date, creator tool, language, etc. But it’s not. Users always tend to think that what they want is basic, and the problem is that handling millions of basic problems is not a basic problem 😄.

> If you give me some guidance how to translate the PDF names into something in Weasyprint, I could possibly rework the PR...

As explained above, the problem is not the quality of your pull request.

The problem is that we don’t want to maintain code that adds a feature that can already be done using a custom script. Otherwise, we’ll end up adding tons of options that’ll take endless time to create, maintain, review, fix, test… We prefer to focus our limited time on problems that can’t be solved otherwise, and there are many open issues for that.

And if you don’t want to write a custom script, you can always use Ghostscript to tweak your PDF metadata, using just the CLI. Ghostscript is a really good example of how a tool can propose thousands of options, and the problems it causes with documentation :smile:.

PR #1987, opened 2023-10-15
Replace md1/sha1 with blake2b where possible

pmjdebruijn:
blake2b is faster and more secure than either, and has been generally available since python 3.6

### DIFF
--- a/weasyprint/document.py
+++ b/weasyprint/document.py
@@ -2,7 +2,7 @@
 
 import functools
 import io
-from hashlib import md5
+from hashlib import blake2b
 from pathlib import Path
 
 from . import CSS, DEFAULT_OPTIONS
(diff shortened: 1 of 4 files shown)


pmjdebruijn:
blake2b seems to be at least as fast as sha1/md5 even for very short messages (16 bytes), but gains relative speed for longer messages.

I realize that md5/sha1 aren't used in either security or performance critical paths right now, so there aren't any immediately tangible gains, but on the flipside there should be little risk of regression either.

In the end, this change adds some internal consistency to the weazyprint codebase, while offering some performance gains in extreme use-cases, with little to no downsides as far as I can tell.

liZe:
Thanks for the PR.

> I realize that md5/sha1 aren't used in either security or performance critical paths right now, so there aren't any immediately tangible gains, but on the flipside there should be little risk of regression either.

I agree, I don’t think that this change is required, as it doesn’t change anything.

> In the end, this change adds some internal consistency to the weazyprint codebase, while offering some performance gains in extreme use-cases, with little to no downsides as far as I can tell.

Performance problems in WeasyPrint are not mainly caused by the hashing method 😄.

About consistency, well… Setting different digest sizes introduces another form of inconsistency. We could probably use 6 in `stream.py` and avoid the slice, and we have to manually set lower values elsewhere because we don’t want 64-char-long filenames and ids.

So, I’d prefer to the code the way it is, as it’s been tested by real users for some time and works well 😄.

Issue #1981, opened 2023-10-08
Variable Font Instantiation Limited


pmjdebruijn:
[variable-font-bug.zip](https://github.com/Kozea/WeasyPrint/files/12840705/variable-font-bug.zip)

It seems that WeasyPrint instantiates each font-weight when it first encounters it, even if later on, it encounters a different font-variation-settings: 'wght' applied to that font-weight. Seeming font-weight is used for "naming" the embedded font, so Roboto-Flex-Bold and font-variation-settings is used to actually determine the real visual weight (wght). However once real visual wght X is assigned/applied to "Bold" it remains fixed after that.

While the current variable font implementation is already quite useful in it's current form, as it allows users to finetune font weights, more creative uses of variable fonts quickly leads to very confusing situations.

If this is not so trivial to fix, maybe consider adding a CLI warning about this limiation?

pmjdebruijn:
Note, it seems that only "Regular" and "Bold" font-weights are used for naming instances, which means that users are limited to two arbitrary real visual weights. Expanding this to include at least Light, Medium, SemiBold would already significantly improve things, although this won't be a "proper" fix.

That being said, given that fonts can have multiple axis of variation, you'll quickly end up having to do something like font-family + '-' + trunc(hmac-sha256(font-family, axis1 + axis2 + axis3 + axis4 + axis5 + etc), 10) I guess?

pmjdebruijn:
In my example, https://github.com/Kozea/WeasyPrint/blob/main/weasyprint/pdf/stream.py#L18 only gets called twice.

pmjdebruijn:
Presumably this is where the problem may reside:
https://github.com/Kozea/WeasyPrint/blob/main/weasyprint/pdf/stream.py#L326

pmjdebruijn:
Removing:
https://github.com/Kozea/WeasyPrint/blob/main/weasyprint/pdf/stream.py#L324
Seems to solve the issue

pmjdebruijn:
Only a minor metadata cosmetic issue remains, all variations are either encoded as "Roboto-Flex" (implicitly regular), or "Roboto-Flex-Bold", presumably this is Pango misbehaving? As I'd expect wght 450-550 to be encoded as "Roboto-Flex-Medium"...

pmjdebruijn:
The following:

```
[...]

PR #1988, opened 2023-10-15
Fix variable font instantiation (#1981)

pmjdebruijn:


liZe:
Thanks for the PR.

It does fix #1981, that’s great!

I wonder why I added this line in commit b8b9e8c.

>This will probably lead us to other subtle and frightening problems in the future, but we’ll let our dear users and customers find them for us.

Bugs with variable fonts are frightening, and I hope that it’s the last problem we’ll find with this code! I’m a bit afraid that this change could store the same font multiple times if some OpenType features are enabled/disabled, but I can’t find a sample with this problem, so let’s merge this PR!

PR #2000, opened 2023-11-04
Added support for Select fields as combo boxes

VagnerNico:
I've decoded two PDFs, one with the combo box and another blank and extracted the binary info necessary to render it and applied to the `anchors.add_inputs` function.

### DIFF
--- a/docs/api_reference.rst
+++ b/docs/api_reference.rst
@@ -765,7 +765,7 @@ The ``resize``, ``cursor``, ``caret-*`` and ``nav-*`` properties are **not**
 supported.
 
 The ``appearance`` property is supported. When set to ``auto``, it displays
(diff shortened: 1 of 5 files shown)


liZe:
Hi!

Some elements you insert in the PDF are hardcoded values (you probably found in your PDF), but they should be real data instead (or even not set at all): `M`, `BC`, `Ff`, etc. You can read the PDF specification to know what they mean and to find the required values.

Moreover, could you please provide a HTML sample showing the result? I can’t get an example that’s working for me.

liZe:
> Hello @liZe, have you had time to check this PR? Is there anything that you want me to change?

I’ll add a comment as soon as I find the time to review it! I’m currently working on other issues.

liZe:
@VagnerNico Feedback is welcome!

PR #2006, opened 2023-11-22
Fix WeasyPrint value for input/textarea

VagnerNico:
Previously we were filling it with a `pydyf.String(None)` when the HTML had no value or even an empty string as value, which was rendering the field with a `None` starting value.

This PR fix this behavior using an empty string on the `/V` if the value is `None`.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -482,15 +482,17 @@ def test_partial_pdf_custom_metadata():
     assert b'value' in stdout
 
 
-@pytest.mark.parametrize('html, field', (
-    (b'<input>', b'/Tx'),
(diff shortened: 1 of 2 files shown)


liZe:
Thanks a lot for this pull request! 😍

I’ll add a test to check that we don’t reintroduce this bug.

I can also fix the [styling errors](https://github.com/Kozea/WeasyPrint/actions/runs/6956973313/job/18938179118?pr=2006#step:10:12). We can safely remove the comment as the code is pretty self-explanatory.

liZe:
Thanks for the quick fix.

You can remove the example and the script, I’ll add a unit test so that we’re sure that the bug will never come back again 😄.

PR #2014, opened 2023-12-07
Remove warnings for PDF/A, PDF/UA compatibility.

m000:
Avoids spamming production logs with PDF/A, PDF/UA compatibility warnings.

### DIFF
--- a/weasyprint/pdf/pdfa.py
+++ b/weasyprint/pdf/pdfa.py
@@ -14,17 +14,11 @@ from functools import partial
 
 import pydyf
 
-from ..logger import LOGGER
 from .metadata import add_metadata
 
 
 def pdfa(pdf, metadata, document, page_streams, compress, version):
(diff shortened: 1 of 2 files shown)


liZe:
Hi!

Thanks for this pull request.

Actually, I think that you can even remove these messages. We’ve probably spotted most of the outstanding bug for now, some issues have been opened but it looks like it works well enough that people can actually generate PDF/A and /UA for real-life cases.

Could you please change the PR to actually remove these messages?

m000:
> Hi!
> 
> Thanks for this pull request.
> 
> Actually, I think that you can even remove these messages. We’ve probably spotted most of the outstanding bug for now, some issues have been opened but it looks like it works well enough that people can actually generate PDF/A and /UA for real-life cases.
> 
> Could you please change the PR to actually remove these messages?

Easy-peasy. Done & thanks!

liZe:
Thank you!

PR #2017, opened 2023-12-11
Support var() in shorthand and multiple-value functions

### LINKED ISSUE #1219: Handle var function in properties with multiple values
I'm using weasyprint in one of my projects and I would like to use CSS variables but they don't work correctly.
I have these rules like these in my `style.css`:
```css
:root {
  --primary-color: #2ea44f;
}

.table thead {
    background-color: var(--primary-color);
    color: #fff;
    text-align: left;
}

.table tbody tr:last-child {
    border-bottom: 2px solid var(--primary-color);
}
```
The rule `.table thead {...}`  works normally and this rule `.table tbody tr:last-child{...}` doesn't work and I get a message in the terminal:
```
WARNING: Ignored `border-bottom: 2px solid var(--primary-color)` at 90:5, invalid value.
```

Any help, please?
Thank you in advance.

liZe:
Fix #1219.

Possible improvement before merging: support `var()` in other functions.

Let’s forget for now: ~support other functions such as `calc()` (see #357.)~

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,10 +8,10 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ['3.11']
+        python-version: ['3.12']
         include:
           - os: ubuntu-latest
-            python-version: '3.7'
+            python-version: '3.8'
           - os: ubuntu-latest
             python-version: 'pypy-3.8'
     steps:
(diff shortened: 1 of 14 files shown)


liZe:
OK, let’s focus on `var` in functions first, it’s already really complicated.

Who on hell allowed things like

`--red: 100; --green: 80; --blue: 0; color: rgb(var(--red) var(--green) var(--blue))`

or

`--red-green: 100, 0; color: rgb(var(--red-green), 200)`

in CSS?

liZe:
> Possible improvement before merging: support `var()` in other functions.

Done!

PR #2018, opened 2023-12-11
Add HEIF/HEIC image support.

Corvus4n6:
Added the ability to render HEIF or HEIC (Apple) images into PDFs.

### DIFF
--- a/weasyprint/images.py
+++ b/weasyprint/images.py
@@ -14,12 +14,15 @@ from xml.etree import ElementTree
 
 import pydyf
 from PIL import Image, ImageFile, ImageOps
+from pillow_heif import register_heif_opener
 
 from . import DEFAULT_OPTIONS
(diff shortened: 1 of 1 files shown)


liZe:
Hi!

Thanks for this pull request.

The HEIF format is not supported in Pillow because of [patent problems](https://github.com/python-pillow/Pillow/issues/2806#issuecomment-338192264), and for the same reason [it’s not supported in Chrome and Firefox](https://caniuse.com/heif). To be honest, I don’t think that it’s a good idea to support it WeasyPrint, I’ll follow what Pillow does.

If you want to take the legal risk, you actually don’t even need to change WeasyPrint’s code. You can write a very small Python wrapper around WeasyPrint that uses the code of this PR before importing WeasyPrint.

But I won’t be there when the police comes for you! 😄

PR #2022, opened 2023-12-14
Filter args with value None before calling html5lib.parse

ViktorShev:
Related to: https://github.com/Kozea/WeasyPrint/issues/2023


liZe:
Hi!

Thanks for this pull request. Could you please explain which problem you’re trying to solve?

ViktorShev:
@liZe Yes!, just filed an issue.

liZe:
Thanks a lot for this PR and the related issue! I’ve added tests, and allowed text-based file objects for `CSS` class too. Feedback is welcome!

PR #2035, opened 2024-01-09
Reduction of "display, float, position" calculation

ssjkamei:
I tried to reduce the process of judging every `__missing__` with `ComputedStyle` in css.

### DIFF
--- a/tests/test_float.py
+++ b/tests/test_float.py
@@ -212,6 +212,46 @@ def test_floats_8():
 
 
 @assert_no_logs
+def test_floats_10():
+    # adjacent-floats-001
+    page, = render_pages('''
+      <div style="float: left">
+        <div style="float: inherit; width: 100px; height: 30px;">
+          test 1
+        </div>
+        <div style="float: inherit; width: 100px; height: 30px;">
+          test 2
+        </div>
+      </div>''')
+    html, = page.children
+    body, = html.children
+    div, = body.children
+    div_1, div_2 = div.children
+    assert outer_area(div_1) == (0, 0, 100, 30)
+    assert outer_area(div_2) == (100, 0, 100, 30)
+
+
+@assert_no_logs
+def test_floats_11():
+    # adjacent-floats-001
+    page, = render_pages('''
+      <div style="float: none">
+        <div style="float: inherit; width: 100px; height: 30px;">
+          test 1
+        </div>
+        <div style="float: inherit; width: 100px; height: 30px;">
+          test 2
+        </div>
+      </div>''')
+    html, = page.children
+    body, = html.children
+    div, = body.children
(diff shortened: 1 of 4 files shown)


liZe:
Hi!

Thanks for the proposal. This solution is appealing, as it’s simple and removes a quite dirty workaround! But I think that it’s actually broken in some corner cases. For example:

```html
<div style="float: none">
  <div style="float: inherit; display: inline">
    test 1
  </div>
  <div style="float: inherit; display: inline">
    test 2
  </div>
</div>
```

doesn’t give the same result before and after your PR (yes, that should be a test!) That’s because the current code resolves many values ('inherit', 'initial', and CSS variables for example) before storing them in the `specified` dictionary. But your code only uses the raw cascaded value as the specified value. In this example, with this PR, the specified `float` value is `inherit` when it should actually be `none`.

But you’re right, it should be possible to remove this `specified` dictionary somehow. It’s just not that easy, because the computed values of the three properties depend on the specified value for each other (see [CSS2](https://www.w3.org/TR/CSS21/visuren.html#dis-pos-flo)), and that this specified value is somewhere between the cascaded value and the computed value.

ssjkamei:
Thank you!

Is the test written correctly?
What I was wondering is that the original code didn't seem to come through either.

liZe:
> Is the test written correctly?

Yes, that’s great. Thanks a lot, my comment was more a reminder for me than an advice for you! 🙏️

> What I was wondering is that the original code didn't seem to come through either.

The current specified values go through the whole `__missing__` code before [being stored](https://github.com/Kozea/WeasyPrint/blob/c3f0307fce65d45b4d875ebb1d925a3f06d75804/weasyprint/css/__init__.py#L784).

ssjkamei:
Sorry, I guess I didn't communicate that well.

When I checked, I wanted to say that the previous code was also what was causing the error. (but wasn't sure if that was correct)

I understood from what you said before that you need to check the inherited value and if there is no float but a display is specified in a higher level node, you need to take that default value. Thanks very much for your help.

liZe:
> When I checked, I wanted to say that the previous code was also what was causing the error. (but wasn't sure if that was correct)

I don’t get the same result.

Here’s what I get from current `main` branch:
Capture d’écran du 2024-01-14 14-08-52

and from your branch:
Capture d’écran du 2024-01-14 14-08-32

The first image is correct, as `float` should be seen as `none`, so the used value of `display` is `inline` as specified. The second image is incorrect, because `float` is set to `inherit` (not resolved to `none`, that’s the bug), and when `float` is set to something different from `none` then `display` is ignored.

liZe:
Closing for now, please add a comment if you want to share new commits on this topic.

Issue #1968, opened 2023-09-19
`tr` with fixed height shrinking below height of content


ericbf:
It seems that WeasyPrint treats the height of `tr` elements different than the browser, causing inconsistent table row heights for table rows with a fixed height less than the size of the content therein.

In the browser, if the height of a `tr` is less than the height of the content, the `tr` will take the height of the content instead:

<img width="605" alt="Screenshot 2023-09-19 at 2 21 29 PM" src="https://github.com/Kozea/WeasyPrint/assets/2483476/8be84e8d-5950-4604-abe3-500faece8098">

However, when that same markup is rendered into a PDF with WeasyPrint, the `tr` will actually use the lesser height, not expanding to the height of the content as expected:

<img width="635" alt="Screenshot 2023-09-19 at 2 21 49 PM" src="https://github.com/Kozea/WeasyPrint/assets/2483476/51bcf7f1-7faa-4a19-bf9e-7bffe931c091">

This is not the case for the `td` elements, where setting a height smaller than the content does not cause the `td` to shrink below the size of the content (see the same example above).

Maybe it has to do with the padding on the `td`s? It seems to be squashing the vertical cell padding. Not sure exactly what behavior is taking place, technically.

Here’s the markup for the example:

```html
<table style="border-collapse: collapse; width: 100%; border: 1px solid rgb(126, 140, 141);" border="1">
	<tbody>
		<tr style="height: 100px;">
			<td style="padding: 5px; vertical-align: middle; border: 1px solid rgb(126, 140, 141); width: 100%;">Tall Row</td>
		</tr>
		<tr>
			<td style="padding: 5px; height: 100px; vertical-align: middle; border: 1px solid rgb(126, 140, 141); width: 100%;">Tall Cell</td>
		</tr>
		<tr style="height: 5px;">
			<td style="padding: 5px; vertical-align: middle; border: 1px solid rgb(126, 140, 141); width: 100%;">Short Row</td>
		</tr>
		<tr>
			<td style="padding: 5px; height: 5px; vertical-align: middle; border: 1px solid rgb(126, 140, 141); width: 100%;">Short Cell</td>
		</tr>
	</tbody>
</table>
```

ericbf:
Hello is anyone going to triage this one?

PR #2039, opened 2024-01-14
Fix table row height calculation to use cell's border-height

alexandergitter:
This is a fix for #1968.

For rows with a specified height, I changed the code that calculates the maximum of the children cell's heights. It now uses the border-box-height instead of just `height`, similar to how it's done in the `row.height == 'auto'` case.


grewn0uille:
Thanks!

PR #2043, opened 2024-01-18
Fix tspan not respecting text-anchor

michaellisitsa:
text-anchor attribute on `tspan` is not respected. It previously worked in v60.1 however the logic for text-anchor was moved into the `svg/__init__.py` file and caused this edge case not be handled, and there were no specific test cases.

See example below where text-anchor="middle" doesn't do anything

Tests pass

Replit reproducing bug: https://replit.com/@mlisitsa/Python-Weasyprint
Example:
```html
<svg
  width="600"
  height="217"
  role="img"
  viewBox="0 0 600 217"
  style="pointer-events: all; width: 100%; height: 100%"
>
  <text x="10" y="30" class="small">
    <tspan x="0" dx="0" dy="0" text-anchor="middle" >Middle</tspan>
    <tspan x="0" dx="0" dy="15">Start</tspan>
  </text>
</svg>
```

### DIFF
--- a/tests/draw/svg/test_text.py
+++ b/tests/draw/svg/test_text.py
@@ -280,6 +280,7 @@ def test_text_tspan(assert_pixels):
       </svg>
     ''')
 
+
 @assert_no_logs
 def test_text_tspan_anchor_middle(assert_pixels):
     assert_pixels('''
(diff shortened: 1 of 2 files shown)


grewn0uille:
Thanks!

PR #2048, opened 2024-01-24
Replace non-virtualenv installation instructions with distribution packages

### LINKED ISSUE #2021: Recommended installation approach broken on Alpine 3.19
First of all, thanks for this amazingly useful piece of software!

In the documentation, the installation for Alpine without a virtualenv (see https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#alpine-3-12) is no longer working with Alpine 3.19, as Python’s package directory is now [marked](https://gitlab.alpinelinux.org/alpine/infra/aports/-/commit/6c2a300f6787544f3194e7f6cc19058baf9b7419) as [externally managed](https://packaging.python.org/en/latest/specifications/externally-managed-environments/); see also https://www.alpinelinux.org/posts/Alpine-3.19.0-released.html.
```
$ docker container run --rm alpine:3.19 /bin/sh -c "apk --no-cache add py3-pip py3-pillow py3-cffi py3-brotli gcc musl-dev python3-dev pango && pip3 --no-cache-dir install weasyprint"
...
error: externally-managed-environment

× This environment is externally managed
╰─>
    The system-wide python installation should be maintained using the system
    package manager (apk) only.

    If the package in question is not packaged already (and hence installable via
    "apk add py3-somepackage"), please consider installing it inside a virtual
    environment, e.g.:

    python3 -m venv /path/to/venv
    . /path/to/venv/bin/activate
    pip install mypackage

    To exit the virtual environment, run:

    deactivate

    The virtual environment is not deleted, and can be re-entered by re-sourcing
    the activate file.

    To automatically manage virtual environments, consider using pipx (from the
    pipx package).

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

Possible solutions to fix the documentation:
- point towards the version available [through `apk`](https://pkgs.alpinelinux.org/packages?name=weasyprint)
- update the installation suggestion to use `pipx`
- note that for Alpine >= 3.19 installation without virtualenv is not supported

alexandergitter:
Fixes #2021

### DIFF
--- a/docs/first_steps.rst
+++ b/docs/first_steps.rst
@@ -70,12 +70,12 @@ in a `virtual environment`_ using `pip`_::
 .. _pip: https://pip.pypa.io/
 
 
-Alpine ≥ 3.12
+Alpine ≥ 3.14
 +++++++++++++
 
-To install WeasyPrint without a virtualenv, you need the following packages::
+To install WeasyPrint using your distribution’s package::
 
-  apk add py3-pip py3-pillow py3-cffi py3-brotli gcc musl-dev python3-dev pango
+  apk add weasyprint
 
 To install WeasyPrint inside a virtualenv using wheels (if possible), you need
 the following packages::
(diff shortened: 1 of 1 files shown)


grewn0uille:
Thanks!

jnoordsij:
Thanks for fixing this!

PR #2049, opened 2024-01-26
add logging.set_log_level function

scottgigante:
currently, log level can only be set by the CLI and does not touch fontTools logging (which can be very verbose). This PR adds functionality to set all three loggers log level at once.


liZe:
Hi!

Thanks for this pull request. I’ve fixed a couple of things to make it work (there was an error in the module name).

Instead of setting level to ERROR when using "quiet", I’d like to remove fontTool’s handlers, but `logging.getLogger('fontTools').handlers` seems to be empty. Could you please share an example that shows some of fontTools’ logging messages?

liZe:
@scottgigante Don’t hesitate to add a comment if needed!

PR #2059, opened 2024-02-08
Test bidirectional position x must be invariant

kygoh:
> ```html
> <div style="float: left; border-right: 100px solid black; direction: rtl">abc</div>
> ```
>
> `x` is **by definition** the distance from the left, it doesn’t depend on the text direction. Believe me, that’s a really bad idea to change this. 😄
>
_Originally posted by @liZe in https://github.com/Kozea/WeasyPrint/issues/2056#issuecomment-1932554075_


liZe:
Thank you!

PR #2067, opened 2024-02-12
suggest "dnf" instead of "yum" to install Fedora packages

FelixSchwarz:
`yum` was superseeded by `dnf` a long time ago.


liZe:
Thank you!

PR #2074, opened 2024-02-15
Improve import startup time

### LINKED ISSUE #2073: Improve import startup time
On my machine, importing `weasyprint` takes more that half a second:

```sh
$ time python -c 'import weasyprint'

real	0m0.638s
user	0m0.494s
sys	0m0.101s
```

Which can get annoying e.g. when running a Django server in dev. with autoreload. Did a bit of profiling to see where it was coming from[^1]:

image

And it seems to be spread across different modules.

I will propose a workaround in a follow up PR, although there isn't only one way to deal with this so open to alternatives 

[^1]: `python -X importtime -c 'import weasyprint' 2> import.log && tuna import.log`

Viicos:
Attempt to fix #2073. With these changes:

```sh
$ time python -c 'import weasyprint'

real	0m0.124s
user	0m0.106s
sys	0m0.020s
```

### DIFF
--- a/weasyprint/__init__.py
+++ b/weasyprint/__init__.py
@@ -7,8 +7,10 @@ importing sub-modules.
 
 import contextlib
 from datetime import datetime
+import importlib
 from os.path import getctime, getmtime
 from pathlib import Path
+from typing import TYPE_CHECKING
 from urllib.parse import urljoin
 
 import cssselect2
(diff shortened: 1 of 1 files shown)


weasyprint/__init__.py
```diff
@@ -181,14 +183,20 @@ def __init__(self, guess=None, filename=None, url=None, file_obj=None,
...
             return [HTML5_UA_STYLESHEET, HTML5_UA_FORM_STYLESHEET]
         return [HTML5_UA_STYLESHEET]
 
     def _ua_counter_style(self):
+        from .html import HTML5_UA_COUNTER_STYLE
+
         return [HTML5_UA_COUNTER_STYLE.copy()]
 
     def _ph_stylesheets(self):
+        from .html import HTML5_PH_STYLESHEET
```
Viicos, line 198: Noticed with failing tests that I also had to import them in functions as well. So the `__getattr__` machinery I added is only there for backwards compatibility in case e.g. users are importing `from weasyprint import HTML_UA_COUNTER_STYLE`

Viicos:
Tested locally and tests are passing

PR #2086, opened 2024-03-04
Fixed selects with empty values displaying None

VagnerNico:
Previously for optional selects with an option without text WeasyPrint was rendering a `None` in the PDF combobox option, so I've started to explicitly send an empty string when the text is `None`.


liZe:
Thank you very much!

Issue #2106, opened 2024-03-22
UnboundLocalError: local variable next_box_height is referenced before assignment


filipweidemann:
Hey there!

This is my first issue here, so let me start by saying that this project is absolutely fantastic. 

Back to topic: 
we recently observed with one of our customers (using extremely weird data), that WeasyPrint has a slight programming error inside of `layout/column.py:220` – here it tries to calculate `next_box_height - empty_space` but unfortunately a case can occur in which `next_box_height` is unbound. This is confirmed by myself after opening the project and inspecting the Pyright LSP output, which also complains about this with a red inline hint of "next_box_height is possibly unbound"

The sad thing is that I have absolutely no clue on how to reproduce this. It seems to be a very weird edge case with nested column layouts & weird `&nbsp` characters that only one of our customers seems to trigger. 
I had no luck on narrowing this down to provide you with an example.

However, this is very clearly in itself an issue (which we just happened to run into in prod) and can be fixed without examples I assume?

My approach would be to just initialize `next_box_height` to `0` in the proper scope, this should probably catch this error. All other cases *should* remain the same, since `next_box_height` would be overwritten in those scopes.

I'll start by opening a PR with my intended fix, if you have any other ideas, just let me know!

Again, thanks very much for maintaining, I'll link the PR here shortly…

liZe:
Hi!

Thanks for your bug report and your pull request.

> The sad thing is that I have absolutely no clue on how to reproduce this. It seems to be a very weird edge case with nested column layouts & weird `&nbsp` characters that only one of our customers seems to trigger.
> I had no luck on narrowing this down to provide you with an example.

I’d really like to get a test for this, because we’ve gone through so many bugs like this for columns and we have to avoid regressions. I’ll try to find a way to reproduce. If you have a large example, don’t hesitate to share it. If you don’t want to do this publicly, you can send it to me by mail.

filipweidemann:
Hey, thanks for answering! 

I can't really dig into this right now, but I will try to come up with a reproduction once I have some spare time at work as well. I also can't rule out that this error is only triggered on malformed HTML because we had proper issues with this when working with customer data. 

Currently the bug is not being triggered anymore, this seems to confirm my suspicion that it is somehow related to the way our content looked like. It also only happened when trying to render columns and/or tables nested inside other columns. 

Thanks for looking into it and I'll add an update here once I found something that specifically triggers this error.

liZe:
[...]

PR #2107, opened 2024-03-22
fix: initialize next_box_height to avoid unboundlocalerrors

filipweidemann:
As promised in #2106, this is my proposed fix for `UnboundLocalError`s inside of the `layout/column.py` module.

I am still not sure yet if initializing the value with `0` is the right move here, however, in our case, monkey-patching it this way seemed to work without any further problems. 

Let me know if this is fine.

### DIFF
--- a/weasyprint/layout/column.py
+++ b/weasyprint/layout/column.py
@@ -176,6 +176,9 @@ def columns_layout(context, box, bottom_space, skip_stack, containing_block,
                 new_boxes.append(new_box)
                 column_skip_stack = resume_at
 
+                # Avoid edge cases resulting in a UnboundLocalError in line 223
+                next_box_height = 0
+
                 # Calculate consumed height, empty space and next box height
                 in_flow_children = [
                     child for child in new_box.children


weasyprint/layout/column.py
```diff
@@ -176,6 +176,9 @@ def columns_layout(context, box, bottom_space, skip_stack, containing_block,
                 new_boxes.append(new_box)
                 column_skip_stack = resume_at
 
+                # Avoid edge cases resulting in a UnboundLocalError in line 223
```
FelixSchwarz, line 179: I wouldn't add a comment here. To me it is obvious that there are multiple conditions where `next_box_height` is not set.

FelixSchwarz:
I basically know next to nothing about the WeasyPrint code but the change looks good to me.

liZe:
> I basically know next to nothing about the WeasyPrint code but the change looks good to me.

It looks good to me too, but I’d like to have a way to reproduce the problem before merging, so that we can add a test. See #2106.

liZe:
Closing, as we didn’t find a way to reproduce this bug. If it happens again, it may come from another error somewhere else that we don’t want to let pass silently.

PR #2123, opened 2024-04-09
Add a setup.cfg file for flake8 to ignore the venv/ directory.

xavidotron:
The instructions on https://doc.courtbouillon.org/weasyprint/stable/contribute.html create a virtual environment under a `venv/` directory.

The `venv/bin/python -m flake8` command, on the same page, by default checks for style issues everywhere under the current directory, including under the `venv/` directory. This is unnecessarily slow and noisy.

By creating a `setup.cfg` file, we can exclude `venv/` by default.

(`tox.ini` or `.flake8` files would work as well, I picked `setup.cfg` arbitrarily.)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,3 @@
+[flake8]
+exclude =
+    venv/


liZe:
Hi!

Thanks for this pull request.

We actually tried hard in the past to remove all the different configuration files, and we’ll keep only `pyproject.toml`, so we won’t merge your pull request.

And unfortunately, [Flake8 doesn’t support `pyproject.toml`](https://github.com/PyCQA/flake8/issues/234).

The solution we chose is to use ruff instead of Flake8 and isort. We already did this for other smaller projects, we’ll do this for WeasyPrint as soon as we find the time.

Issue #2124, opened 2024-04-10
Add support for border-image properties


xavidotron:
[CSS Backgrounds and Borders Module Level 3](https://www.w3.org/TR/css-backgrounds-3/#border-images) defines a series of properties to use images for box borders. Especially with SVG images, this can allow for fancy borders beyond what can otherwise be defined in CSS.

This includes the following properties:
* border-image
* border-image-source
* border-image-slice
* border-image-width
* border-image-outset
* border-image-repeat

I'd like to implement these, and since (unless you prefer otherwise) I'd like to implement this in a series of pull requests instead of all in one huge pull request, I wanted to create an issue to attach my pull requests to.

liZe:
Closed by #2125.

PR #2125, opened 2024-04-10
Add basic support for border-image-source and border-image-slice.

xavidotron:
This adds basic support for border-image-source and border-image-slice.

This is sufficient for common uses of image borders in CSS.

It does not yet support non-default values for border-image-width, border-image-outset, or border-image-repeat, which I plan on addressing in future pull requests.

This is my first substantial contribution, so let me know if I should be putting this logic in a different place or otherwise approaching it differently. Thanks for your review!

First part of #2124.

### DIFF
--- a/docs/api_reference.rst
+++ b/docs/api_reference.rst
@@ -558,10 +558,10 @@ background layers per box), including the ``background``, ``background-color``,
 WeasyPrint also supports the `rounded corners part`_ of this module, including
 the ``border-radius`` property.
 
-WeasyPrint does **not** support the `border images part`_ of this module,
-including the ``border-image``, ``border-image-source``,
-``border-image-slice``, ``border-image-width``, ``border-image-outset`` and
-``border-image-repeat`` properties.
+WeasyPrint also supports the `border images part`_ of this module, including the
+``border-image``, ``border-image-source``, ``border-image-slice``,
+``border-image-width``, ``border-image-outset`` and ``border-image-repeat``
+properties.
 
 WeasyPrint does **not** support the `box shadow part`_ of this module,
 including the ``box-shadow`` property. This feature has been implemented in a
--- a/tests/css/test_expanders.py
+++ b/tests/css/test_expanders.py
@@ -400,6 +400,62 @@ def test_border_radius_invalid(rule, message):
 
 @assert_no_logs
 @pytest.mark.parametrize('rule, result', (
+    ('url(border.png) 27', {
+        'border_image_source': ('url', 'https://weasyprint.org/foo/border.png'),
+        'border_image_slice': ((27, None),),
+    }),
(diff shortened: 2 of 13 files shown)


liZe:
That’s great, thanks a lot! The code seems to be almost perfect. ❤️

I’ll fix a couple of things and push small commits so that you can easily track what I’ve changed.

I think that we’ll merge this PR just after 42 is released (hopefully on Friday), because it already contains a lot of features and may cause a lot of bug reports.

xavidotron:
Great, thanks for your review!

liZe:
It looks like I’ve changed a lot of lines, but the modifications are actually minor. Thanks a lot, that’s a great pull request! 🙏🏽 

Here are the commits I’ve added:

- Rename tests and remove comments related to specification (you can ignore this commit, that’s a problem in my head).
- Simplify the SVG file used for tests (another pathological behavior).
- Separate [specified values](https://www.w3.org/TR/css-cascade/#specified-value) from [computed values](https://www.w3.org/TR/css-cascade/#computed). If you don’t know the difference between specified and computed values, this separation looks quite strange, but that’s important for many internal aspects of CSS. Blindly following what’s in the specification for this topic is always a good idea IMO, it avoids complex problems in the future. 😄
- Clean some code, mainly to use longer variable names and follow the famous (but unwritten) WeasyPrint code style. I’ve also used the default image sizing algorithm that can handle funny cases like images with no intrinsic size.
- Add CSS validation tests. We try to test specified values for complex cases, and generally rely on the rendering tests to check computed values. We’ll add more rendering tests when all the related properties are supported (I may have broken computed values for unsupported properties).

I also rebased the branch on top of "main", mainly to use and check the new linting rules.

If you want to support other properties, don’t hesitate to add new commits to this PR, we have some time before we merge it. There’s also the `border-image` shorthand to add into [`expanders.py`](https://github.com/Kozea/WeasyPrint/blob/main/weasyprint/css/validation/expanders.py).

xavidotron:
I believe my existing change to `expanders.py` handles the `border-image` shorthand, though I may have missed something.

xavidotron:
I believe this now supports all the border-image-* properties, including some basic test coverage.

liZe:
> I believe my existing change to `expanders.py` handles the `border-image` shorthand, though I may have missed something.

I missed that!

> I believe this now supports all the border-image-* properties, including some basic test coverage.

Thank you! If everything is OK, it may even be ready for version 62.

liZe:

All the relevant tests of the W3C testing suite pass. 🚀

We could add more drawing tests in WeasyPrint, but the code is already in very good shape and W3C tests give a good reason to be optimistic. Let’s merge if everything’s OK for you!

xavidotron:
Sounds great! This works for my needs, so I'd be happy for it to be merged. Thanks for your work to improve this as well!

PR #2139, opened 2024-04-28
Add support for mask-border-* properties

xavidotron:
Add support for the mask-border-* properties as specified in https://drafts.fxtf.org/css-masking/#the-mask-border.

These can be handy when you want a background to extend exactly to an irregular image border, but don't want the background to stretch the same way the border does. It can also be useful for other cases where you want a non-rectangular box with a size depending on its content, and such.

This was surprisingly straightforward to add since I could reuse a lot of code from #2125.

Note for any comparison testing that Chrome uses -webkit-mask-box-image-* properties rather than the "standard" names.

Let me know if you need anything else. Thanks for your consideration!

### DIFF
--- a/tests/css/test_expanders.py
+++ b/tests/css/test_expanders.py
@@ -456,6 +456,74 @@ def test_border_image_invalid(rule, reason):
 
 @assert_no_logs
 @pytest.mark.parametrize('rule, result', (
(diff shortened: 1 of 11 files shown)


liZe:
Thank you!

Issue #1585, opened 2022-03-01
'created' timestamp seems very low; regarding as unix timestamp


zopyx:
Where is the message `'created' timestamp seems very low; regarding as unix timestamp` coming from and why?
I have a fresh Ubuntu 21.10 installation, EXT4 filesystem, the timestamps on the filesystem look fine.

liZe:
That’s a warning given by [fontTools](https://github.com/fonttools/fonttools/blame/main/Lib/fontTools/ttLib/tables/_h_e_a_d.py#L57). There’s probably a font on your system whose "created" field is wrong.

liZe:
Closing. If there’s a problem, it’s definitely in fontTools that should not display logs on stderr by default.

(I wasn’t sure about silencing library logs by default, but #412 really convinced me.)

Django-27:
I ran into the same issue and found this warning in the fontTools source code and followed the lead to solve it in the hope that it will help other people who come across it


1. first， find the warning location

Envs\py311\Lib\site-packages\fontTools\ttLib\tables\_h_e_a_d.py
`
for stamp in "created", "modified":
    if value < 0x7C259DC0:  # January 1, 1970 00:00:00
        log.warning(
            "'%s' timestamp seems very low; regarding as unix timestamp", stamp
        )
        value += 0x7C259DC0
        setattr(self, stamp, value)`

2. second, find the .ttf file in your project 

3. third, open .ttf file edit website like https://font.qqe2.com/#, and modify the content like the picture

![image](https://github.com/Kozea/WeasyPrint/assets/10075605/b5ce3c2c-505a-4075-a87f-b40e925bd8b5)

kdipippo:
Created https://github.com/fonttools/fonttools/issues/3538 to advise, as I think either strategy of checking the font when it's added in `stream.py::add_font()` or checking the results of `Subsetter::subset()` requires this input.

PR #2169, opened 2024-05-28
ISSUE-1585 | Surface fontTools logs for easier investigation

kdipippo:
Work-in-progress.

Taking the feedback from the following 2 issues into this PR:
- https://github.com/Kozea/WeasyPrint/issues/1585
- https://github.com/fonttools/fonttools/issues/3538

### DIFF
--- a/weasyprint/pdf/stream.py
+++ b/weasyprint/pdf/stream.py
@@ -14,6 +14,19 @@ from ..text.constants import PANGO_STRETCH_PERCENT
 from ..text.ffi import ffi, harfbuzz, pango, units_to_double
 from ..text.fonts import get_hb_object_data, get_pango_font_hb_face, get_pango_font_key
 
+import logging
+
+fonttools_logging = logging.getLogger("fontTools")
+fonttools_logging.setLevel(logging.WARNING)
+fonttools_warning_messages = []
+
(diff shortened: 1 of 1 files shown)


liZe:
After thinking about the best way to solve this issue, here’s what seems to be the best solution for me.

The code needed to capture logs is already in WeasyPrint: it’s used for tests, to check that we get no warning or the intended warning messages, depending on the test.

https://github.com/Kozea/WeasyPrint/blob/6e93e95354dc968180b09986fb23c3e35c8cbfe1/tests/testing_utils.py#L73-L100

What we could do is:
- Move `CallbackHandler` and `capture_logs` to `logger.py`.
- Add a parameter to `capture_logs` to define the logger we want to capture. Its default value would be `'weasyprint'` to keep the current default behaviour.
- Use `with capture_logs('fontTools') as logs:` around the call to `subsetter.subset()` and display the logs thanks to `LOGGER.warning()`, just as you did in your PR.

What do you think of this solution?

As the code has changed a bit in `main`, you should probably create a new branch from `main`, create a new commit and push force to your branch.

liZe:
I’ve fixed the problem in 61852c4. Thanks a lot for your investigation and for your proposal. 💜

Issue #1128, opened 2020-05-27
Image prints with black background


wesleyi23:
I am trying to use Weasyprint to generate pdf reports for a Django web application. The HTML file the report is based on has images in base 64, some of the images generate with a black background instead of a white background.  I am trying to understand where this issue might be or how to go about correcting it.  Does anyone have any ideas where to start?

liZe:
Hello!

Could you please provide an example?

wesleyi23:
Sure. Here is an example HTML file (saved as a txt file):
[example.txt](https://github.com/Kozea/WeasyPrint/files/4690694/example.txt)

liZe:
Looks like a problem with CMYK, really like #315. It was supposed to be fixed in Cairo 1.15.4, I don’t know why it’s still there with 1.16.

wesleyi23:
Hey thanks!  I will double check to make sure I have the latest version of Cairo and post back what I find tomorrow.

liZe:
> Looks like a problem with CMYK, really like #315. It was supposed to be fixed in Cairo 1.15.4, I don’t know why it’s still there with 1.16.

Here is why: [the fix has been reverted](https://gitlab.freedesktop.org/cairo/cairo/-/commit/bf597b89288b6271f56031d5a20bfaf59f146d4c).

Hopefully Cairo is soon gone forever.

A quick workaround is to save your JPG file as RGB.

liZe:
The bug remains, even without Cairo.

germinator1512:
This still happens .. also converting client pictures, so a solution would be much appreciated

liZe:
> so a solution would be much appreciated

Some help would be really appreciated on this topic. If you could spend some time to understand what’s wrong in the generated PDF, that would be wonderful. Thank you!

liZe:
> Theres nothing wrong with the pdf itself, just the color.

I mean, there’s something in the generated PDF that makes the colors look inverted. Maybe there’s a place where we can tell the PDF that this image is CMYK? Reading [the PDF specification](https://opensource.adobe.com/dc-acrobat-sdk-docs/standards/pdfstandards/pdf/PDF32000_2008.pdf) and playing with [this code](https://github.com/Kozea/WeasyPrint/blob/22b35bbc53ef700761f89da98982286dfaf4b6e3/weasyprint/images.py#L42) could help to find what’s wrong and how to fix this.

endaxi:
[...]

PR #2179, opened 2024-06-07
CMYK JPEGs appearing inverted in PDF output

endaxi:
Addresses issue [#1128](https://github.com/Kozea/WeasyPrint/issues/1128)

According to [libjpeg](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/3c17063ef1ab43f5877f19d670dc39497c5cd036/libjpeg.txt#L1569-L1582) "it appears that Adobe Photoshop writes inverted data in CMYK JPEG files"

An Adobe JPEG can be identified by the presence of the [APP14](https://exiftool.org/TagNames/JPEG.html#Adobe) segment.

The code now checks for the `APP14` segment in `RasterImage` and adds a Decode Array to the XObject when rendering the CMYK JPEG. The value of the Decode Array is the inverse of the default value for DeviceCMYK according to the PDF spec. This has the effect of inverting the inverted image back to normal.

### DIFF
--- a/weasyprint/images.py
+++ b/weasyprint/images.py
@@ -66,6 +66,7 @@ class RasterImage:
         self.height = pillow_image.height
         self.ratio = (self.width / self.height) if self.height != 0 else inf
         self.optimize = optimize = options['optimize_images']
+        self.app14 = getattr(original_pillow_image, 'app', {}).get('APP14')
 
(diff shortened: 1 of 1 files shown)


liZe:
Thank you! ❤️

PR #2195, opened 2024-06-27
Update `cache` argument in docs

OmeGak:
The `image_cache` option specified in the documentation is not correct. The correct option is `cache`. Using `image_cache` leads to no cache being applied at all, as the option is completely ignored without warning. It would be useful to fail when passing invalid options, to avoid them silently not being applied.


liZe:
> The `image_cache` option specified in the documentation is not correct.

Thank you!

> It would be useful to fail when passing invalid options, to avoid them silently not being applied.

Yes, it would be useful. Comparing the `options` keys to the ones in `DEFAULT_OPTIONS` in `render` should work.

liZe:
Thanks!

OmeGak:
It was my pleasure! 🙏

PR #2199, opened 2024-07-10
Add safe argument

AndreyFrolov44:
Hi!

I suggest adding the safe argument, in which url_fetcher will be used, in which it will be impossible to download local files


liZe:
Hi!

Thanks for this pull request.

Unfortunately, I don’t think that adding a `safe` option is a good idea, because it may give the impression that using it solves all the security concerns. Avoiding access to local files is only one small part of all [the possible security problems](https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#security) that may happen with WeasyPrint, and there’s no silver bullet to avoid them all.

PR #2211, opened 2024-07-26
Fixing link to weasyprint.org/#samples/

gteti:


liZe:
Hi!

Thanks for this pull request.

That’s an example showing how links work with WeasyPrint, fixing this URL is not consistent with the explanation above. There’s only a home page on WeasyPrint’s website now, so we have to change the example. Maybe you could use the `https://www.courtbouillon.org/blog-articles/` URL, changing `/samples/` into `/blog-articles/` and WeasyPrint into CourtBouillon?

gteti:
> Hi!
> 
> Thanks for this pull request.
> 
> That’s an example showing how links work with WeasyPrint, fixing this URL is not consistent with the explanation above. There’s only a home page on WeasyPrint’s website now, so we have to change the example. Maybe you could use the `https://www.courtbouillon.org/blog-articles/` URL, changing `/samples/` into `/blog-articles/` and WeasyPrint into CourtBouillon?

Since the link is clickable in the rendered readthedocs documentation I thought it was intended for the proposed link. 
Sorry :smile:

PR #2220, opened 2024-08-01
Add extra information to debug logs

dhdaines:
Because WeasyPrint has a lot of very deeply nested recursive call chains, it is super easy to get lost in the code.  When adding debugging log messages (which WeasyPrint could really use more of) to trace, it is nice to be able to know what file/line/function they come from.

Possibly useful for bug reporting too?

### DIFF
--- a/weasyprint/__main__.py
+++ b/weasyprint/__main__.py
@@ -178,7 +178,14 @@ def main(argv=None, stdout=None, stdin=None, HTML=HTML):  # noqa: N803
         LOGGER.setLevel(logging.INFO)
     if not args.quiet:
         handler = logging.StreamHandler()
-        handler.setFormatter(logging.Formatter('%(levelname)s: %(message)s'))
+        if args.debug:
+            # Add extra information when debug logging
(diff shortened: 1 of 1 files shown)


liZe:
To be honest, problems in WeasyPrint can happen in so many places that using the debugger is often the only reliable way to understand what’s going on. Adding extra information in debug logs is a good idea, but I’m not sure that adding more debugging information could actually help to debug without being too verbose.

dhdaines:
> To be honest, problems in WeasyPrint can happen in so many places that using the debugger is often the only reliable way to understand what’s going on. Adding extra information in debug logs is a good idea, but I’m not sure that adding more debugging information could actually help to debug without being too verbose.

True!  It's possiby just me who has never developed the habit of using the Python debugger 🙂

liZe:
> True! It's possiby just me who has never developed the habit of using the Python debugger 🙂

😄 pdb is not a perfect debugger, but at least we now have the `breakpoint()` function in Python to use it easily!

PR #2242, opened 2024-08-28
Handle svg tags with no size

### LINKED ISSUE #2240: SVG rendering problem - part of picture not visible in pdf render
(continued from #2239)

Thanks, you're right - markers were missing. But after removing the markers - still the picture is not rendered correctly.
But this time i'm not getting any error message.
pdf:
image
svg: something like
image

Here's the test case - updated svg.
[test_p2.zip](https://github.com/user-attachments/files/16729627/test_p2.zip)
RG

_Originally posted by @lafar6502 in https://github.com/Kozea/WeasyPrint/issues/2239#issuecomment-2307145347_

(edit) i think nesting of SVG elements might be related to this problem

liZe:
Fix #2240.

### DIFF
--- a/weasyprint/svg/__init__.py
+++ b/weasyprint/svg/__init__.py
@@ -81,6 +81,7 @@ DEF_TYPES = frozenset((
     'mask',
     'path',
     'pattern',
+    'symbol',
 ))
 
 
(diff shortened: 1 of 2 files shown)


liZe:
It would be nice to add a test before merging.

PR #2243, opened 2024-08-29
Enable limits to domain/scheme for external resources

dmjohnsson23:
The [WeasyPrint documentation](https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#access-to-local-files) states that, for security reasons, it is best to use a custom URL fetcher that either doesn't allow `file://` URLs, or that filters these to known safe values. However, there is no way to accomplish this via the CLI--only via Python.

This pull request expands the `default_url_fetcher` into an extensible class with two new properties that will allow limiting the fetcher either to certain schemes (thus enabling a user to only allow HTTP and HTTPS resources) or to certain domains (allowing a user to limit resources to coming from trusted domains).

I've also exposed these two new options to the CLI as `--fetch-scheme` and `--fetch-domain`, though I'm open to different names for these. The default behavior if these options are not used remains the same, so this should not break backward compatibility. 

I've added tests for the new code and run the full test suite. One test (layout/test_inline.py::test_font_stretch) is failing, but I do not believe this failure is related to any change that I made.

Using a class for the URL fetcher rather than a function will also open the door to even further improvements in the future, such as potentially allowing file paths but limiting them to certain parent directories, or mapping URLs to paths, etc..., though none of these are part of this PR. However, I think there is value in having a single highly configurable URL fetcher rather than requiring all users to implement their own.

### DIFF
--- a/tests/test_url_fetcher.py
+++ b/tests/test_url_fetcher.py
@@ -1,0 +1,62 @@
+"""Test the UrlFetcher class"""
+
+import pytest
+from urllib.request import pathname2url
+import os
+
+from weasyprint.urls import UrlFetcher
+
+
+@pytest.mark.parametrize('url, allowed_schemes', (
+    ('http://weasyprint.org/', ('https',)),
+    ('file:///etc/passwd', ('http', 'https',)),
+))
+def test_forbidden_scheme(url, allowed_schemes):
+    fetcher = UrlFetcher(allowed_schemes=allowed_schemes)
+    with pytest.raises(ValueError):
+        fetcher.validate(url)
+
+
+@pytest.mark.parametrize('url, allowed_schemes', (
+    ('http://weasyprint.org/', ('http', 'https')),
+    ('https://weasyprint.org/', ('http', 'https')),
+    ('https://weasyprint.org/', ('https',)),
+    ('https://weasyprint.org/', None),
+    ('file:///home/me/my_photo.jpg', ('http', 'https', 'file')),
+))
+def test_allowed_scheme(url, allowed_schemes):
+    fetcher = UrlFetcher(allowed_schemes=allowed_schemes)
+    assert fetcher.validate(url) == url
+
+
+@pytest.mark.parametrize('url, allowed_domains', (
+    ('https://evil.net/', ('weasyprint.org',)),
+))
+def test_forbidden_domain(url, allowed_domains):
+    fetcher = UrlFetcher(allowed_domains=allowed_domains)
+    with pytest.raises(ValueError):
+        fetcher.validate(url)
+
+
+@pytest.mark.parametrize('url, allowed_domains', (
+    ('https://weasyprint.org/', ('weasyprint.org',)),
+    ('https://weasyprint.org/', ('weasyprint.org', 'example.com')),
+    ('https://example.com/', ('weasyprint.org', 'example.com')),
+    ('https://weasyprint.org/', None),
+))
+def test_allowed_domain(url, allowed_domains):
+    fetcher = UrlFetcher(allowed_domains=allowed_domains)
+    assert fetcher.validate(url) == url
+
+
+def test_fetch_https():
+    fetcher = UrlFetcher(allowed_domains=('weasyprint.org',), allowed_schemes=('https',))
+    result = fetcher('https://weasyprint.org/css/img/logotype-black.svg')
+    assert 'string' in result or 'file_obj' in result
+
+
+def test_fetch_file():
+    fetcher = UrlFetcher(allowed_schemes=('file',))
+    result = fetcher('file://%s' % pathname2url(os.path.abspath('./resources/icon.png')))
+    assert 'string' in result or 'file_obj' in result
+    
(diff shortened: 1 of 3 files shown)


liZe:
Hi!

> The [WeasyPrint documentation](https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#access-to-local-files) states that, for security reasons, it is best to use a custom URL fetcher that either doesn't allow `file://` URLs, or that filters these to known safe values. However, there is no way to accomplish this via the CLI--only via Python.

There are actually many things to do to get a safe environment, and filtering `file://` URLs is only a small part of the things users have to do.

There already have been different proposals to include different options to the CLI, but we’re generally reluctant to include these changes: they only provide a small part of what `url_fetcher` can do, and we’re afraid to get many PRs with many specific cases, causing a complex list of options to document, maintain, test.

And mixing these options could be quite a nightmare…

That’s why we usually don’t merge pull requests like this one.

We really do appreciate your effort to open this PR, and it doesn’t mean that such a feature would be useless. But we probably won’t merge it.

> Using a class for the URL fetcher rather than a function will also open the door to even further improvements in the future, such as potentially allowing file paths but limiting them to certain parent directories, or mapping URLs to paths, etc..., though none of these are part of this PR. However, I think there is value in having a single highly configurable URL fetcher rather than requiring all users to implement their own.

I understand your point of view, but I also disagree on this topic. There are so many different cases that we will never provide a class that’s actually useful to a lot of people: if you’re using Flask, or Django, or FastAPI, if you want to refuse local files, or just in a folder, or with a regular expression, or depending on the mimetype, if you want to refuse network access, or just local network, or just a list of IPs, or with a mask, if you want to limit the number of files, or the size, or… These are just simple, common examples. We’ll never cover a major part of them.

Browsers, that are massive pieces of software, don’t even provide such options by default. There are plugins for that.

Note that defining a URL fetcher is already a fine-tuned solution to handle security concerns, but that there are other options if you don’t know how to write Python. Using a VM or a container is also possible, and will provide many, many features that a CLI will never be able to provide.

dmjohnsson23:
You make good points, and there could be better approaches to the problem than what I have here.

I do think some change is needed on this front though, even if it differs from the specifics of what I have here. Many people will likely use the application in its default state, and if that cannot be secured easily, many people will simply not bother. The other potential future niceties I mentioned, such as URL mapping, may be more appropriate for a plugin, but I do not feel a plugin should be necessary for basic security considerations.

I will also add that, as my PR currently stands, many of the use cases you mentioned *are* supported. Namely:

* Refuse network access: this could be done using `--fetch-scheme file` to allow *only* local file access
* A list of IPs: this could be done by specifying multiple fetch domains using `--fetch-domain 10.0.0.25 --fetch-domain 10.0.0.50`

And, while numerous other use cases could be conceived of, I do think limiting schemes and domains does meet a very generally applicable baseline use case. I suspect most people will only want to be fetching resources from locations they control and expect to use, which I think for a large portion of users could be expressed even with a single domain: the domain of their own application. 

By way of comparison to other similar tools:

* wkhtmltopdf has a `--disable-local-file-access` option
* HTMLDOC has a `--no-localfiles` option
* pandoc has a `--sandbox[=true|false]` option that limits IO operations

So, there is a precedent at very least for PDF generation tools forbidding access to the local filesystem. Would you be open to an option for this being added?

liZe:
I understand your point of view, thanks for sharing.

The feature would be useful, for sure. But we don’t want to maintain it, because we think that there already are better alternatives.

> I will also add that, as my PR currently stands, many of the use cases you mentioned _are_ supported.

Of course, some of them are supported, otherwise that would be really useless 😄.

The problem is that many, many other use cases are not supported, and we won’t have any good reason to refuse many other pull requests to handle these cases. Handling these cases could be really useful, but that’s **a lot** of work to test, maintain, document. That’s not something we want to do, we’re just two maintainers, we prefer to let this responsibility to our users.

> And, while numerous other use cases could be conceived of, I do think limiting schemes and domains does meet a very generally applicable baseline use case.

Everybody says this, with the topics that they find "general" or "common". I know that you’re sincere here, but everybody has good arguments to defend their needs.

> By way of comparison to other similar tools:

There are many tools that can sandbox any application, and they are much more performant and flexible that what we could ever do. As a user who really cares about security, I’d really use a sandbox tool (with real security experts behind) than any "security" options added by a PR reviewed by two random people!

Again, I’m not saying that it’s not useful. I’m saying that we don’t want to take this responsibility.

> Or, alternatively, would you be open to an officially-supported plugin system that would allow specifying a URL fetcher defined in another package from the command line?

There already are different projects that provide a specific `url_fetcher`, like Flask-WeasyPrint or Django-WeasyPrint. If you want to provide and maintain specific URL fetchers, we’ll be happy to add a link to your projects in the documentation.

> E.g. something like `--url-fetcher mypackage.MyCustomFetcher`.

There’s no need to extend the CLI for that. If a user can write or download a `mypackage` module, they can actually write or download a Python script that uses the `weasyprint` module.

We already maintain a PDF generation library, a CSS parser, a CSS cascade engine, a CSS layout engine, an hyphenation library. We don’t want to maintain a sandbox system, there are already great tools for that :smile:.

liZe:
If some users look for a sandbox, they can use bubblewrap, the sandbox used by Flatpak. For example, to remove network access:

`bwrap --bind / / --unshare-net weasyprint https://weasyprint.org weasyprint.pdf`

gives

`weasyprint.urls.URLFetchingError: URLError: <urlopen error [Errno -3] Temporary failure in name resolution>`

Issue #2147, opened 2024-05-03
Create single page pdf for thermal printer


erlipsa:
Hello,
 I want create single page content pdf for directly print using thermal printer.
Here, my print design is
[invoice_print_html.docx](https://github.com/Kozea/WeasyPrint/files/15198178/invoice_print_html.docx)

Please provide all available solutions in your response.

liZe:
Hi @erlipsa,

We use GitHub issues to track bugs, not to ask for support for HTML/CSS design.

If you want some community help about a specific design, you can ask on forums such as StackOverflow, or read some documentation about CSS on MDN for example. You can also study [WeasyPrint’s samples](https://weasyprint.org/#samples). There are plenty of resources online to learn how to design documents with HTML and CSS!

We, at [Courtbouillon](https://courtbouillon.org/), also provide professional support if you want to delegate the creation of your documents.

patbengr:
This issue doesn't relate to HTML/CSS design. CSS @page rule, does not support dynamic page sizing. While CSS @page does allow specific page size declarations, it doesn't offer dynamic height adjustments.

WeasyPrint lacks support for dynamic page height adjustments. Similarly, alternatives like wkhtmltopdf also do not provide this feature.
The current workarounds involve setting a large fixed height and cropping the excess, or attempting to calculate the height based on invoice items, which may not always yield accurate results.

I think it can be implemented in WeasyPrint but it will need changes in the python code.

Issue #193, opened 2014-05-13
Dynamic page height, fit to content


kujiu:
Add an option on page size (for example for height) to adapt the size of the page to fit to its content. The document will have a dynamic size and not fixed-size pages.

dmalan:
@SimonSapin, for a grading tool, we're actually hoping to do something similar so that each page in a PDF is only as tall as it needs to be to contain the page's content. Inspired by your comments and @sebasmagri's comments at https://github.com/Kozea/WeasyPrint/issues/274, we've figured out how to get the actual height of the content in the `body` of a page, as via `document.pages[0]._page_box.children[0].children[0].height`. But to keep everything on one page, is there perhaps a more elegant solution than

1. initially specifying a very large height and then cutting it down as needed (for which we'd need to guess a max possible height), or
1. letting the content span multiple pages initially, sum the pages' heights, and create a brand-new document whose `@page` height equals that sum (which feels a bit inefficient)?

Many thanks!

josesanch:
Hi, 

I've been working on this because I need to have dynamic height documents and I've found a workaround that works well for me.

Generating a 1cm height pages and then counting the result.

I hope this could help anybody.

Best Regards.

```

css_1cm = [CSS(string="@page {height: 1cm; page-break-inside: always;}}}")]
render = html.render(stylesheets=css + css_1cm)
css += [ CSS(string="@page {height: " + str(len(render.pages)) + "cm}}}")
html.write_pdf(stylesheets=css)



```

AndydeCleyre:
I'll add here my use case for which I think I want the same feature: I'm generating images of syntax-colored code, for viewing on screens. 

- I don't want the lines to wrap, though it's not so bad that they do.
- The generated image has a transparent border I don't want, but I trim that post-creation with graphicsmagick.
- The generated image is extra long when the code is short, but I trim a second time with the same graphicsmagick invocation.
- When the code more than 50 lines, I get a [transparent gap](https://user-images.githubusercontent.com/1787385/123166935-f009bf00-d443-11eb-9dae-56e7896c7a1d.png) between what I guess are pages, but I'm confused by the notion of "pages" in an image file altogether.

My current usage looks like this:

```python
from plumbum import local
from weasyprint import HTML
try:
    convert = local['gm']['convert']
except CommandNotFound:
    convert = local['convert']


def mk_png(html: str, folder=None) -> str:
    """Return generated PNG file path"""
[...]

PR #2250, opened 2024-09-08
Add `plotter` media type to print on a single page that fit automatically the height of content of the document

FrancoisCapon:
Hello,

First I know that this not in the W3C specification, but [I also think that printed on a single page with WeasyPrint would be very useful](https://github.com/Kozea/WeasyPrint/issues/2147).

So, this PR add a dedicated media type named `plotter` to print on a **single page that fit automatically the height of content of the document**. It use the `@media rollingpaper` for the page dimensions  (or default page dimensions if the media don't exist in css).

Using a dedicated media type avoid to add new option or paramater to Weasprint, so the addition does not modify existing APIs and will not be impacted by future developments.

* `weasyprint -s 01.css 01.html 01-print.pdf`
01-print pdf

* `weasyprint -s 01.css -m plotter 01.html 01-plotter.pdf`
01-plotter pdf

* `01.html`
```html
<body>
    <main>
        <div id="one">
            <h1>ONE</h1>
        </div>
        <div id="two">
            <h1>TWO</h1>
        </div>
        <div id="three">
            <h1>THREE</h1>
        </div>
    </main>
</body>
```
* `01.css`
```css
/* css media used with --media-type plotter */
@media rollingpaper {
    @page {
        size: 12cm;
        margin: 1cm;
        /* 120 x 400mm computed :-) */
    }
}
html, html * {
    margin: 0;
    padding: 0;
}
html,
body {
    background-color: lightgreen;
}
div {
    box-sizing: border-box;
    width: 10cm;
    height: 12cm;
    background-color: blueviolet;
    border: solid 0.2cm black;
    margin-bottom: 1cm;
}
div:last-child {
    margin-bottom: 0;
}
h1 {
    color: aliceblue;
    text-align: center;
    font-family: sans-serif;
    display: flex;
    justify-content: center;
    align-items: center;
    height: 100%;
}
```

[01-print.pdf](https://github.com/user-attachments/files/16922033/01-print.pdf)
[01-plotter.pdf](https://github.com/user-attachments/files/16922032/01-plotter.pdf)

### DIFF
--- a/weasyprint/document.py
+++ b/weasyprint/document.py
@@ -20,6 +20,8 @@ from .matrix import Matrix
 from .pdf import VARIANTS, generate_pdf
 from .text.fonts import FontConfiguration
 
+import xml.etree.ElementTree as ElementTree
+import time
 
 class Page:
     """Represents a single rendered page.
@@ -247,6 +249,9 @@ class Document:
 
     @classmethod
     def _render(cls, html, font_config, counter_style, options):
+        if html.media_type == 'plotter':
+            return Document._render_plotter(html, font_config, counter_style, options)
+        
         if font_config is None:
             font_config = FontConfiguration()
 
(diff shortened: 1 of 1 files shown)


liZe:
Hi!

Thanks a lot for your pull request. That would be a really useful feature, for sure, as discussed in #193 for example.

> First I know that this not in the W3C specification

Yes, and that’s the problem with this kind of features: without a specification, we don’t know what to do. There are many corner cases introduced by such a feature: what happens for absolutely positioned elements with `bottom` set, what happens with the page background’s position, what happens with blocks whose size depends on the page’s height?

That’s why we "just" want to implement a specification, and not write our own rules: specifying is a very hard work, and when it’s done in a bad way, users complain :smile:.

If we add such a feature, we’ll have to document and maintain it. It’s already hard to do this with a specification (ie. after smart people already spent countless months solving corner cases, and bikeshedding of course :smile:). We don’t want to do both the specification and implementation/documentation/maintenance work.

Hopefully, it doesn’t mean that it’s not possible to do this with WeasyPrint: you can actually use your code as a base to write a short script that generates a single page, without changing WeasyPrint’s code. You can hack it, WeasyPrint has been designed for it! You’ll be able to define what to do in undefined cases (or even ignore them), and adapt the rendering to what you want exactly. And other people will be able to adapt this code to fit their specific needs too. The difference is that you’ll be responsible for maintaining your script, instead of us being responsible for specifying rules and maintaining code to fulfill the different needs of a lot of users.

PR #2263, opened 2024-10-02
Build macOS executable for M- series Macs

JakenHerman:
I noticed there was a GitHub Action that generated a Windows (`.exe`) executable, but nothing for macOS. This pull request adds a GitHub Action to generate an executable binary for macOS.

## Testing
There is a test step in the GitHub Action, so no manually testing should be necessary, however, if you'd like you can download the artifact produced from the GitHub Action run. You'll need to run `chmod +x weasyprint-macos/dist/weasyprint` in order to be able to execute, but once you do, you can run `./weasyprint-macos/dist/weasyprint --info` to verify the executable is working.

## Use cases
This executable can be bundled with [Electron](https://www.electronjs.org) or [Tauri](https://tauri.app) applications so PDF generation is a possibility without requiring users of the built applications to install additional resources.

### DIFF
--- a/.github/workflows/m-series-mac-executable.yml
+++ b/.github/workflows/m-series-mac-executable.yml
@@ -1,0 +1,41 @@
+name: WeasyPrint’s M-series macOS binary generation
+on: [push]
+
+jobs:
+  generate:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - name: Install dependencies
+        run: |
+          brew install pango gobject-introspection cairo glib
+          brew install python@3.12
+          echo "export PATH=\"/opt/homebrew/opt/python@3.12/bin:$PATH\"" >> $GITHUB_ENV
+      - name: Ensure absolute imports
+        run: |
+          sed -i '' 's/^from \. /from weasyprint /' weasyprint/__main__.py
+          sed -i '' 's/^from \./from weasyprint\./' weasyprint/__main__.py
+      - name: Install requirements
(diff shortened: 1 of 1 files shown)


JakenHerman:
I'm not sure where the appropriate place to update documentation on using this binary would be, but I would be happy to contribute to the documentation as the team sees fit. You can see the [result of the GitHub action run on my fork here](https://github.com/JakenHerman/WeasyPrint/actions/runs/11134694316).

liZe:
Hi!

Thanks for this pull request.

For now, we’re not sure that we want to generate and support a macOS executable. We tried with Windows, and it actually introduced a number of bugs that are hard to understand, reproduce, and fix for us. The problem is that we don’t have regular Windows or macOS testers in the team, and that nobody wants to take care of this in the long term. So, if we have a large number of users ready to try, report and fix bugs, why not, but we can’t promise anything.

That being said, from a technical point of view, there’s a couple of things I don’t understand:

- Isn’t PyInstaller supposed to find libraries? All the `--add-binary` should be useless if we do set the `DYLD_FALLBACK_LIBRARY_PATH` environment variable when calling PyInstaller (see this line: https://github.com/Kozea/WeasyPrint/blob/db288a4e241afb7972e8cdbe2f04e967fb3ed5e9/.github/workflows/tests.yml#L44-L45). Even better: if we don’t use `setup-python` and rely on Homebrew’s Python instead, we shouldn’t need to set the environment variable at all.
- I don’t think that we should set `DYLD_LIBRARY_PATH` at runtime. PyInstaller should be able to put the libraries somewhere at runtime and to let Python find them automatically, with no runtime hook. That’s at least what it does for Windows, and I think that it was the case for macOS too last time we tried to generate a macOS executable.

PR #2271, opened 2024-10-07
Hyphenation: consider words that contain a NBSP

ondras:
The main goal here is to fix https://github.com/Kozea/WeasyPrint/issues/2270.

My change is simple:
  - adjust the `get_next_word_boundaries` function so it considers the whole *word-nbsp-word* cluster as a single potentially-hyphenable word,
  - ignore suggested hyphenation dictionary item(s) that end with a nbsp

Tests seem to pass. However:
xVyoSl

### DIFF
--- a/weasyprint/text/line_break.py
+++ b/weasyprint/text/line_break.py
@@ -10,6 +10,8 @@ from .ffi import FROM_UNITS, TO_UNITS, ffi, gobject, pango, pangoft2, unicode_to
 from .fonts import font_features, get_font_description
 
 
+NBSP = '\u00a0'
+
 def line_size(line, style):
     """Get logical width and height of the given ``line``.
 
@@ -400,7 +402,7 @@ def split_first_line(text, style, context, max_width, justification_spacing,
             dictionary = pyphen.Pyphen(lang=lang, left=left, right=right)
             context.dictionaries[dictionary_key] = dictionary
         dictionary_iterations = [
-            start for start, end in dictionary.iterate(next_word)]
+            start for start, end in dictionary.iterate(next_word) if start[-1] != NBSP]
     else:
         dictionary_iterations = []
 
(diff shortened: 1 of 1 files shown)


liZe:
Thanks a lot for the pull request!

I think your code helps to improve the result for some cases, but I’m not confident with giving "word-nbsp-word" to Pyphen. We used to (and still do?) have problems with punctuation marks and other non-letter characters, as Pyphen is only designed to handle clean words, and I’d like to avoid this kind of problems again. We’ll also get right or wrong results depending on the hyphenation methods used by different languages.

Moreover, I think that `get_next_word_boundaries` should really find word boundaries, even the ones that are not line break opportunities (like with no-break space, but there are [others](https://unicode-explorer.com/c/202F), using [`is_line_break`](https://docs.gtk.org/Pango/struct.LogAttr.html) may help).

The origin of the problem is that we assume that lines break at word boundaries, and that the next word is the only one we try to hyphenate:
https://github.com/Kozea/WeasyPrint/blob/becf6dcafe39a55a2d7caad48690a1003707251b/weasyprint/text/line_break.py#L362-L379

Instead of finding only the next word, we should find the next words until we reach a line-break opportunity, and try to hyphenate these words separately. It requires some work, but it seems to be possible.

liZe:
@ondras I’ve open #2416 with a more reliable way to find word boundaries for hyphenation. It should fix your problem, don’t hesitate to give feedback if you try it!

ondras:
Hi @liZe , thanks a lot! I tested your branch and it seems to work fine on my testcase.

One thing of interest, which is probably not related to the discussed change at all: WeasyPrint is now displaying contents of my `<style>` tags. I am absolutely sure this was not the case several months before -- and it definitely looks like a bug. Is this a known behavior, or shall I file an issue?

liZe:
> Hi @liZe , thanks a lot! I tested your branch and it seems to work fine on my testcase.

Good to know, thanks! We’ll merge it for version 66 then.

> One thing of interest, which is probably not related to the discussed change at all: WeasyPrint is now displaying contents of my `<style>` tags. I am absolutely sure this was not the case several months before -- and it definitely looks like a bug. Is this a known behavior, or shall I file an issue?

You probably didn’t update CSSSelect2, WeasyPrint now requires the latest version.

PR #2284, opened 2024-10-23
Fix duplicate instructions in installation docs

### LINKED ISSUE #2282: [docs] Identical Alpine install instructions for virtualenv using wheels and without wheels
Documentation guide for Apline installation contains identical instructions for virtualenv using wheels and without wheels.

https://github.com/Kozea/WeasyPrint/blob/1aae1452da384de0160f599d4e2500aa1f3bebfc/docs/first_steps.rst?plain=1#L80-L88

Either remove "branching" if there is indeed no difference meant, or replace instructions with unidentical commands.

[Doc Ref](https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#alpine-3-17)

MrLightful:
Fix #2282

### DIFF
--- a/docs/first_steps.rst
+++ b/docs/first_steps.rst
@@ -77,13 +77,7 @@ To install WeasyPrint using your distribution’s package::
 
   apk add weasyprint
 
-To install WeasyPrint inside a virtualenv using wheels (if possible), you need
-the following packages::
-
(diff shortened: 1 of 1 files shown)


liZe:
Thank you!

PR #2299, opened 2024-11-06
Update install instructions for alpine

jo-bitsch:
As per the discussion in #699, document an easier way to install WeasyPrint from PyPI when using prebuilt wheels.

To keep the document consistent, I copied the context descriptions over from the other Linux flavors.

### DIFF
--- a/docs/first_steps.rst
+++ b/docs/first_steps.rst
@@ -77,7 +77,13 @@ To install WeasyPrint using your distribution’s package::
 
   apk add weasyprint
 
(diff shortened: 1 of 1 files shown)


jo-bitsch:
I updated the install instructions for packages without wheels and also added `libwebp-dev`, so that the default sample command
```
weasyprint https://weasyprint.org/ /tmp/test.pdf
```
also works for the non-wheel build.

(The prebuilt wheels already include support for `webp`.)

Thanks for being so responsive about this PR!

PR #2314, opened 2024-11-26
Use CSS 'image-rendering' attribute for images in SVGs

H4kor:
Currently the `image-rendering` attribute is not passed to the draw function. 
In the case of `image-rendering: pixelated;` this leads to blurry images inside SVGs if the defined size is larger than the source image.


liZe:
Thank you!

liZe:
We’ll release 63.1 before merging your pull request, we’ll be back soon!

liZe:
Perfect, thank you!

PR #2321, opened 2024-12-07
Add example invocation of WeasyPrint on the "Contribute" page

alexandergitter:
As someone who doesn't usually work with Python, I always have to figure this out from scratch when I come back to WeasyPrint. So I figured I'd just add it to the documentation.


liZe:
Thank you!

PR #2329, opened 2024-12-16
Refactor text.line_break.get_log_attrs()

kesara:
`bytestring` returned from `get_log_attrs()` doesn't seemed to be used anywhere.


liZe:
Thank you!

PR #2343, opened 2025-01-09
Add support for outline-offset CSS property.

xavidotron:
outline-offset provides a convenient way to give a box two different borders with space between them, useful for common ways of doing slightly fancier boxes without needing to go to the trouble of an image border. This pull request adds support for outline-offset and basic unit test coverage.


liZe:
Thank you, that’s really cool to have great PRs like yours!

PR #2345, opened 2025-01-11
Handle small-caps synthesis

### LINKED ISSUE #123: Support font-variant: small-caps approximation
Hi,

It is written in the documentation that I should report as a bug if some CSS 2.1 feature is not working. I found that the `font-variant: small-caps` CSS is not working (it works in the browser, but not in the rendered PDF).

Hope this helps, regards, Sz'

liZe:
Fix #123.

Capture d’écran du 2025-01-11 23-09-32

Before merging this pull request, we should:
- [x] add tests for all values (I only tested small-caps manually),
- [x] check what happens with Pango < 1.50 (because of recently added [variants](https://docs.gtk.org/Pango/enum.Variant.html)),
- [x] check what happens with fonts with real small caps,
- [x] optimize (if needed) the new font size calculation,
- [x] update documentation.

### DIFF
--- a/docs/api_reference.rst
+++ b/docs/api_reference.rst
@@ -369,10 +369,6 @@ WeasyPrint does **not** support the ``@font-feature-values`` rule and the
 values of ``font-variant-alternates`` other than ``normal`` and
 ``historical-forms``.
 
-The ``font-variant-caps`` property is supported but needs the small-caps variant of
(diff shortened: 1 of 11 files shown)


liZe:
The result is cool with fonts that include small caps:
Capture d’écran du 2025-01-12 15-31-52

PR #2350, opened 2025-01-14
Use setLevel to restore logger level in capture_logs

noamkush:
This replaces the direct assignment of the logger level in `capture_logs`. This causes an issue due to internal caching in the logging module, which is not refreshed when assigning directly.


liZe:
Thank you!

Issue #2269, opened 2024-10-06
Non-breaking spaces have different size


ondras:
In this rendering: 
![obrazek](https://github.com/user-attachments/assets/1cc2b3a2-feb7-41d7-9c8b-c0a239c86ac9)

First two words (`o kliknutí`) have a NBSP between them (czech typography rules). However, IMHO this only forbirds line breaks; there is no reason for this space to have a different size than other inter-word spaces on the same line.

liZe:
You’re right. The current justification code adds extra space to space characters only. There are many places in the code when we assume that spaces are only "normal" spaces.

In this case, the problem is in:
https://github.com/Kozea/WeasyPrint/blob/1aae1452da384de0160f599d4e2500aa1f3bebfc/weasyprint/layout/inline.py#L1119-L1125

We count spaces and use it to set `justification_spacing`. We should change our space detection here:
https://github.com/Kozea/WeasyPrint/blob/1aae1452da384de0160f599d4e2500aa1f3bebfc/weasyprint/layout/inline.py#L1131

And fix our spacing adjustment here:
https://github.com/Kozea/WeasyPrint/blob/1aae1452da384de0160f599d4e2500aa1f3bebfc/weasyprint/text/line_break.py#L182-L196

For now we don’t have to support all [justification opportunities](https://www.w3.org/TR/css-text-3/#justification-opportunity) as we don’t support `text-justify`, but we can at least support [word separators](https://www.w3.org/TR/css-text-3/#word-separator). I’m not sure that a real list is actually defined in Unicode, as there are exceptions such as punctuation and fixed-width spaces. We can at least start with the list given by the specification.

liZe:
> I’m not sure that a real list is actually defined in Unicode, as there are exceptions such as punctuation and fixed-width spaces.

Pango provides [`is_expandable_space`](https://docs.gtk.org/Pango/struct.LogAttr.html), we have everything we need to support this correctly.

luca-vercelli:
Hello. I commited a fix for the issue. Should I write a test for it? If yes, how?
I did not use Pango's `is_expandable_space`, as it just searches for `\u0020` and `\u0040` .

liZe:
Hi @luca-vercelli,

Thank for your commit.

> I did not use Pango's `is_expandable_space`, as it just searches for `\u0020` and `\u0040` .

[...]

PR #2390, opened 2025-02-22
Fix justify non breaking spaces

luca-vercelli:
Fix issue #2269
I am not sure of results of test `test_text_align_justify_nbsp_2`. Is this the expected behaviour?


liZe:
Thank you.

> I am not sure of results of test `test_text_align_justify_nbsp_2`. Is this the expected behaviour?

Lines don’t break with no-break spaces, so that’s expected! I’ve merged the two tests in just one with both spaces and no-break spaces.

Issue #2253, opened 2024-09-16
li market position issue when a superscript is present.


kesara:
The marker position for `li` differs when `sup` (superscript) is present. There are no such issues with `sub` (subscript).

Example screenshots:
<img width="364" alt="Screenshot 2024-09-16 at 22 51 23" src="https://github.com/user-attachments/assets/6fb83de4-5dc3-42ff-ad4f-d144a06aa421">
<img width="358" alt="Screenshot 2024-09-16 at 22 51 31" src="https://github.com/user-attachments/assets/c4835a21-dca6-4390-aee8-dc759be76839">

Test HTML:
```
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8">
    <title>li marker position test</title>
  </head>
  <body>
    <h1><code>li</code> marker position test</h1>
    <h4>List <code>ul</code> 0</h4>
    <ul>
      <li>This list has an ordinary line.</li>
    </ul>
    <h4>List <code>ul</code> 1</h4>
    <ul>
      <li>This list has a <sup>superscript</sup>.</li>
    </ul>
    <h4>List <code>ul</code> 2</h4>
    <ul>
      <li>This list has a <sub>subscript</sub>.</li>
    </ul>
    <h4>List <code>ul</code> 3</h4>
    <ul>
      <li>This list has an ordinary line.</li>
      <li>This list has a <sup>superscript</sup>.</li>
      <li>This list has a <sub>subscript</sub>.</li>
      <li>This list has another ordinary line.</li>
    </ul>
    <h4>List <code>ol</code> 0</h4>
    <ol>
      <li>This list has an ordinary line.</li>
    </ol>
    <h4>List <code>ol</code> 1</h4>
    <ol>
      <li>This list has a <sup>superscript</sup>.</li>
    </ol>
    <h4>List <code>ol</code> 2</h4>
    <ol>
      <li>This list has a <sub>subscript</sub>.</li>
    </ol>
    <h4>List <code>ol</code> 3</h4>
    <ol>
      <li>This list has an ordinary line.</li>
      <li>This list has a <sup>superscript</sup>.</li>
      <li>This list has a <sub>subscript</sub>.</li>
      <li>This list has another ordinary line.</li>
    </ol>
  </body>
</html>
```

Sample PDF generated with WeasyPrint `62.3`: 
[li.pdf](https://github.com/user-attachments/files/17011714/li.pdf)

kesara:
[...]

PR #2393, opened 2025-02-25
list-style-position outside appearance

luca-vercelli:
fix issue #1557

### DIFF
--- a/weasyprint/formatting_structure/build.py
+++ b/weasyprint/formatting_structure/build.py
@@ -170,7 +170,8 @@ def element_to_box(element, style_for, get_image_from_uri, base_url,
         marker_boxes = list(marker_to_box(
             element, state, style, style_for, get_image_from_uri,
             target_collector, counter_style))
-        children.extend(marker_boxes)
+        if not style['list_style_position'] == 'outside':
+            children.extend(marker_boxes)
 
     children.extend(before_after_to_box(
         element, 'before', state, style_for, get_image_from_uri,
@@ -255,7 +256,12 @@ def element_to_box(element, style_for, get_image_from_uri, base_url,
         box.children.insert(0, marker)
 
     # Specific handling for the element. (eg. replaced element)
-    return html.handle_element(element, box, get_image_from_uri, base_url)
+    box_handled_list = html.handle_element(element, box, get_image_from_uri, base_url)
+
+    if marker_boxes and style['list_style_position'] == 'outside':
+       return list(marker_boxes) + list(box_handled_list)
+    else:
+        return box_handled_list
 
 
 def before_after_to_box(element, pseudo_type, state, style_for,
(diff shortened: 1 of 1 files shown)


liZe:
Thanks for your pull request!

I think that we can find a better solution. The current code is a dirty workaround (shame on me, I wrote it!), and your PR is a workaround on top of a workaround. 😄 We’ll get other bugs for sure if we keep the current code, even with your fix.

The [specification](https://drafts.csswg.org/css-lists-3/#list-style-position-outside) is unfortunately helpless:

> This is handwavey nonsense from CSS2, and needs a real definition.

So, this is not as easy as it looks, finally…

But we should at least follow what’s proposed:

> The marker box is a block container and is placed outside the principal block box.

Currently, it’s placed inside the list item, not outside, leading to the problem we’re trying to solve.

https://github.com/Kozea/WeasyPrint/blob/0998c31783808f76488630d02b353d0c54e63a45/weasyprint/formatting_structure/build.py#L169-L173

I tried to look at what browsers do. In Firefox, the position of the outline marker is set and modified in so many places: see each call to `getOutsideMarker` in [`nsBlockFrame.cpp`](https://searchfox.org/mozilla-central/source/layout/generic/nsBlockFrame.cpp).

Even if we don’t support all these cases (including vertical alignment #2253, floats, line breaks at the top of the list item…), keeping the absolute position, removing the translation and putting the marker outside the list item should be better than what we have, with less code. I’m not sure that it’s easy to do, because `element_to_box` is supposed to return only one box, but if you have other ideas don’t hesitate to share them!

(Or maybe we should implement [CSS Position](https://drafts.csswg.org/css-anchor-position-1/) and use it to position outside markers.)

luca-vercelli:
I moved the marker outside. This seems to work.

Could I add any test?

BTW, 
>  I’m not sure that it’s easy to do, because element_to_box is supposed to return only one box

this is not true, luckily `element_to_box `is designed to return a list of boxes (even if usually it returns 1).

A side note. For debug purposes I also modified `Box` class, I don't know if this can help somebody (I did not commit this):

```
class Box:
    ...
    
    def __str__(self):
        return f'<{type(self).__name__} {self.element_tag}>'

    def __repr__(self, tabs=0):
        s = ' ' * tabs + str(self)
        for c in self.all_children():
           s += '\n'
           s += c.__repr__(tabs + 1)
        return s
```

liZe:
> I moved the marker outside. This seems to work.

Cool!

> Could I add any test?

Looks like many tests fail now (I suppose because they rely on the previous layout), I’ll fix them and see if it’s easy to add a test!

liZe:
> Many tests failed before this fix, too!

You mean, after your first commit?

I tried to render:

```html
<ul>
  <li>a</li>
  <li>b</li>
  <li>c</li>
</ul>
```

It looks like there’s something wrong here.

luca-vercelli:
> You mean, after your first commit?
Yes, I see many tests failing in the main branch, don't you?

> It looks like there’s something wrong here.

You are right :( sorry don't waste your time, I will check.

liZe:
> Yes, I see many tests failing in the main branch, don't you?

No, tests pass for me, and for the [CI](https://github.com/Kozea/WeasyPrint/actions/workflows/tests.yml).

There are [a few requirements](https://doc.courtbouillon.org/weasyprint/stable/contribute.html#tests), but if it doesn’t work for you, maybe there’s another problem.

> You are right :( sorry don't waste your time, I will check.

OK :+1:

PR #2412, opened 2025-03-20
allow no output filename

vpoulailleau:
Allow to call weasyprint CLI without output filename. The default output filename is input filename followed by `.pdf`.

(Everytime I use weasyprint, I forget to provide output filename…)

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -355,6 +355,9 @@ def test_command_line_render(tmp_path):
         _run('combined.html out2.pdf')
         assert (tmp_path / 'out2.pdf').read_bytes() == pdf_bytes
 
+        _run('combined.html')
+        assert (tmp_path / 'combined.html.pdf').read_bytes() == pdf_bytes
+
         _run('combined-UTF-16BE.html out3.pdf --encoding UTF-16BE')
         assert (tmp_path / 'out3.pdf').read_bytes() == pdf_bytes
 
--- a/weasyprint/__main__.py
+++ b/weasyprint/__main__.py
@@ -67,7 +67,8 @@ PARSER = Parser(prog='weasyprint', description='Render web pages to PDF.')
 PARSER.add_argument(
     'input', help='URL or filename of the HTML input, or - for stdin')
 PARSER.add_argument(
-    'output', help='filename where output is written, or - for stdout')
+    'output', help='filename where output is written, or - for stdout',
+    nargs="?", default=None)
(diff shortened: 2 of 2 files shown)


liZe:
Hi,

Thanks for this pull request! 🙏🏽 

That’s a good idea, but I have some concerns about the current implementation:
- When `document.html` is given, I think I’d prefer to get `document.pdf`, not `document.html.pdf`.
- When a URL is given (for example `weasyprint https://weasyprint.org/`), we currently get a crash.

We can first decide which "good" filename we want to get for these common cases (and maybe look at what other widely used tools do):
- `document`
- `document.html`
- `../../folder/document.html`
- `https://example.org/`
- `https://example.org/index`
- `https://example.org/index.php`
- `https://example.org/index.php?a=b&c=d`
- … (you’ll for sure find other ones!)

And then, we’ll find a simple, clean and reliable way to get this! 😄

luca-vercelli:
Hello, who was that question for?
Here is my unrequested opinion:

|   source               |  target     |
|--------------------|-----------|
|document.html| ./document.pdf |
|document| ./document.pdf |
|document.foo| ./document.pdf |
|document.foo.bar| ./document.foo.pdf |
|../../folder/document.html| ./document.pdf  (*) |
|https://example.org/ | ./output.pdf (or some other default) |
|https://example.org/path/index| ./index.pdf |
|https://example.org/path/index.php| ./index.pdf |
|https://example.org/path/index.php?a=b&c=d| ./index.pdf |
|https://example.org/path/index#a=b;c=d| ./index.pdf |

(*) not sure of this, but I think that's a good choice about write permissions on filesystem

vpoulailleau:
@liZe I don't have the time right now, maybe in july. The suggestions of @luca-vercelli look good to me. What do you think?

liZe:
> @liZe I don't have the time right now, maybe in july. The suggestions of @luca-vercelli look good to me. What do you think?

@vpoulailleau Looks good to me. You can write a dedicated function for that, so that we can easily test it.

(I really hope that it won’t be too complex though. Comments like the ones in [this thread](https://stackoverflow.com/questions/295135/turn-a-string-into-a-valid-filename) may help to find corner cases.)

Firefly7-7:
about the case of `../../folder/document.html`, I agree with you, not necessarily because of permissions but simply because the default behavior most users expect (and is what most programs implement, I believe) is that if you run a process in a directory, the output will be in that directory.

PR #2430, opened 2025-04-12
WeasyPrint hierarchical documentation

jbaehr66:
This is a comprehensive set of tiered documentation for WeasyPrint I produced in order to analyse issues and enhancements for the CSS layout. 

In addition to the core documentation it provides failure analysis and a plan for making CSS enhancements

### DIFF
--- a/documentation/md/00_index.md
+++ b/documentation/md/00_index.md
@@ -1,0 +1,65 @@
+# WeasyPrint Documentation
+
+**Version:** 65.0
+
+This documentation provides comprehensive information about the WeasyPrint HTML-to-PDF conversion library.
+
+## Table of Contents
+
+1. [Overview](01_overview.md) - High-level introduction to WeasyPrint
+2. [API Reference](02_api_reference.md) - Detailed API documentation
+3. [CSS Support](03_css_support.md) - Supported CSS features and usage patterns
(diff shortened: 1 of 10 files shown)


jbaehr66:
Let me know if you want to incorporate this

liZe:
Hi,

First of all, we’re currently satisfied with our current documentation, even if it’s not perfect. We used to have a lot of reported issues in the past, we worked hard to find a way to better organize and improve the documentation, and the number of related issues has fallen drastically.

We have read the pull request, and to be honest, we can really feel the common flaws of AI-generated content:
- it **looks** good, but…
- there are many errors,
- there is *a lot* of well written content, but the really useful content is scarce,
- some content may look interesting at first sight, but is actually pretty stupid.

Do we really need a full Python script when we can use a browser or curl to check that a URL works? Do we really want to alter HTML and CSS using `replace` or [regular exceptions](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags) to "detect" endless loops or layout problems? No, we don’t. Actually, we don’t want to let users think that complex problems can be easily solved by copy-pasting Python snippets without understanding them, we want to help them learn and understand how things actually work.

And if some users don’t want to take the time to understand, we have no problem with that! They can already ask their favorite AI tool to to solve their problems, they don’t even need this documentation. 😄

PR #2445, opened 2025-04-28
Add functionality to have additional http headers for url_fetcher

dedalusMohantyMa:
Add a parameter `http_headers` to `default_url_fetcher` which allows developers to add additional HTTP headers, such as `Authorization` or custom headers, so that you do not have to copy the whole implementation.


liZe:
Hi,

Thanks for your pull request! That’s a good idea and that’s compatible with previous versions, let’s merge it!

PR #2466, opened 2025-05-28
add code to allow utils.transform to modify origin based on presence …

### LINKED ISSUE #2256: SVG transform-origin ignored
The attribute "transform-origin" sets the reference point used in transformations set by the "transform" attribute. In the examples below, a rotation was used with the origin set to the center of the triangle. In WeasyPrint, the origin is ignored and the triangle is rotated relative to (0, 0). When the object is far from (0, 0), it can rotate far from it's intended location and far outside the viewport.

The following SVG reproduces the problem. The rotation is small in order to make the error more visible.

```
<svg xmlns="http://www.w3.org/2000/svg"  viewBox="0 0 400 400">

<defs>
  <g id="triangle" transform="scale(0.6)">
    <path d="M 0,-20 l 23,40 l -46,0 Z" stroke="purple" fill="none" />
  </g>
</defs>


<use href="#triangle" x="10mm" y="10mm"/>
<use href="#triangle" x="10mm" y="10mm" transform-origin="10mm 10mm" transform="rotate(9)" />

</svg>
```

In Firefox, the triangle is rotated relative to the center of the triangle:

svgtransformorigin-firefox

But WeasyPrint rotates it relative to (0, 0). It this example it looks slightly offset because the rotation is small.

svgtransformorigin-weasyprint

Reference files:
[svgtransformorigin-firefox.pdf](https://github.com/user-attachments/files/17058980/svgtransformorigin-firefox.pdf)
[svgtransformorigin.pdf](https://github.com/user-attachments/files/17058981/svgtransformorigin.pdf)

siaslas98:
…of transform-origin property

Fix #2256.

### DIFF
--- a/tests/draw/svg/test_transform.py
+++ b/tests/draw/svg/test_transform.py
@@ -411,3 +411,28 @@ def test_transform_unknown(assert_pixels):
               stroke-width="2" stroke="red" fill="none" />
       </svg>
     ''')
+
+
+@assert_no_logs
+def test_transform_origin(assert_pixels):
+    assert_pixels('''
+        ___RRRRRR
+        ___RRRRRR
+        ___RR___R
+        ___RR___R
+        ___RRRRRR
+        ___RRRRRR
+        _________
+        _________
+        _________
+    ''', '''
+      <style>
+        @page { size: 9px }
+        svg { display: block }
+      </style>
+      <svg width="9px" height="9px" xmlns="http://www.w3.org/2000/svg">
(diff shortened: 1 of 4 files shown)


liZe:
> Ok, code is ready for review.

Thanks a lot! I’ll add a test and handle secondary cases before merging.

siaslas98:
No, thank you! I really appreciate your guidance throughout the process. This was my first time contributing to a GitHub project beyond personal work, and I learned a lot from the experience!

Issue #1836, opened 2023-03-12
PDF/UA page header and footer


jjsarton:
Page header and footer content are not marked as artifact and therefor spoken from a screen reader. This is wrong!

liZe:
Hi!

Could you please share an example, so that we can perfectly understand what you are talking about?

liZe:
If you’re interested in this issue, it’s time to test https://github.com/Kozea/WeasyPrint/pull/2471! Feedback will be highly appreciated, even to say that it just works. 🙏

oraslaci:
Hi!
Really awesome progress on PDF/UA, but unfortunately this issue still persists.
Acrobat reader reads aloud the contents of the footer first thing on every page.
@jjsarton is right, according to recommendations. headers and footers should be tagged as artifiacts, so they are avoided by screenreaders.
Have a good one,
Laszlo

liZe:
> according to recommendations. headers and footers should be tagged as artifiacts

That’s supposed to be the case, but it looks like it’s not. I’ll take a look at that.

liZe:
That’s a bit complex.

What we currently do is to mark the content of page margins, but we don’t include the marked content in the main content tree, so that it doesn’t appear in PAC for example.

What we’d like to do is to include the marked content of margins in Artifact nodes at the root of the main content tree. But this is only possible in PDF 2.0 (PDF/UA-2), not before.

What the PDF Association asks is to mark the content of page margins as Artifacts. It means that we can’t tag the inner content. Currently, there’s no way to do this easily in WeasyPrint.

@oraslaci In the current PDFs, is margin content really read by screenreaders? It’s present in the list of tagged content by Adobe (because the content is actually tagged), but if screenreaders use the main content tree (they probably do, to get the right order?), we could probably keep this feature for later.

nathalie-tate:
Hello, I just wanted to note that in my testing, the NVDA screenreader does in fact read page footers aloud on every page 

eta: it doesn't *always* do it. I suspect it may have something to do with the selected settings in adobe acrobat. sorry for any confusion.

liZe:
> Hello, I just wanted to note that in my testing, the NVDA screenreader does in fact read page footers aloud on every page
> 
> eta: it doesn't _always_ do it. I suspect it may have something to do with the selected settings in adobe acrobat. sorry for any confusion.

Hi @nathalie-tate 

Thanks for your report, that’s really useful to get feedback from people using screenreaders 🙏 

[...]

PR #2474, opened 2025-06-09
pdf/ua improvements

alvarogf97:


liZe:
@alvarogf97 Thanks a lot for the pull request. :heart:

Most of the work I did is to simplify and deduplicate the code, but the overall idea is great!

I fixed a couple of problems I had with PAC, everything seems to work well with VeraPDF.

The only feature I removed is the merge for Part tags. I understand the idea, that’s something that would be interesting, but it would be better with all tags, not only Parts. I think that we can keep the current behavior for the next release, see how it goes, and improve the support (with tag deduplication, useless Span removal…) in future versions when we have more feedback.

If everything’s OK for you, I’ll merge.

alvarogf97:
> @alvarogf97 Thanks a lot for the pull request. ❤️
> 
> Most of the work I did is to simplify and deduplicate the code, but the overall idea is great!
> 
> I fixed a couple of problems I had with PAC, everything seems to work well with VeraPDF.
> 
> The only feature I removed is the merge for Part tags. I understand the idea, that’s something that would be interesting, but it would be better with all tags, not only Parts. I think that we can keep the current behavior for the next release, see how it goes, and improve the support (with tag deduplication, useless Span removal…) in future versions when we have more feedback.
> 
> If everything’s OK for you, I’ll merge.

I test it again and found some errors:
- Links are no longer created with the OBJR requested in PDF structures 
image
- Footer is no longer tagged as we use bottom-left-corner and those tags defined in the documentation
- Links are no longer be iterable using tab (I think this issue is caused by the first problem)

liZe:
Thanks for your feedback.

> * Links are no longer created with the OBJR requested in PDF structures

I’m not sure about that. I’ve removed this line (if that’s what you’re talking about), and re-adding this line breaks the validation with PAC. I’ll try to find a solution.

https://github.com/Kozea/WeasyPrint/blob/4238f70adbb455fede03f4222076f0d8e8c1a049/weasyprint/pdf/tags.py#L255

> * Footer is no longer tagged as we use bottom-left-corner and those tags defined in the documentation

There’s a discussion about this topic in #1836, with a link explaining that elements in page margins should be marked as Artifacts. As Artifacts are not allowed in the tree in PDF 1.7 (they are in 2.0), elements are just tagged but not put in tree, I’ve added a comment to explain that:

https://github.com/Kozea/WeasyPrint/blob/8686f3cc4d1f642e04a0cf560a2e18d0b899b190/weasyprint/pdf/tags.py#L187-L189

liZe:
@alvarogf97 There’s something I don’t understand with the way links work. With the line re-added, the tree seems to be right, but PAC complains about an "inconsistent entry" for the link while checking the PDF syntax.

[pdfua.pdf](https://github.com/user-attachments/files/20714956/pdfua.pdf)

alvarogf97:
> @alvarogf97 There’s something I don’t understand with the way links work. With the line re-added, the tree seems to be right, but PAC complains about an "inconsistent entry" for the link while checking the PDF syntax.
> 
> [pdfua.pdf](https://github.com/user-attachments/files/20714956/pdfua.pdf)
> 

Links should contain both elements the text itself and the OBJR that refers the link. Maybe  there's something wrong with that reference? Checking the previous code It seems to have the same error. I'll try to find a solution and It looks like the error is in the OBJR reference

liZe:
> Links should contain both elements the text itself and the OBJR that refers the link. Maybe there's something wrong with that reference? Checking the previous code It seems to have the same error. I'll try to find a solution and It looks like the error is in the OBJR reference

I think that I’ve found the problem: Nums should contain Links, not OBJRs. 0d9c5b2 gives correct results for PAC and VeraPDF for my examples, and the tree seems to be what we want.

alvarogf97:
> > Links should contain both elements the text itself and the OBJR that refers the link. Maybe there's something wrong with that reference? Checking the previous code It seems to have the same error. I'll try to find a solution and It looks like the error is in the OBJR reference
> 
> I think that I’ve found the problem: Nums should contain Links, not OBJRs. [0d9c5b2](https://github.com/Kozea/WeasyPrint/commit/0d9c5b2d5dbcc71db2af2943fffe487f8adcac6c) gives correct results for PAC and VeraPDF for my examples, and the tree seems to be what we want.

Although the problem that you cannot tab in Adobe moving across links has not been resolved, they are now tagged ok. Maybe it's an Adobe problem? In chrome it works. Any ideas?

PR is ready to be merged, any other problem should be resolved in another one.

Thanks for your support @liZe 😃

liZe:
> Although the problem that you cannot tab in Adobe moving across links has not been resolved, they are now tagged ok. Maybe it's an Adobe problem? In chrome it works. Any ideas?

It could be useful to compare with a PDF/UA document where it works. As you mentioned, it works well in browsers, though.

PR #2475, opened 2025-06-09
Add support for 'lh' and 'rlh' units

randomstuff:


liZe:
Thanks a lot for the pull request, it looks nice!

The code seems to work well. Even if we could cache the result as we do for `ex` or `ch`, there’s no need to go that deep until we find a cool way to get the context. We also have the same problem for tests, I’ll add a commit with a comment for that.

PR #2491, opened 2025-07-03
Add shared cache for cli usage

mdziadkowiec:
Add `--shared-cache` parameter to preserve image cache across multiple PDF generations. Only works together with `--cache-folder`.

### Problem
When using WeasyPrint from the CLI with image caching enabled, the cache is automatically cleared after each PDF is created. This behavior forces images to be re-downloaded for every PDF generation, which is inefficient when creating multiple PDFs that share common images.

### Proposed Solution

Add a new `--shared-cache` CLI parameter that allows users to manually manage cache persistence. When this flag is used, the image cache will be preserved after PDF creation, enabling reuse across multiple PDF generations.

External system will be responsible for reseting/removing cached files.

###  Implementation Details

The new parameter is opt-in and does not affect existing behavior
When `--shared-cache` is not specified, the current cache-clearing behavior is maintained
This change only affects CLI usage; Python script usage remains unchanged
Users can manually manage cache lifecycle when needed

### Benefits

- Improved performance when generating multiple PDFs with shared images
- Reduced bandwidth usage and faster generation times
- Backward compatibility maintained
- Provides flexibility for different use cases

### Usage Example

```bash
# Generate multiple PDFs while preserving cache
weasyprint --shared-cache --cache-folder "/tmp/cache" input1.html output1.pdf
weasyprint --shared-cache --cache-folder "/tmp/cache" input2.html output2.pdf
weasyprint --shared-cache --cache-folder "/tmp/cache" input3.html output3.pdf
```
This enhancement provides better control over caching behavior while maintaining full backward compatibility with existing workflows.

### DIFF
--- a/weasyprint/__init__.py
+++ b/weasyprint/__init__.py
@@ -84,6 +84,7 @@ DEFAULT_OPTIONS = {
     'full_fonts': False,
     'hinting': False,
     'cache': None,
+    'shared_cache': False,
 }
 
 __all__ = [
--- a/weasyprint/__main__.py
+++ b/weasyprint/__main__.py
@@ -120,6 +120,9 @@ PARSER.add_argument(
     help='store cache on disk instead of memory, folder is '
     'created if needed and cleaned after the PDF is generated')
 PARSER.add_argument(
+    '--shared-cache', action='store_true',
+    help='keep the cache folder after PDF generation for sharing between multiple runs (only works with --cache-folder)')
+PARSER.add_argument(
     '-D', '--dpi', type=int,
     help='set maximum resolution of images embedded in the PDF')
 PARSER.add_argument(
@@ -172,6 +175,7 @@ def main(argv=None, stdout=None, stdin=None, HTML=HTML):  # noqa: N803
 
     options = {
         key: value for key, value in vars(args).items() if key in DEFAULT_OPTIONS}
+    options['shared_cache'] = args.shared_cache
 
     # Default to logging to stderr.
     if args.debug:
--- a/weasyprint/document.py
+++ b/weasyprint/document.py
@@ -153,11 +153,12 @@ class DiskCache:
 
     """
 
-    def __init__(self, folder):
+    def __init__(self, folder, shared_cache=False):
         self._path = Path(folder)
         self._path.mkdir(parents=True, exist_ok=True)
         self._memory_cache = {}
         self._disk_paths = set()
+        self._shared_cache = shared_cache
 
     def _path_from_key(self, key):
         digest = md5(key.encode(), usedforsecurity=False).hexdigest()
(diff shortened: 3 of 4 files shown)


liZe:
Hi!

Thanks for this pull request.

Providing an option to keep the cache after generating a document is a good idea! There are 2 topics I’d like to solve before merging the pull request:

1. We usually try to avoid extra options when it’s possible, as it can quickly become a mess. Maybe there’s another solution: if the cache folder exists when WeasyPrint is launched, we can use it and keep the data in it; when it doesn’t exist, we create it and remove it at the end (that’s what’s done currently). What do you think of this behavior?
2. Instead of changing the logic in `images.py`, I think that we can keep all the logic changes in `DiskCache` instead and don’t change `images.py` at all. When the cache folder exists, we could populate the cache with existing files when `DiskCache` is created. We’ll need to use coding (base64?) or keep the key (if the key is harmless as a filename) instead of hashing (currently md5) to create the filename, so that we can rebuild the cache keys from the filenames. Does it make sense for you?

liZe:
The more I think about this, the more I’m not sure about how the feature should work…

When multiple processes share a cache on disk naively, we can easily find cases of race conditions. For WeasyPrint, two concurrent renderings may want to access the same images at the same time, to write and/or to read them, leading to data corruption (at least). It won’t happen for simple cases, but problems will sure be reported when some people start to use it at larger scales.

To get something clean and solid, we should at least have atomic writes, or more generic concurrency management solutions (based on a database for example). Obviously, we don’t want to do this.

So, until you have a great solution (I’d love that!), I’m don’t think that we can achieve this in a way that’s maintainable in the future. :cry:

liZe:
After 2 weeks thinking about this, I think that the idea is good but that the implementation would be complex. We don’t want to maintain an implementation that either cleanly deals with file access concurrency (too complex for us) or doesn’t handle it at all (bugs will be reported for sure).

Cache by document works well and covers a part of the cache needs. For multi-document cache in a single process, we already propose [a solution in memory](https://doc.courtbouillon.org/weasyprint/stable/common_use_cases.html#cache-and-optimize-images). For a solution on disk, the best way is to follow the solution in memory with a dict-like class, as we do with `DiskCache`.

For multiprocess uses, there’s no other way than using a kind of semaphore. If we want to avoid dealing with atomic disk operations (I think we do!), the easiest way is probably SQLite, as that’s what some (all?) browsers use for cache. As SQLite is built in Python, it could be possible to write a "simple" implementation with a SQLite equivalent of `DiskCache`. It could even replace `DiskCache` if the performance is OK.

We’ll close this PR for now, but we’d be really happy to discuss this topic further and eventually find a maintainable solution. Thanks a lot for your work, don’t hesitate to add comments even if the PR is closed!

PR #2510, opened 2025-07-27
Use pattern matching instead of scan-time edits for @font-face rules

### LINKED ISSUE #1073: Incorrect use of local font instead of using remote font referenced by @font-face rule
I am trying to render a page including the following style rules:

```css
@font-face {
  font-family: 'Asap';
  font-style: normal;
  font-weight: 400;
  src: local('Asap Regular'), local('Asap-Regular'), url(https://fonts.gstatic.com/s/asap/v11/KFOoCniXp96ayzse5Q.ttf) format('truetype');
}
```
The PDF is rendered with an incorrect font because the font `Asap Regular` is missing on my system and WeasyPrint does not try to use the remote font reference.

After having done some digging on my Linux system, when _Fontconfig_ tries to match a local font with `postscriptname` and `fullname` properties (as done by WeasyPrint) and does not find it, it returns a matching pattern with the `postscriptname` property set to the best matching font (as expected) but the `fullname` property is still set to the same value than the requested font.

https://github.com/Kozea/WeasyPrint/blob/d6ef067deb3234c36a5e364fb20711705e820094/weasyprint/fonts.py#L327-L334

For instance, if _Fontconfig_ tries to match 'Asap Regular` as font and it is missing, it returns the following properties:

```
# Equivalent to WeasyPrint code
λ fc-match -v ":fullname=Asap Regular:postscriptname=Asap Regular"
Pattern has 36 elts (size 48)
        family: "Adobe Utopia"(s)
        style: "Regular"(s)
        stylelang: "en"(s) "en"(s)
        fullname: "Asap Regular"(s)
        slant: 0(i)(s)
        weight: 80(f)(s)
        width: 100(f)(s)
        size: 12(f)(s)
        pixelsize: 12(f)(s)
        foundry: "Adobe"(w)
        antialias: False(w)
        hintstyle: 1(i)(w)
        hinting: True(s)
        verticallayout: False(s)
        autohint: False(s)
        globaladvance: True(s)
        file: "/usr/share/fonts/75dpi/UTRG__12-ISO8859-1.pcf.gz"(w)
        index: 0(i)(w)
        outline: False(w)
        scalable: False(w)
        dpi: 75(f)(s)
        scale: 1(f)(s)
        charset:
        0000: 00000001 ffffffff ffffffff 7fffffff 00000000 ffffffff ffffffff ffffffff
(w)
        lang: aa|ay|bi|br|ch|da|de|en|es|eu|fj|fo|fur|fy|gd|gl|gv|ho|ia|id|ie|io|is|it|lb|mg|nb|nds|nl|nn|no|nr|oc|om|pt|rm|sma|smj|so|sq|ss|st|sv|sw|tl|ts|uz|vo|wa|xh|yap|zu|an|fil|ht|jv|kj|kwm|li|ms|ng|pap-an|pap-aw|rn|rw|sc|sg|sn|su|za(s)
        fontversion: 0(i)(s)
        fontformat: "PCF"(w)
        embeddedbitmap: True(s)
        decorative: False(s)
        namelang: "en"(s)
        prgname: "fc-match"(s)
        postscriptname: "Adobe-Utopia"(s)
        color: False(w)
        symbol: False(s)
        variable: False(s)
        pixelsizefixupfactor: 1(f)(w)
        scalingnotneeded: True(w)
```

The exact font is not found but WeasyPrint follows anyway the happy path and use the substituted local font instead of retrieving the remote font referenced in the `@font-face` rule. WeasyPrint thinks the exact font is found because the resulting `fullname` value is the same as the requested one.

I don't know if this is a bug from WeasyPrint, Fontconfig or my system's font configuration.

### LINKED ISSUE #2507: segfault with pango 1.56.4-1
**Synopsis**
Updated my Arch Linux system today and WeasyPrint CLI immediately started segfaulting producing PDFs on some projects but not others. No apparent correlation of properties among the projects that failed and those that didn't. Stacktrace pointed to Pango for the immediate fault. Downgrading Pango from 1.56.4-1 to 1.56.3-1 seems to have resolved the problem.

I don't necessarily believe this is particularly a WeasyPrint problem, but I couldn't get any other software to evidence such a crash. So this report is mostly to help others that might come across the same issue.

**Versions**
I'm using WeasyPrint from git commit aefe4af, v66.0, link installed to a pip virtual environment for each project.

Following the Pango downgrade `weasyprint --info` reports:
```
% python -m weasyprint --info
System: Linux
Machine: x86_64
Version: #1 SMP PREEMPT_DYNAMIC Thu, 17 Jul 2025 21:05:29 +0000
Release: 6.15.7-arch1-1

WeasyPrint version: 66.0
Python version: 3.13.5
Pydyf version: 0.11.0
Pango version: 15603
```

Prior to the Pango downgrade `weasyprint --info` reported:
```
% python -m weasyprint --info
System: Linux
Machine: x86_64
Version: #1 SMP PREEMPT_DYNAMIC Thu, 17 Jul 2025 21:05:29 +0000
Release: 6.15.7-arch1-1

WeasyPrint version: 66.0
Python version: 3.13.5
Pydyf version: 0.11.0
Pango version: 15604
```

**Crash**
No Python error was reported, simply a segfault.

Running WeasyPrint in verbose mode indicated the crash occurring after beginning "Creating layout - Page 1".

The top of the coredump stacktraces look like this:
```
                #0  0x00007f16dcac5934 n/a (libpangoft2-1.0.so.0 + 0xb934)
                #1  0x00007f16dcac6eb6 n/a (libpangoft2-1.0.so.0 + 0xceb6)
                #2  0x00007f16dcaca4bf n/a (libpangoft2-1.0.so.0 + 0x104bf)
                #3  0x00007f16dcacaa8d n/a (libpangoft2-1.0.so.0 + 0x10a8d)
                #4  0x00007f16dd1717fc n/a (libpango-1.0.so.0 + 0x247fc)
                #5  0x00007f16dd175345 n/a (libpango-1.0.so.0 + 0x28345)
                #6  0x00007f16dd17fc9a n/a (libpango-1.0.so.0 + 0x32c9a)
                #7  0x00007f16dd183915 pango_layout_get_line_readonly (libpango-1.0.so.0 + 0x36915)
                #8  0x00007f16dd778052 ffi_call_unix64 (_cffi_backend.cpython-313-x86_64-linux-gnu.so + 0x26052)
                #9  0x00007f16dd7763dc ffi_call_int (_cffi_backend.cpython-313-x86_64-linux-gnu.so + 0x243dc)
                #10 0x00007f16dd7743c7 cdata_call (_cffi_backend.cpython-313-x86_64-linux-gnu.so + 0x223c7)
                #11 0x00007f16def60e3b _PyObject_MakeTpCall (libpython3.13.so.1.0 + 0x160e3b)
                #12 0x00007f16def72400 _PyEval_EvalFrameDefault (libpython3.13.so.1.0 + 0x172400)
```

liZe:
The logic is equivalent, but Pango seems to like this solution better. It also ensures that the font is used instead of using a system font with the same name.

There are 2 problems left with this solution:
- we don’t change the family name of the font, so the font in the PDF has the original name of the font (that’s why one test fails),
- we don’t mask the original name of the font for pattern matching, so it’s still possible to reference an `@font-face` font with `font-family: Original Font Family` (that is the case before this PR too).

These two problems could go away with extra FontConfig configuration lines, but they currently make Pango crash.

It’s just a draft for now, not ready for production. But feedback would be really appreciated for anyone who uses `@font-face` rules.

Fix #1073, fix #2507.

### DIFF
--- a/weasyprint/text/fonts.py
+++ b/weasyprint/text/fonts.py
@@ -188,23 +188,25 @@ class FontConfiguration:
             # Create Fontconfig XML config file.
             mode = 'assign_replace'
             root = Element('fontconfig')
-            match = SubElement(root, 'match', target='scan')
-            test = SubElement(match, 'test', name='file', compare='eq')
-            SubElement(test, 'string').text = str(font_path)
-            edit = SubElement(match, 'edit', name='family', mode=mode)
-            SubElement(edit, 'string').text = rule_descriptors['font_family']
+
+            match = SubElement(root, 'match', target='pattern')
+            test = SubElement(match, 'test', name='family', compare='eq')
+            SubElement(test, 'string').text = rule_descriptors['font_family']
             if 'font_style' in rule_descriptors:
-                edit = SubElement(match, 'edit', name='slant', mode=mode)
+                test = SubElement(match, 'test', name='slant', compare='eq')
                 text = FONTCONFIG_STYLE[rule_descriptors['font_style']]
-                SubElement(edit, 'const').text = text
+                SubElement(test, 'const').text = text
             if 'font_weight' in rule_descriptors:
-                edit = SubElement(match, 'edit', name='weight', mode=mode)
+                test = SubElement(match, 'test', name='weight', compare='eq')
                 integer = FONTCONFIG_WEIGHT[rule_descriptors['font_weight']]
-                SubElement(edit, 'int').text = str(integer)
+                SubElement(test, 'int').text = str(integer)
             if 'font_stretch' in rule_descriptors:
-                edit = SubElement(match, 'edit', name='width', mode=mode)
+                test = SubElement(match, 'test', name='width', compare='eq')
                 text = FONTCONFIG_STRETCH[rule_descriptors['font_stretch']]
-                SubElement(edit, 'const').text = text
+                SubElement(test, 'const').text = text
+            edit = SubElement(match, 'edit', name='file', mode=mode)
+            SubElement(edit, 'string').text = str(font_path)
+
             match = SubElement(root, 'match', target='font')
             test = SubElement(match, 'test', name='file', compare='eq')
             SubElement(test, 'string').text = str(font_path)
@@ -225,6 +227,7 @@ class FontConfiguration:
                     range_ = SubElement(charset, 'range')
                     for value in (unicode_range.start, unicode_range.end):
                         SubElement(range_, 'int').text = f'0x{value:x}'
+
             header = (
                 b'<?xml version="1.0"?>',
                 b'<!DOCTYPE fontconfig SYSTEM "urn:fontconfig:fonts.dtd">')


liZe:
Pattern matches brought other problems: as some fonts are aliased system wide, they are always included in the list of matching fonts, causing problems when using fonts both installed on the system and in `@font-face` rules.

Let’s use a more simple solution that avoids the crash. With older versions of Pango, the same problem caused various font mismatches: wrong weight, wrong style, wrong glyphes… This bug seems to be fixed too. And #1073 is now OK too (it probably already was before the PR).

PR #2523, opened 2025-08-11
Display tofu for missing glyphs

liZe:
Fix unicode-org/core-spec#166.

### DIFF
--- a/docs/first_steps.rst
+++ b/docs/first_steps.rst
@@ -18,7 +18,7 @@ WeasyPrint |version| depends on:
 * cssselect2_ ≥ 0.8.0
 * Pyphen_ ≥ 0.9.1
 * Pillow_ ≥ 9.1.0
-* fontTools_ ≥ 4.52.0
+* fontTools_ ≥ 4.59.2
 
 .. _Python: https://www.python.org/
 .. _Pango: https://pango.gnome.org/
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ dependencies = [
   'cssselect2 >=0.8.0',
   'Pyphen >=0.9.1',
   'Pillow >=9.1.0',
-  'fonttools[woff] >=4.52.0',
+  'fonttools[woff] >=4.59.2',
 ]
 classifiers = [
   'Development Status :: 5 - Production/Stable',
--- a/weasyprint/pdf/fonts.py
+++ b/weasyprint/pdf/fonts.py
@@ -150,17 +150,9 @@ class Font:
                 self.variations['slnt'] = slnt
             if 'ital' in axes and 'ital' not in self.variations:
                 self.variations['ital'] = int(self.style == 2)
-            # TODO: can use static=True when fontTools 4.60 is out.
-            for axis in axes:
-                if axis not in self.variations:
-                    self.variations[axis] = None
             partial_font = io.BytesIO()
             try:
-                ttfont = instantiateVariableFont(ttfont, self.variations)
-                # TODO: see fonttools#3918, can be removed when 4.60 is out.
-                for key, (advance, bearing) in ttfont['hmtx'].metrics.items():
-                    if advance < 0:
-                        ttfont['hmtx'].metrics[key] = (0, bearing)
+                ttfont = instantiateVariableFont(ttfont, self.variations, static=True)
                 ttfont.save(partial_font)
             except Exception as exception:
                 LOGGER.warning(f'Unable to instantiate "{self.family}" variable font')


weasyprint/draw/text.py
```diff
@@ -176,10 +177,21 @@ def draw_first_line(stream, textbox, text_overflow, block_ellipsis, matrix):
...
+                # Display empty glyph, such as space.
                 string += f'>{-width / font_size}<'
                 continue
+            elif glyph & pango.PANGO_GLYPH_UNKNOWN_FLAG:
+                # Display tofu for missing glyph.
+                string += f'{0:02x}' if font.bitmap else f'{0:04x}'
+                string += f'>{-width / font_size}<'
+                utf8_slice = slice(*sorted(utf8_positions[i:i+2]))
+                characters = utf8_text[utf8_slice].decode()
+                codepoints = ', '.join(str(ord(character)) for character in characters)
```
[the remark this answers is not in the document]
liZe: Are you some kind of Unicode geek? :rofl: 

Of course you’re right, thanks for the idea, it’s fixed.

liZe:
> The [Windows .exe](https://github.com/Kozea/WeasyPrint/actions/runs/16880677100) may be the most straightforward one to test?

Yes. You can also install it using `pip install --force git+https://github.com/Kozea/WeasyPrint.git@tofu`, or replace `weasyprint` by `git+https://…` in the requirements file.

weasyprint/draw/text.py
```diff
@@ -176,10 +177,21 @@ def draw_first_line(stream, textbox, text_overflow, block_ellipsis, matrix):
...
+                # Display empty glyph, such as space.
                 string += f'>{-width / font_size}<'
                 continue
+            elif glyph & pango.PANGO_GLYPH_UNKNOWN_FLAG:
+                # Display tofu for missing glyph.
+                string += f'{0:02x}' if font.bitmap else f'{0:04x}'
+                string += f'>{-width / font_size}<'
+                utf8_slice = slice(*sorted(utf8_positions[i:i+2]))
+                characters = utf8_text[utf8_slice].decode()
+                codepoints = ', '.join(f'U+{ord(char)}:04x' for char in characters)
```
lianghai, line 190: The format is wrong: https://github.com/unicode-org/core-spec/actions/runs/16891391687/job/47851784654#step:13:29

The mistakes like “U+3164:04x” should be “U+0C5C” (not “U+0c5c” either).

```suggestion
                codepoints = ', '.join(f'U+{ord(char):04X}' for char in characters)
```

lianghai:
I’m able to see the new behavior [in the logging](https://github.com/unicode-org/core-spec/actions/runs/16891391687/job/47851784654#step:13:29) now (a suggestion has been made above for the formatting mistake), so this already addresses the concern of not being warned about unsupported characters. I’ll start using this branch in production.

However, indeed the rendering of .notdef in PDF is weird (selected part is copiable as U+100000 in Acrobat):

Screenshot 2025-08-11 at 22 45 53

Neither the first nor the last font (in the [font-family fallback list](https://github.com/unicode-org/core-spec/blob/99181f4802dab09dccf71b740b24df4182e3529f/src/pages/pdf/index.astro#L77))’s .notdef is used:

Screenshot 2025-08-11 at 22 48 15

For your reference, Safari and Chrome both use the first font’s .notdef glyph, which makes sense to me too, as it’s usually unpredictable which font is the last tried one.

Here’s the control group for your reference, when the special font [Adobe NotDef](https://github.com/adobe-fonts/adobe-notdef) is used as the last font in the font-family fallback list, therefore proper .notdef glyphs are displayed but technically these aren’t PANGO_GLYPH_UNKNOWN_FLAG event:

Screenshot 2025-08-11 at 22 46 07

weasyprint/draw/text.py
liZe: Sorry, my mistake.

liZe:
It looks like there’s something wrong in both the .notdef glyph and the spacing.

> selected part is copiable as U+100000 in Acrobat

That’s a proof that our strategy is wrong. Using the same CID 0 for all unknown glyphs doesn’t give the possibility to map it to the original Unicode codepoint, so we’ll lose real original text.

What we want instead is to include different CIDs in text that are not in the font, map them to the Unicode codepoints, and use .notedef as a fallback for the display.

I’ll go back to the PDF spec to understand how to do that.

lianghai:
If PDF’s architecture makes it difficult to make the original code points copiable, I feel it’s reasonable to just make sure the .notdef glyphs don’t have any copiable text. It’s more important to draw the .notdef glyphs in a normal way (in addition to the logging warnings).

lianghai:
What if you don’t treat PANGO_GLYPH_UNKNOWN_FLAG as a special condition for drawing, but just issue a warning and let the rest of `string` assembling logic finish? Even for the space branch I’m not sure if stuff like x_offset should be skipped, because Latin letters are regularly kerned against space.

liZe:
> What if you don’t treat PANGO_GLYPH_UNKNOWN_FLAG as a special condition for drawing, but just issue a warning and let the rest of `string` assembling logic finish?

That’s because the normal logic requires the glyph size (for kerning, and to store it in a table required for the PDF), that doesn’t exist when there’s no glyph. I have found a solution for that part, we now get the right width and copiable text, the only "detail" missing is the .notdef glyph.

lianghai:
Following https://github.com/Kozea/WeasyPrint/blob/main/docs/contribute.rst I managed to run this branch locally.

By simply running `weasyprint http://localhost:4321/pdf/ test.pdf`, I get the following PDF rendering **without** “missing glyph” warnings:

Screenshot 2025-08-13 at 13 36 21

This is because macOS’s system font .LastResort, just like Adobe NotDef, is a catch-all font.

Therefore I make sure system fonts aren’t used with `FONTCONFIG_FILE="" weasyprint http://localhost:4321/pdf/ test.pdf`, and get this result:

Screenshot 2025-08-13 at 13 36 01

This is the problem I keep seeing in PDFs generated from GitHub workflow, as in the .notdef glyph is not rendered.

I’ve started to feel there’s probably some misunderstanding about .notdef? See https://learn.microsoft.com/en-us/typography/opentype/spec/recom#glyph-0-the-notdef-glyph about how .notdef exists in every font, and normally has a width and outlines. The glyph .notdef is always GID 0 and you can always refer to it when you need to draw an unsupported character.

Therefore I did this hack and the result (below) finally looks like expected, as when a character is not covered by any fonts (which causes `PANGO_GLYPH_UNKNOWN_FLAG`), a text rendering environment should simply draw .notdef using the primary font (not one of the fallbacks):

Screenshot 2025-08-13 at 13 43 49

Screenshot 2025-08-13 at 13 43 01

I don’t really understand why this works. What the term “glyph” refers to varies a lot and is very confusing in these various contexts (Pango, PDF…). There’s a lot of misunderstandings in the industry about how text encoding and fonts work, largely because some of these architectures were designed with older technologies in mind and don’t use today’s conventional terminology.

Because in my hack `glyph` is set to 0 for `PANGO_GLYPH_UNKNOWN_FLAG`, it doesn’t make sense to add info to `font.cmap` using `glyph` as key anymore:

Screenshot 2025-08-13 at 13 52 01

I haven’t figured out the actual data type of `font` and `cmap` though. The term “cmap” typically means a (Unicode) code point to glyph ID mapping, but here it seems to be reversed, and also the values are strings instead of single code points?

liZe:
> Following https://github.com/Kozea/WeasyPrint/blob/main/docs/contribute.rst I managed to run this branch locally.

Nice!

> This is because macOS’s system font .LastResort, just like Adobe NotDef, is a catch-all font.

Then it’s normal, we use Fontconfig’s fallback system that finds a font that includes a glyph for a given Unicode codepoint, even if it’s not in the list of font families in the CSS. That’s mandatory for emojis for example, you don’t have to put an emoji font in the list to get colored emojis rendered.

> This is the problem I keep seeing in PDFs generated from GitHub workflow, as in the .notdef glyph is not rendered.

Yes, this is not fixed yet. My goal is to find a way to display .notdef instead of the space, while keeping the original text in the PDF.

> .notdef exists in every font, and normally has a width and outlines. The glyph .notdef is always GID 0 and you can always refer to it when you need to draw an unsupported character.

That’s what I had in mind too after reading the PDF spec.

> Therefore I did this hack and the result (below) finally looks like expected

That’s what I did at first. The problem with this solution is that we don’t have the "right text" in the PDF anymore, and lose the ability to copy/paste or search. Text in PDF contains glyphs, and these glyphs are mapped to Unicode codepoints so that we have the original text to search or copy/paste. If we use .notdef in the text, we can’t map it to codepoints because .notdef can be used to represent different missing glyphs corresponding to different codepoints.

Hopefully, PDF provides a fallback system that can display a specific glyph (by default .notdef) when a glyph in the PDF "text" is missing from the embedded font. This fallback mechanism doesn’t work for us, and that’s what I’m currently trying to understand. It’s in chapter 9.7.6.3 of the PDF 2.0 specification.

> What the term “glyph” refers to varies a lot and is very confusing in these various contexts

Yes, that’s not the best variable name. Here, line 186, it’s the glyph ID in the font. That’s both the number we put in the PDF "text" and the index in the font that will be embedded.

> The term “cmap” typically means a (Unicode) code point to glyph ID mapping, but here it seems to be reversed, and also the values are strings instead of single code points?

Here, `font.cmap` is the mapping used to map glyph IDs (an integer) to a list of Unicode codepoints (a Python string, which is basically a list of Unicode codepoints). The value is a list because a glyph can represent multiple codepoints, for custom ligatures for example. It’s called the ToUnicode CMap in the PDF specification, explained in chapter 9.10.3.

It would be really useful to change some variable names in this code, I’ll do that too.

lianghai:
> Yes, this is not fixed yet. My goal is to find a way to display .notdef instead of the space, while keeping the original text in the PDF.

> That’s what I did at first. The problem with this solution is that we don’t have the "right text" in the PDF anymore, and lose the ability to copy/paste or search. Text in PDF contains glyphs, and these glyphs are mapped to Unicode codepoints so that we have the original text to search or copy/paste. If we use .notdef in the text, we can’t map it to codepoints because .notdef can be used to represent different missing glyphs corresponding to different codepoints.

If you want both the .notdef outlines and the original text, and you rely on the glyph to text mapping for text extraction, you need to duplicate the .notdef glyph (instead of only drawing a whitespace) for each different combination of font + text.

Or, I feel not being able to copy text from .notdef is already below the existing noise of not being able to copy text reliably from PDF.

> Here, font.cmap is the mapping used to map glyph IDs (an integer) to a list of Unicode codepoints (a Python string, which is basically a list of Unicode codepoints). The value is a list because a glyph can represent multiple codepoints, for custom ligatures for example. It’s called the ToUnicode CMap in the PDF specification, explained in chapter 9.10.3.

I see. Then similar to what may need to be done for .notdef, the condition below should also check the text and duplicate the glyph when necessary, otherwise you’re not keeping the original text correctly when different texts are shaped into the same glyph:

https://github.com/Kozea/WeasyPrint/blob/89f859b0f803b972382a7072a54711a4ca7fe692/weasyprint/draw/text.py#L231-L233

liZe:
> If you want both the .notdef outlines and the original text, and you rely on the glyph to text mapping for text extraction, you need to duplicate the .notdef glyph (instead of only drawing a whitespace) for each different combination of font + text.

I really think that the fallback systems that PDF provides can work without duplicating the glyph in the font.

> Then similar to what may need to be done for .notdef, the condition below should also check the text and duplicate the glyph when necessary, otherwise you’re not keeping the original text correctly when different texts are shaped into the same glyph:

That’s another feature that’s not supported yet, but nobody ever reported such a case.

liZe:
I’ve updated the branch to use the fallback system. It works with Poppler, Ghostscript and browsers, I hope that it works too with Acrobat and Apple Preview.

The code is not perfect though, I added 2 TODOs related to the glyph id we use for missing glyphs (that may actually be used in the font already) and .notdef width (which is set to a wrong value instead of being corrected by kerning). For the second point, we should also find where the width given by Pango for the missing glyph comes from.

lianghai:
In the PDF generated by https://github.com/unicode-org/core-spec/actions/runs/16994818236, there are [four unsupported characters](https://github.com/unicode-org/core-spec/blob/1dcfdf579614bfe87eed6e87275729cbb89a7946/src/pages/pdf/index.astro#L34) that should be shown as four gaps between five pipes, but the last character isn’t rendered properly:

Screenshot 2025-08-17 at 03 18 23

Is that because it’s beyond U+FFFF?

liZe:
> Is that because it’s beyond U+FFFF?

There’s a bug when glyphs go beyond U+FFFF, but we should get .notdef chars drawn. Here’s what I get with only this paragraph, using the font from the repository:

Capture d’écran du 2025-08-17 11-40-57

This bug happen because when Pango can’t find a glyph, it uses the Unicode codepoint + `pango.PANGO_GLYPH_UNKNOWN_FLAG`. What we do now to get a fake glyph id is to use the codepoint, which is just a dirty workaround, as it may conflict with other used glyph ids (there’s a TODO for that). With a 0xffff modulo I get this:

Capture d’écran du 2025-08-17 11-56-05

I don’t know why you don’t have the .notdef chars. Could you please try with the modulo? (I’ve pushed a commit.)

If it doesn’t work, I’ll try with the whole document on my computer.

When we get a working proof of concept, I’ll find a clean fix for the spacing problem (find where this default spacing from Pango comes from) and for the fake glyph id. (And change some variable names.)

(As a side note: WeasyPrint doesn’t support glyph ids beyond U+FFFF, because they’re not supported by PDF out of the box and require a specific mapping with what PDF calls Type 0 fonts. We decided to use Type 0 fonts everywhere in WeasyPrint, so that we’ll be able to handle glyph ids beyond U+FFFF more easily in the future, but that’s not done yet. As far as I can remember, the limit of 65536 glyphs is/was in TrueType and OpenType too, so we never had the problem in real life. My limited knowledge on the topic is quite old, do you know if we still have this limitation now, and if fonts without this limitation exist and are widely used?)

lianghai:
> This bug happen because when Pango can’t find a glyph, it uses the Unicode codepoint + pango.PANGO_GLYPH_UNKNOWN_FLAG.

I see. Now I understand this mixed architecture of GID and Unicode code point better.

> I don’t know why you don’t have the .notdef chars. Could you please try with the modulo? (I’ve pushed a commit.)

I don’t see a new commit yet, but sounds like `% 0xffff` defeats the whole idea of preserving the original code point.

> When we get a working proof of concept, I’ll find a clean fix for the spacing problem (find where this default spacing from Pango comes from) and for the fake glyph id. (And change some variable names.)

Uh probably just stick to the actual GID 0 .notdef of the original font.

> There’s a bug when glyphs go beyond U+FFFF …
> WeasyPrint doesn’t support glyph ids beyond U+FFFF …

Uh, judging from your terminology, you may be having a deep misunderstanding about Unicode code points and GIDs. I was talking about Unicode code points beyond U+FFFF, not glyph IDs beyond 0xFFFF:

- The U+XXXX notation is a convention of the Unicode Standard, and it specifically refers to a Unicode code point (not any text encoding unit), which has a valid range from U+0000 to U+10FFFF (inclusive). Unicode code points are the keys in the cmap table.
- Glyph IDs, on the other hand, have nothing to do with text encoding (besides being the values in the cmap table). In the current OpenType format the upper limit is 0xFFFE, because the number of glyphs is stored as an uint16 in the [maxp](https://github.com/harfbuzz/boring-expansion-spec/blob/main/beyond-64k.md) table.

I’m not familiar with the font architecture in PDF. But sounds like you’ve been confused by how Pango’s API piggybacks its usual GID with “Unicode codepoint + pango.PANGO_GLYPH_UNKNOWN_FLAG”, and that caused you to assume you can solve these pseudo GIDs like how actual GIDs work. These pseudo GIDs (Unicode code point + PANGO_GLYPH_UNKNOWN_FLAG) have nothing to do with a font, and it sounds like it’s your responsibility to decide how to render them.

> As far as I can remember, the limit of 65536 glyphs is/was in TrueType and OpenType too, so we never had the problem in real life. My limited knowledge on the topic is quite old, do you know if we still have this limitation now, and if fonts without this limitation exist and are widely used?

Handing a Unicode code point beyond U+FFFF is entirely orthogonal to handling a font with a glyph count higher than 0xFFFF. In my last comment I was asking if the code point U+11B60 (which is beyond U+FFFF) was the problem (eg, a font can simply have two glyphs: GID 0 .notdef, and GID 1 that is cmap-ed to some code point beyond U+FFFF), but now I see the problem is how the code points and GIDs are mixed up in this rendering architecture.

I know you’re trying very hard to both draw the .notdef glyphs and preserve the unsupported code points. But now after reading my clarification, is it possible that you’ve underestimated the challenge?

lianghai:
After https://github.com/Kozea/WeasyPrint/pull/2523/commits/06e5eb0378542d52f9d47afb4cfa21d56dce8c1b, the U+11B60 character doesn’t explode in PDF anymore:
 
Screenshot 2025-08-17 at 18 56 13

And in Acrobat I can copy U+11B60 from that last whitespace. Now I see the GID modified by `%= 0xffff` isn’t critical to preserving the text (which is the ToUnicode CMap’s responsibility), but then why do you have to use some integer based on the code point returned by Pango as the GID?

liZe:
> I don’t see a new commit yet, but sounds like `% 0xffff` defeats the whole idea of preserving the original code point.

I forgot to push, sorry.

We don’t want to preserve the original code point, we just want to have one unused glyph id for each missing original codepoint. The modulo is just a workaround as a proof of concept, until I write better but more complex code.

> Uh probably just stick to the actual GID 0 .notdef of the original font.

I can’t. Pango puts this space for us during the layout, not .notedef, and there’s nothing we can do about this as far as I know. If I put a .notedef char in the PDF string, I’ll still get the spacing problem (and lose search on top of that). If I use the .notdef width in the PDF, I break the layout.

> Uh, judging from your terminology, you may be having a deep misunderstanding about Unicode code points and GIDs. I was talking about Unicode code points beyond U+FFFF, not glyph IDs beyond 0xFFFF

I think we fully agree on that, and everything you wrote below.

> Handing a Unicode code point beyond U+FFFF is entirely orthogonal to handling a font with a glyph count higher than 0xFFFF.

Of course.

WeasyPrint has no problem with Unicode codepoints above U+FFFF (when the corresponding glyph(s) is/are in the font) because the Unicode characters are stored as UTF-16 in the ToUnicode table.

> In my last comment I was asking if the code point U+11B60 (which is beyond U+FFFF) was the problem

I understood that too, maybe my answer was not clear.

- PDF can embed fonts with glyph ids beyond 0xFFFF using a special font format called Type 0 (composite fonts).
- WeasyPrint uses Type 0 fonts, but doesn’t support glyph ids beyond 0xFFFF.
- When a glyph is missing, Pango returns a fake glyph id that is Unicode codepoint + `PANGO_GLYPH_UNKNOWN_FLAG`. That’s not a real glyph id because it’s not in the font, but it’s stored in the field used by Pango to usually store glyph ids.
- In the PDF, we use the Unicode codepoint (Pango fake glyph id - `PANGO_GLYPH_UNKNOWN_FLAG`) as a glyph id that is hopefully missing in the font. That’s not reliable, that’s temporary, until I write the code to get a glyph id that’s really missing in the font.
- For Unicode codepoint U+11B60, as there’s no glyph in the font to represent it, we were using U+11B60 in the PDF as the glyph id missing in the font. It would have worked if WeasyPrint supported glyph ids beyond 0xFFFF in Type 0 fonts.
- As a temporary solution, I used a modulo to get a glyph id missing in the font that’s lower than 0xFFFF.

I hope that it’s clearer!

> * In the current OpenType format the upper limit is 0xFFFE, because the number of glyphs is stored as an uint16 in the [maxp](https://github.com/harfbuzz/boring-expansion-spec/blob/main/beyond-64k.md) table.

That’s it, thanks, I didn’t remember where the limitation comes from. I remember there were ideas to avoid this limitation some time ago, but let’s hope it won’t happen soon. :crossed_fingers: 

> I know you’re trying very hard to both draw the .notdef glyphs and preserve the unsupported code points. But now after reading my clarification, is it possible that you’ve underestimated the challenge?

Your clarification is what I had in mind, so that’s no bad news. I was skeptical at one moment, but now that it actually works with your sample text, I’m quite confident. I "only" have to understand why it doesn’t work with the whole document.

Does [this PDF](https://github.com/user-attachments/files/21824635/tofu.pdf) work correctly with your PDF readers?

(Yes, I often underestimate challenges, otherwise I wouldn’t develop a browser. :smile:)

> And in Acrobat I can copy U+11B60 from that last whitespace.

Cool. It works everywhere I tried except in Chrome.

> Now I see the GID modified by `%= 0xffff` isn’t critical to preserving the text (which is the ToUnicode CMap’s responsibility),

:+1: 

> but then why do you have to use some integer based on the code point returned by Pango as the GID?

The final clean code won’t use the code point as the GID, it will choose a GID that’s not in the font.

For now, it uses the codepoint because in many cases, if a glyph is missing in the font, we can assume that the glyph id that is equal to the codepoint is missing from the font too. That’s not true in all cases, but it is in your example, and it’s an easy shortcut for a proof of concept.

lianghai:
> Does [this PDF](https://github.com/user-attachments/files/21824635/tofu.pdf) work correctly with your PDF readers?

Yes. It works as expected. The .notdef glyphs display fine. The underlying text of those .notdef glyphs can’t be copied from macOS Preview (which is sort of expected), but can be copied correctly from Acrobat or Chrome.

> We don’t want to preserve the original code point [as the GID], we just want to have one unused glyph id for each missing original codepoint.

> In the PDF, we use the Unicode codepoint (Pango fake glyph id - PANGO_GLYPH_UNKNOWN_FLAG) as a glyph id that is hopefully missing in the font. That’s not reliable, that’s temporary, until I write the code to get a glyph id that’s really missing in the font.

> For now, it uses the codepoint because in many cases, if a glyph is missing in the font, we can assume that the glyph id that is equal to the codepoint is missing from the font too. That’s not true in all cases, but it is in your example, and it’s an easy shortcut for a proof of concept.

I see. Earlier I thought you were planning to merge this solution (now I know it’s only a proof of concept) to the main branch…

> I can’t. Pango puts this space for us during the layout, not .notedef, and there’s nothing we can do about this as far as I know. If I put a .notedef char in the PDF string, I’ll still get the spacing problem (and lose search on top of that). If I use the .notdef width in the PDF, I break the layout.

I see, so you need to preserve the original width returned by Pango to not disturb the subsequent glyphs’ positions in the line. In this case the outlines don’t really matter, but you only need a glyph to support the ToUnicode CMap.

How many aspects do you need to differentiate between these placeholder .notdef glyphs? For example:

1. Width of a (nonexistent) glyph returned by Pango, for layout
  - Not yet clear if this width is constant
3. Requested Unicode code point, for warning and text extraction
5. .notdef outlines of the specified font, for user familiarity

The current WP release only handles 1, and this PR is almost ready for 2. It feels like 3 can be postponed, especially since it may increase the number of such placeholder .notdef glyphs you may need to handle.

Will it be easier if you actually use [Adobe NotDef](https://github.com/adobe-fonts/adobe-notdef) to catch all Unicode code points so this special case of PANGO_GLYPH_UNKNOWN_FLAG never arises?

Btw, fyi, this comment doesn’t match Pango’s documentation about what [GLYPH_EMPTY](https://docs.gtk.org/Pango/const.GLYPH_EMPTY.html) indicates:

https://github.com/Kozea/WeasyPrint/blob/06e5eb0378542d52f9d47afb4cfa21d56dce8c1b/weasyprint/draw/text.py#L180-L183

lianghai:
> Will it be easier if you actually use [Adobe NotDef](https://github.com/adobe-fonts/adobe-notdef) to catch all Unicode code points so this special case of PANGO_GLYPH_UNKNOWN_FLAG never arises?

Or, instead of relying on ToUnicode, can you just use the ActualText mechanism to annotate such cases? Its support is probably much more limited than ToUnicode, but .notdef is really a form of errors and it’s probably not necessary to try so hard to enable text extraction.

liZe:
> I "only" have to understand why it doesn’t work with the whole document.

@lianghai It was working on my computer, I found why: I have harfbuzz-subset installed and I included the .notdef character with it, but it’s not installed on GitHub Actions’ Ubuntu by default, so it uses (slower) fontools as a fallback. I’ve added the .notdef character with fonttools too, the glyph should be displayed on the PDF now.

If the result is OK for you (logging, selectable text, drawn notdef character, wrong width), I can clean the code (use unused gids, change variable names, fix wrong comments…) and merge it. I’ll probably investigate why the width is wrong later, but we can consider that it’s not mandatory to solve that for Unicode.

lianghai:
> @lianghai It was working on my computer, I found why: I have harfbuzz-subset installed and I included the .notdef character with it, but it’s not installed on GitHub Actions’ Ubuntu by default, so it uses (slower) fontools as a fallback.

Ah I see. I only referred to https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#linux (`pip install`) but not https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#ubuntu-20-04 (`apt install`), and thus didn’t install the additional dependencies.

> I’ve added the .notdef character with fonttools too, the glyph should be displayed on the PDF now.

Yes, now the latest run shows the .notdef glyphs: https://github.com/unicode-org/core-spec/actions/runs/17210685009.

> If the result is OK for you (logging, selectable text, drawn notdef character, wrong width), …

Yes, I can confirm that. Everything works as expected now, including the text extraction that works in Acrobat and Chrome but not in Preview, and the expected wrong width.

> …, I can clean the code (use unused gids, change variable names, fix wrong comments…) and merge it. I’ll probably investigate why the width is wrong later, but we can consider that it’s not mandatory to solve that for Unicode.

Thank you! The Unicode 17.0 release is coming up soon (Tue 9 Sep, in two weeks), therefore I wasn’t counting on this PR to be merged but I created a temporary QA branch instead: https://github.com/unicode-org/core-spec/compare/main...lianghai/pdf-qa. Therefore merging to main and releasing this feature would be great but no need to hurry.

liZe:
> Yes, I can confirm that. Everything works as expected now, including the text extraction that works in Acrobat and Chrome but not in Preview, and the expected wrong width.

Cool!

> The Unicode 17.0 release is coming up soon (Tue 9 Sep, in two weeks)

Good luck for the next two weeks.

> Therefore merging to main and releasing this feature would be great but no need to hurry.

I think that I’ll clean and merge it during the current week but there won’t be a release, so relying on the branch is better for 17.0.

> I’ll probably investigate why the width is wrong later

The width is [`pango_font_metrics_get_approximate_char_width`](https://docs.gtk.org/Pango/method.FontMetrics.get_approximate_char_width.html) according to the [relevant code](https://gitlab.gnome.org/GNOME/pango/-/blob/main/pango/pangoft2-render.c#L206). It’s unrelated to the .notdef glyph on purpose, as they try hard to not rely on glyph availability.

Pango also has a renderer, and [in this case](https://gitlab.gnome.org/GNOME/pango/-/blob/main/pango/pangocairo-render.c#L467) it [manually draws a rectangle](https://gitlab.gnome.org/GNOME/pango/-/blob/main/pango/pangocairo-render.c#L99). We could actually do the same, what you think? That wouldn’t affect text selection.

lianghai:
> The width is [`pango_font_metrics_get_approximate_char_width`](https://docs.gtk.org/Pango/method.FontMetrics.get_approximate_char_width.html) according to the [relevant code](https://gitlab.gnome.org/GNOME/pango/-/blob/main/pango/pangoft2-render.c#L206). It’s unrelated to the .notdef glyph on purpose, as they try hard to not rely on glyph availability.

Uh, that’s bizarre to avoid getting the width from .notdef.

> Pango also has a renderer, and [in this case](https://gitlab.gnome.org/GNOME/pango/-/blob/main/pango/pangocairo-render.c#L467) it [manually draws a rectangle](https://gitlab.gnome.org/GNOME/pango/-/blob/main/pango/pangocairo-render.c#L99). We could actually do the same, what you think? That wouldn’t affect text selection.

Displaying .notdef is a strong convention in the industry. These days it’s really unusual to see environments that do not follow that convention (Firefox is probably the only notable one). I’d argue that drawing the actual .notdef is still a better choice even without the correct width. Otherwise, you may as well just leave the glyph empty like what’s on the main branch currently, which (now combined with the warnings) is good enough.

I also suggest that you should have a brief section in the documentation about WeasyPrint’s .notdef behavior, where you can also explain why you’re not able to make the behavior ideal.

liZe:
> Displaying .notdef is a strong convention in the industry. These days it’s really unusual to see environments that do not follow that convention (Firefox is probably the only notable one)

On Linux, both GTK and Qt, the two main GUI toolkits, draw rectangles with codepoints as Firefox does. Chrome does that too in its interface, but not for web pages.

> I’d argue that drawing the actual .notdef is still a better choice even without the correct width. Otherwise, you may as well just leave the glyph empty like what’s on the main branch currently, which (now combined with the warnings) is good enough.

Let’s keep .notdef then!

> I also suggest that you should have a brief section in the documentation about WeasyPrint’s .notdef behavior, where you can also explain why you’re not able to make the behavior ideal.

OK.

lianghai:
> On Linux, both GTK and Qt, the two main GUI toolkits, draw rectangles with codepoints as Firefox does. Chrome does that too in its interface, but not for web pages.

I was referring to environments comparable to what WeasyPrint does, therefore web pages and document authoring. Since WeasyPrint takes web pages as the input, it’s helpful to stick the conventional behavior people see in web pages whenever possible.

Text layout and text shaping are complicated, and these inconsistent behaviors of low-level and history-burdened tools like Pango and GUI toolkits mislead people to feel that there isn’t a strong convention to follow. That’s not true.

Showing the code points is probably the only alternative behavior that’s worth considering. It’s unconventional but at least useful, it’s just more cumbersome to implement if Pango doesn’t already come with it. And you still need to deal with the unpredictable width problem.

liZe:
> I was referring to environments comparable to what WeasyPrint does, therefore web pages and document authoring. Since WeasyPrint takes web pages as the input, it’s helpful to stick the conventional behavior people see in web pages whenever possible.
> 
> Text layout and text shaping are complicated, and these inconsistent behaviors of low-level and history-burdened tools like Pango and GUI toolkits mislead people to feel that there isn’t a strong convention to follow. That’s not true.

:+1: 

> Showing the code points is probably the only alternative behavior that’s worth considering. It’s unconventional but at least useful, it’s just more cumbersome to implement if Pango doesn’t already come with it. And you still need to deal with the unpredictable width problem.

If we don’t draw the font’s .notdef but a custom notdef glyph (simple like [these ones](https://en.wikipedia.org/wiki/Notdef_glyph), or more complex with codepoints like Firefox), that will actually solve the width problem, as we can use Pango’s approximate width for our manual drawing. The code for that would simple, short for a rectangle with a cross or a question mark, longer for codepoints as we would have to manually draw the right digits at the right places.

liZe:
@lianghai I think that everything is ready before merging, tell me if I forgot anything.

lianghai:
Since https://github.com/Kozea/WeasyPrint/pull/2523#issuecomment-3220604740, there’s no noticeable changes in the generated PDF in terms of the .notdef, but there’s other regressions. Compare:

- Before: https://github.com/unicode-org/core-spec/actions/runs/17210685009
  - Logging: https://github.com/unicode-org/core-spec/actions/runs/17210685009/job/48821422097#step:12:1
- After: https://github.com/unicode-org/core-spec/actions/runs/17250306498
  - Logging: https://github.com/unicode-org/core-spec/actions/runs/17250306498/job/48950698167#step:12:1

“WARNING: Unable to mutate variable font” is a critical regression (you can see the main title “The Unicode® Standard” in the “After” PDF is not in bold anymore). You may need to tighten up your test cases about VFs.

Also, the “Missing glyph” warnings are gone, because of the overriding of `glyph_id` before the second check of `PANGO_GLYPH_UNKNOWN_FLAG`:

https://github.com/Kozea/WeasyPrint/blob/069f99d647b665cc63b593e3cf0f93f8fbbf41eb/weasyprint/draw/text.py#L186-L207

docs/api_reference.rst
```diff
@@ -182,14 +182,35 @@ specifications.
 Fonts
 ~~~~~
 
-WeasyPrint can use any font that Pango can find installed on the system. Fonts
-are automatically embedded in PDF files.
+WeasyPrint can use any font that Pango_ can find installed on the system. Fonts are
+automatically embedded in PDF files and are subset by default to only include the glyphs
+used in the PDF. Subsetting is done with Harfbuzz-Subset_ when available on the system,
```
lianghai, line 187:
```suggestion
used in the PDF. Subsetting is done with hb-subset_ when available on the system,
```

docs/api_reference.rst
```diff
@@ -182,14 +182,35 @@ specifications.
...
+etc. But some of these default rules can sometimes interfere with CSS rules, and it may
+be interesting to disable them if you need to tweak details about font management.
+
+When a glyph is missing from a font and all its fallbacks, a `Notdef glyph`_ is
+displayed instead, and a warning is displayed in logs. Because of the way Pango handles
+this case, the glyph may be larger than the space it’s displayed in, but the layout of
+the other glyphs is correct. Text search and selection work as if the glyph was
+available.
+
+.. _HarfBuzz-Subset: https://harfbuzz.github.io/harfbuzz-hb-subset.html
```
lianghai, line 208:
```suggestion
.. _hb-subset: https://harfbuzz.github.io/harfbuzz-hb-subset.html
```

docs/api_reference.rst
```diff
@@ -182,14 +182,35 @@ specifications.
...
-new font. WeasyPrint should support the major font formats handled by Harfbuzz.
+new font. WeasyPrint should support the major font formats handled by HarfBuzz_.
+
+WeasyPrint follows the Fontconfig_ configuration of the system. This default
+configuration is often useful for many cases: font fallbacks for missing glyphs, aliases
+for standard font families like "serif" or "monospace", colored variants for emojis,
+etc. But some of these default rules can sometimes interfere with CSS rules, and it may
+be interesting to disable them if you need to tweak details about font management.
+
+When a glyph is missing from a font and all its fallbacks, a `Notdef glyph`_ is
```
lianghai, line 202: In text shaping, a font’s public API is queried with Unicode code points, not glyph IDs. Therefore it’s really “a code point is not supported by fonts” instead of “a glyph is missing in fonts”.

```suggestion
When a Unicode code point is not supported by a font and its fallbacks, the font’s `.notdef glyph`_ is
```

docs/api_reference.rst
```diff
@@ -182,14 +182,35 @@ specifications.
...
+this case, the glyph may be larger than the space it’s displayed in, but the layout of
+the other glyphs is correct. Text search and selection work as if the glyph was
+available.
+
+.. _HarfBuzz-Subset: https://harfbuzz.github.io/harfbuzz-hb-subset.html
+.. _fontTools: https://fonttools.readthedocs.io/en/latest/
+.. _Pango: https://www.pango.org/
+.. _Fontconfig: https://www.freedesktop.org/wiki/Software/fontconfig/
+.. _HarfBuzz: https://harfbuzz.github.io/
+.. _Notdef glyph: https://en.wikipedia.org/wiki/Notdef_glyph
```
lianghai, line 213: This terminology choice doesn’t matter to average users, but is more natural to professionals.

```suggestion
.. _.notdef glyph: https://en.wikipedia.org/wiki/Notdef_glyph
```

weasyprint/draw/text.py
```diff
@@ -172,15 +173,40 @@ def draw_first_line(stream, textbox, text_overflow, block_ellipsis, matrix):
...
+            # Create mapping between glyphs and Unicode codepoints.
+            if glyph_id not in font.to_unicode:
+                utf8_slice = slice(*sorted(utf8_positions[i:i+2]))
+                text = font.to_unicode[glyph_id] = utf8_text[utf8_slice].decode()
+
+            # Log warning for missing glyphs.
+            if glyph_id & pango.PANGO_GLYPH_UNKNOWN_FLAG:
+                codepoints = ', '.join(f'U+{ord(character):04X}' for character in text)
+                LOGGER.warning(
+                    'Missing glyph when displaying "%s", Unicode codepoint(s) %s',
```
lianghai, line 206: Besides moving this logging to precede `glyph_id = font.get_unused_glyph_id(codepoint)`, there’s also some editorial suggestion:

```suggestion
                    '.notdef glyph rendered for Unicode string unsupported by fonts: "%s" (%s)',
```

docs/api_reference.rst
```diff
@@ -182,14 +182,35 @@ specifications.
...
+
+WeasyPrint follows the Fontconfig_ configuration of the system. This default
+configuration is often useful for many cases: font fallbacks for missing glyphs, aliases
+for standard font families like "serif" or "monospace", colored variants for emojis,
+etc. But some of these default rules can sometimes interfere with CSS rules, and it may
+be interesting to disable them if you need to tweak details about font management.
+
+When a glyph is missing from a font and all its fallbacks, a `Notdef glyph`_ is
+displayed instead, and a warning is displayed in logs. Because of the way Pango handles
+this case, the glyph may be larger than the space it’s displayed in, but the layout of
```
lianghai, line 204:
```suggestion
this case, the .notdef glyph may be rendered with an incorrect width (advance), but the layout of
```

weasyprint/draw/text.py
```diff
@@ -172,15 +173,36 @@ def draw_first_line(stream, textbox, text_overflow, block_ellipsis, matrix):
...
+            # Display zero-width empty glyph.
+            if glyph_id == pango.PANGO_GLYPH_EMPTY:
                 string += f'>{-width / font_size}<'
                 continue
 
+            # Display .notdef and log warning for missing glyphs.
+            if glyph_id & pango.PANGO_GLYPH_UNKNOWN_FLAG:
+                codepoint = glyph_id - pango.PANGO_GLYPH_UNKNOWN_FLAG
+                LOGGER.warning(
+                    '.notdef glyph rendered for Unicode string unsupported by fonts: '
```
lianghai, line 190: Oh I see now it’s only one code point then we don’t need to log it as a string anymore:

```py
                LOGGER.warning(
                    '.notdef glyph rendered for Unicode code point unsupported by fonts: '
                    f'U+{codepoint:04X} "{chr(codepoint)}"')
```

liZe:
I’ve fixed all the problems I’ve found.

> You may need to tighten up your test cases about VFs.

We definitely should. I already had great times "enhancing" the venerable Ahem font used by the W3C for their test suite because I thought it was not enough for our tests. I added kerning and ligatures to test `font-features`, converted it into woff to test the format, and even built a bitmap variant (yes, we have clients who use that for pixel-perfect e-ink labels). I didn’t find the time yet to learn how to create a testing variable font, and bitmap emojis, and SVG emojis… That’s on my TODO list.

This problem has been introduced by #2532. On top of that there’s a bug in fontTools, I opened a PR: fonttools/fonttools#3918.

> Also, the “Missing glyph” warnings are gone

That’s another untested case, as we depend on system fonts to correctly test that. Most of WeasyPrint's code is already tests, but it’s never enough.

lianghai:
The latest workflow run looks all good now: https://github.com/unicode-org/core-spec/actions/runs/17272631593.

> I didn’t find the time yet to learn how to create a testing variable font, and bitmap emojis, and SVG emojis… That’s on my TODO list.

I’m happy to help with the VF test. We can schedule a quick meeting some day after the Tue 9 Sep release day of Unicode 17.0, or you can point me to an issue that outlines what you need.

liZe:
> The latest workflow run looks all good now

Cool!

If it’s OK for you I’ll merge this PR. For Unicode 17.0, you can reference 836132a in your dependency URL. `main` may change a lot in the next days, you shouldn’t rely on it.

> I’m happy to help with the VF test.

Thanks a lot. There’s no hurry, I’ll send you a message after 17.0.

lianghai:
> If it’s OK for you I’ll merge this PR. For Unicode 17.0, you can reference https://github.com/Kozea/WeasyPrint/commit/836132a70059ded0b416af806db2c8b0ff829fd1 in your dependency URL.

Sure, no problem.

> `main` may change a lot in the next days, you shouldn’t rely on it.

Yeah I only use released versiones for production, and PRs for testing.

Issue #2140, opened 2024-04-30
Support of `@media only`


vpoulailleau:
With weasyprint 62, I get `WARNING: Invalid media type ' only print ' the whole @media rule was ignored at 3:5.` for the following HTML:

```html
<html lang="fr">
  <head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Titre de la page</title>
    <style>
    h1 {color: red;}
    @media only print {
      h1 {
        color: green;
      }
    }
    </style>
  </head>
  <body>
    <main>
      <h1>Contenu Principal</h1>
      <p>Ceci est un paragraphe de texte.</p>
    </main>
  </body>
</html>
```

According to https://www.w3.org/TR/mediaqueries-5/#typedef-media-query-list this is valid CSS. `only` seems not supported yet (I haven't tested `not`, `or`, `and` for `@media not screen` for instance).

I can easily workaround this limitation by removing `only`.

liZe:
Hi!

Thanks for the report. Media queries are not very well supported, this is an old problem, see #494.

The main problem we have with media queries is that it’s hard to define the viewport dimension on paged media. But supporting "only" (and probably "not", "or" and "and", as long as we have only media types) is actually much easier.

Interested in adding this feature? That’s a good first issue if you have some Python skills and no idea about how WeasyPrint works!

The magic happens here:
https://github.com/Kozea/WeasyPrint/blob/d5d7ce369aef035712cf73446f9085a32105846f/weasyprint/css/media_queries.py#L25-L39

vpoulailleau:
@liZe I currently don't have the time right now. But I would be glad to help.

BTW, will there be a coding sprint for weasyprint at PyCon FR 2024? I planned to come and help Julien for the Python docs translation, but I would be happy to better understand weasyprint that I use quite often. We saw each other during last AFPy's assemblée générale !

liZe:
> I currently don't have the time right now. But I would be glad to help.

No problem, let’s save this chance to another newcomer!

> BTW, will there be a coding sprint for weasyprint at PyCon FR 2024? I planned to come and help Julien for the Python docs translation, but I would be happy to better understand weasyprint that I use quite often. We saw each other during last AFPy's assemblée générale !

😄

[...]

PR #2524, opened 2025-08-13
Fix: Add support for color-hint in CSS gradients

### LINKED ISSUE #1961: Color hinting in gradients not recognized
It seems that the parser doesn't recognize the color-hint term in gradient definitions. This is a bare percentage that may be specified between any two adjacent stops to fine-tune the interpolation. See [here](https://gist.github.com/Gayo/d66181a59507d4a842bf83eb890ca07e) for a minimal example. The code anticipates that it can apply parse_color_stop to each comma-separated item of the stop list, and thus throws an error on the color hint.

FazleRabbbiferdaus172:
This pull request resolves the issue where the parser fails to recognize the color-hint syntax in CSS gradient definitions.

As noted in the original issue, the current parser expects every comma separated item in a gradient's stop list to be a full color stop. It throws an error when it encounters a color hint, which is a percentage value used to control the transition midpoint between two color-stops.

It is still a draft, need to add test cases, review is needed

Fixes #1961

### DIFF
--- a/weasyprint/css/utils.py
+++ b/weasyprint/css/utils.py
@@ -622,8 +622,8 @@ def get_image(token, base_url):
     name = token.lower_name
     if name in ('linear-gradient', 'repeating-linear-gradient'):
         direction, color_stops = parse_linear_gradient_parameters(arguments)
-        color_stops, color_hint = get_color_stop_and_hint(color_stops)
         if color_stops:
+            color_stops, color_hint = get_color_stop_and_hint(color_stops)
             return 'linear-gradient', LinearGradient(
                 [parse_color_stop(stop) for stop in color_stops],
                 direction, 'repeating' in name,
--- a/weasyprint/css/utils.py
+++ b/weasyprint/css/utils.py
@@ -5,6 +5,7 @@ import math
 from abc import ABC, abstractmethod
 from urllib.parse import unquote, urljoin
 
+from tinycss2.ast import PercentageToken
 from tinycss2.color4 import parse_color
 
 from .. import LOGGER
--- a/weasyprint/css/utils.py
+++ b/weasyprint/css/utils.py
@@ -620,11 +620,13 @@ def get_image(token, base_url):
     arguments = split_on_comma(remove_whitespace(token.arguments))
     name = token.lower_name
     if name in ('linear-gradient', 'repeating-linear-gradient'):
-        direction, color_stops = parse_linear_gradient_parameters(arguments)
+        direction, color_stops, color_hint = parse_linear_gradient_parameters(arguments)
+        color_stops, color_hint = get_color_stop_and_hint(color_stops)
         if color_stops:
             return 'linear-gradient', LinearGradient(
                 [parse_color_stop(stop) for stop in color_stops],
-                direction, 'repeating' in name)
+                direction, 'repeating' in name,
+                color_hint = [get_length(hint, negative=False, percentage=True) for hint in color_hint])
     elif name in ('radial-gradient', 'repeating-radial-gradient'):
         result = parse_radial_gradient_parameters(arguments)
         if result is not None:
@@ -635,9 +637,12 @@ def get_image(token, base_url):
             position = 'left', FIFTY_PERCENT, 'top', FIFTY_PERCENT
             color_stops = arguments
         if color_stops:
+            color_stops, color_hint = get_color_stop_and_hint(color_stops)
             return 'radial-gradient', RadialGradient(
                 [parse_color_stop(stop) for stop in color_stops],
-                shape, size, position, 'repeating' in name)
+                shape, size, position, 'repeating' in name,
+                color_hint=[get_length(hint, negative=False, percentage=True) for hint in color_hint]
+            )
 
 
 def _get_url_tuple(string, base_url):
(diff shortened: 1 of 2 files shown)


FazleRabbbiferdaus172:
Hi @liZe,

This is a draft PR based on my understanding of the steps you mentioned. So far, the color hint property is being parsed, all current tests are passing, and the "minimal example" from the issue is generated correctly.

I haven't added new tests yet because I'm not sure I fully understand how a color hint should be calculated, especially the default value if a hint is omitted.

My understanding from MDN is that "If omitted, the midpoint of the color transition is the midpoint between two color stops." This seems to imply the midpoint doesn't shift, but it's not clear if that means the default value should be 50%. From the steps you mentioned [here](https://github.com/Kozea/WeasyPrint/issues/1961#issuecomment-1713387563) the step 3 seems to change the alpha value, So how does the 50% or mid point coverts to alpha value?

Could you please help clarify this? I would also appreciate your feedback on the code so far. Thanks!

liZe:
> Could you please help clarify this? I would also appreciate your feedback on the code so far. Thanks!

Thanks a lot, I’ll review your pull request soon!

liZe:
> This is a draft PR based on my understanding of the steps you mentioned. So far, the color hint property is being parsed, all current tests are passing, and the "minimal example" from the issue is generated correctly.

It looks like `radial_gradients_5` `and linear_gradients_13` actually fail on the CI.

> My understanding from MDN is that "If omitted, the midpoint of the color transition is the midpoint between two color stops." This seems to imply the midpoint doesn't shift, but it's not clear if that means the default value should be 50%.

It does. For example (that’s really simplified), 50% means that if you have a gradient with red at the beginning and blue at the end, the purple will be in the middle. That’s the default. 25% means that 25% of the gradient is blue-to-purple-ish, and 75% is purple-to-red-ish.

> From the steps you mentioned [here](https://github.com/Kozea/WeasyPrint/issues/1961#issuecomment-1713387563) the step 3 seems to change the alpha value, So how does the 50% or mid point coverts to alpha value?

The goal is to store the hint as a third value in `alpha_couples` and `color_couples`. For colors it’s currently set to 1 (that corresponds to 50%), it should be set to the actual hint value. For alpha, the third value has to be added.

https://github.com/Kozea/WeasyPrint/blob/aefe4af4f825b55846bb35164d5b5196f5a62701/weasyprint/images.py#L492-L498

For alpha, the hint should be used in the sub function (instead of 1 currently), just as it’s done for color a few lines above, using `n`.

https://github.com/Kozea/WeasyPrint/blob/aefe4af4f825b55846bb35164d5b5196f5a62701/weasyprint/images.py#L531-L533

https://github.com/Kozea/WeasyPrint/blob/aefe4af4f825b55846bb35164d5b5196f5a62701/weasyprint/images.py#L516-L518

50% should be stored as 1 in `alpha_couples` and `color_couples`, but I’m not sure for other values because we use exponential functions in PDF. As a first approximation, you can store hint × 2 (1 for 50%, 0.5 for 25%, etc.) We’ll do the math later!

> I would also appreciate your feedback on the code so far. Thanks!

It’s great so far. You can try to fix the failing tests, set the values in `alpha_couples` and `color_couples` (step 3) and manually try with simple cases. I’ll review the code in detail when step 3 is done, you’re on the right tracks already!

FazleRabbbiferdaus172:
I forgot to push the latest commit then so the radial_gradients_5 and linear_gradients_13 was failing in the ci. Now it is pushed.

I will try to refactor the code according to you instructions

FazleRabbbiferdaus172:
Updated third value in alpha_couples and color_couples and hints are used in subfunctions. But when i manually tested noticed some difference between current version and version before [15b1ae5](https://github.com/Kozea/WeasyPrint/pull/2524/commits/15b1ae5d46189a0349ae9d532675969416fdf9bc) commit. I guess the math needs to be corrected to match html version. For reference check the following images.

1. Current version:
image
image

2. Version before commit [15b1ae5](https://github.com/Kozea/WeasyPrint/pull/2524/commits/15b1ae5d46189a0349ae9d532675969416fdf9bc)  :
image
image

3. html version:
image
image

liZe:
Thanks a lot! With some extra math, here’s what I get (left is the browser, right is WeasyPrint):

Capture d’écran du 2025-08-31 12-55-37

The 50% point is the same but the gradient is slightly different, I suppose because we use an exponential function between control points and CSS wants a linear function. We can definitely live with that for now!

I’ll clean the code, handle corner cases (0% or less, 100% or more, positions other than percentages, wrong syntaxes) and see if I can add some tests.

liZe:
@FazleRabbbiferdaus172 Thanks for your hard work on this. If everything is OK for you with my latest fixes, it’s time to merge!

FazleRabbbiferdaus172:
Sounds great, and thank you for your guidance on this. The latest fixes look good to me. I've learned a lot and really enjoyed working on this and would be interested in working on another issue/feature if there's anything available.

liZe:
> Sounds great, and thank you for your guidance on this. The latest fixes look good to me. I've learned a lot and really enjoyed working on this and would be interested in working on another issue/feature if there's anything available.

:pray: 

Another interesting issue could be #2140. Supporting `only` is a first step, but you can continue with `not` and the rest of the specification if you’re interested!

Issue #2531, opened 2025-08-21
WeasyPrint using deprecated fontTools API


marpulli:
I fontTools version [4.59.1](https://github.com/fonttools/fonttools/releases/tag/4.59.1) the `fonttools varLib.mutator.instantiateVariableFont` function was deprecated in favour of `fonttools varLib. instancer.instantiateVariableFont`.

WeasyPrint is using the deprecated API [here](https://github.com/Kozea/WeasyPrint/blob/eb47d81dd1dc4f7080bb45fa75e0515cf2c5ad5e/weasyprint/pdf/fonts.py#L12).

PR #2532, opened 2025-08-21
Use fonttools instancer instead of deprecated mutator API

marpulli:
Fixes https://github.com/Kozea/WeasyPrint/issues/2531

### DIFF
--- a/weasyprint/pdf/fonts.py
+++ b/weasyprint/pdf/fonts.py
@@ -140,9 +140,6 @@ class Font:
             partial_font = io.BytesIO()
             try:
                 ttfont = instantiateVariableFont(ttfont, self.variations)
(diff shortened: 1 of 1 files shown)


liZe:
Thank you! :heart: 

I think that it would require at least fonttools 4.52.0 because we need fonttools/fonttools#3506. Could you please update `pyproject.toml` and `docs/first_steps.rst`? I’ll check that everything works as expected, especially with CFF2.

(It also means that we can remove our workaround for fonttools/fonttools#2827, I’ll add a commit for that.)

marpulli:
Thanks for taking a look so quickly! I've added the new fontTools constraints in those two places

liZe:
Thank you!

PR #2538, opened 2025-08-29
fix TypeError: 'NoneType' object is not iterable for arguments.extend(resolve_var())

kuypan:
I get TypeError: 'NoneType' object is not iterable when running arguments.extend(resolve_var()) in css/__init__.py.
i add a check for the result of resolve_var(). now it works.


liZe:
Hi,

thanks for this pull request. Could you please provide an example that triggers this error, so that we can find the root of the problem and add a test?

kuypan:
https://mp.weixin.qq.com/s/YcahZnJ76hK_p3CvGKgJGQ

liZe:
Short way to reproduce: `<style>body { color: unknown(unknown(), var(--a)) }</style>`

liZe:
I’ve removed the comment and added a test, thanks for this pull request!

Issue #494, opened 2017-07-29
@media only produces parse error


MattSturgeon:
I have the following line in a stylesheet:

```css
@media only screen and (min-width: 480px) {  }
@media only screen and (min-width: 768px) {  }
@media print {  }
```

When I use this with WeasyPrint I get the following warning:

```
WARNING: Parse error at 1:8, expected a media type, got IDENT, IDENT, IDENT, (
WARNING: Parse error at 2:8, expected a media type, got IDENT, IDENT, IDENT, (
```

From what I can tell, the `only` keyword is not being recognised, so perhaps unknown media types produce a parse error instead of a more descriptive warning **and** `only` is an unknown media type?

`only` is [intended](https://stackoverflow.com/questions/8549529) to filter out older browsers who don't support media queries but do support comma-separated media types.

liZe:
[Media queries](https://www.w3.org/TR/css3-mediaqueries/#media0) are not supported at all for the moment, it's a good feature to have.

doekman:
Another use-case for media queries (just making my case to prioritize this issue):

With `@media (max-height:` you can for example select the first page (because the margins are bigger), and display a complex page footer (for example a table) with display:fixed that only shows up on the first page.

liZe:
My bad, it's not a good first issue at all:
- layout needs a formatting structure,
- the formatting structure needs CSS rules (at least `display`)
- some CSS rules depend on media queries,
- media queries define different CSS rules for different pages,
- pagination is done during the layout,
- layout needs a formatting structure…

liZe:
I've pushed a [`media-queries` branch](https://github.com/Kozea/WeasyPrint/tree/media-queries) with initial work for supporting media queries.

Next steps:
- make `style_for` object change its `_cascaded_styles` and `_computed_styles` attribute change for each page depending on the curent page attributes (mainly width and height), and cache them to avoid recalculation for known attributes,
- change the logic to use a different `root_box` for each page in `Document._render`, using the updated `style_for` object,
- change the `resume_at` attribute to be independent from the `root_box` (that's the really hard point).

liZe:
> We could even add a custom `-weasy-page-name` feature for that, instead of relying on page size .

And left/right pages, and page numbers, and…

Lucas-C:
Just sharing some simple code I used to silence those warnings:
```
16:42:54 weasyprint [WARNING] Expected a media type, got '(min-width: 768px)'
[...]

PR #2547, opened 2025-09-05
[FEATURE] add support logical operator only

FazleRabbbiferdaus172:
This pull request support for the only keyword in CSS media queries to improve compliance with the W3C specification.

It is still a draft, need to add test cases, review is needed.

Fixes https://github.com/Kozea/WeasyPrint/issues/2140

### DIFF
--- a/weasyprint/css/media_queries.py
+++ b/weasyprint/css/media_queries.py
@@ -10,6 +10,18 @@ from ..logger import LOGGER
 from .tokens import remove_whitespace, split_on_comma
 
 
+MEDIA_TYPES = {
+    'all',
+    'print',
+    'screen'
+}
+
+# For now support for only operator
+LOGICAL_OPERATORS = {
+    'only'
+}
+
+
 def evaluate_media_query(query_list, device_media_type):
     """Return the boolean evaluation of `query_list` for the given
     `device_media_type`.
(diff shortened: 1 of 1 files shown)


liZe:
Thank you :heart: 

If you’re interested in going further on the topic, don’t hesitate to ask! The long-term goal would be to fix #494.

The number of different properties can be impressive, but most of them only require configuration flags (like `prefers-contrast`) or pre-defined true/false answers (like `hover` that is always false).

Most of the work would actually be the parsing part, and it looks like the work you wanted to do in your first commits. :smile:

Thanks again!

FazleRabbbiferdaus172:
Thanks for your guidance! I'd love to help with #494.

My pace might be a bit slower while I job hunt as I am on tight timeline in a new country 😐, but I'll poke around the parsing logic whenever I can manage time. Thanks!

liZe:
> Thanks for your guidance! I'd love to help with #494.

Cool!
 
> My pace might be a bit slower while I job hunt as I am on tight timeline in a new country 😐, but I'll poke around the parsing logic whenever I can manage time. Thanks!

No need to hurry, good luck for your job hunt. :crossed_fingers:

FazleRabbbiferdaus172:
Thank you. 😄

PR #2563, opened 2025-09-29
Don’t avoid float collisions for atomic flex items

### LINKED ISSUE #2479: img alignment broken
Since v65, img elements with flexbox do not align anymore, for example despite space-between it overlaps:

Example
```
<style>
  .page-top {
    display: flex;
    align-items: center;
    justify-content: space-between;
    width: 100%;

    img {
      display: block;
      height: 15mm;
      width: auto;
    }
  }
</style>
<div class="page-top">
  <h1>Title</h1>
  <img src="myurl.com" alt="">
</div>
```

Wrapping it in a div element fixes the issue:
```
<style>
  .page-top {
    display: flex;
    align-items: center;
    justify-content: space-between;
    width: 100%;

    img {
      display: block;
      height: 15mm;
      width: auto;
    }
  }
</style>
<div class="page-top">
  <h1>Title</h1>
  <div class="logo">
    <img src="myurl.com" alt="">
  </div>
</div>
```

liZe:
This is just the same workaround as the one for non-atomic boxes. The real fix would be to avoid calling `block_level_layout_switch()` in flex layout, as flex items are not block boxes.

But we can keep this real fix for later, as we need the same fix for grid, tables, columns, etc.

Fix #2479.

### DIFF
--- a/tests/layout/test_flex.py
+++ b/tests/layout/test_flex.py
@@ -1951,3 +1951,19 @@ def test_flex_image_min_width():
     article, = body.children
     img, div = article.children
     assert article.height == img.height == img.width == div.height == 10
+
+
+@assert_no_logs
+def test_flex_image_justify_content():
+    page, = render_pages('''
+      <article style="width: 100px; display: flex; justify-content: end">
+        <img src="pattern.png">
+      </article>
+    ''')
+    html, = page.children
+    body, = html.children
+    article, = body.children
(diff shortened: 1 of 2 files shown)


liZe:
Quick note to myself: please add some tests before merging.

liZe:
> Quick note to myself: please add some tests before merging.

Done, thank you!

Issue #2573, opened 2025-10-06
Wrong PDF/A conformance levels for PDF/A-4


malte-laukoetter:
Currently weasyprint supports PDF/A-4 conformance levels it calls `PDF/A-4b` and `PDF/A-4u`. These conformance levels do not exists for PDF/A-4 (but did for PDF/A-2 and PDF/A-3). Instead PDF/A-4 has the conformance levels `PDF/A-4e` and `PDF/A-4f`.

liZe:
Hi!

Thanks for the report. I opened #2611 to fix this issue and test PDF/A. Feedback is welcome!

PR #2575, opened 2025-10-06
Add support for PDF/A-1a, PDF/A-2a and PDF/A-3a

### LINKED ISSUE #2574: Support creating PDFs conforming to PDF/A-1a, PDF/A-2a and PDF/A-3a
Weasyprint currently does not support creating files conforming to the `a`-conformance-levels of PDF/A. This conformance level indicates that the files are additionally to the requirements of the `u`-variant (or `b`-variant for PDF/A-1) also is accessible. Weasyprint already supports creating PDF/UA-1 conforming (and therefore accessible) files.

Could you add support for the `a`-conformance-levels as well?

In my testing I was able to create PDF/A-2a conforming files by choosing PDF/A-2u and also using the `--pdf-tags` flag to create an accessible pdf file. The only PDF/A-2a rule the file was then not conforming to was that the conformance-level is not indicated in the metadata (https://github.com/veraPDF/veraPDF-validation-profiles/wiki/PDFA-Parts-2-and-3-rules#rule-664-3) By manually overwriting the metadata afterwards I was able to create a PDF/A-2a conforming file.

Therefore the following addition to `pdfa.py` should be enough to fix this:
```python
'pdf/a-1a': (
    partial(pdfa, version=1, variant='A'),
    {'version': '1.4', 'identifier': True, 'srgb': True, 'pdf_tags': True}),
'pdf/a-2a': (
    partial(pdfa, version=2, variant='A'),
    {'version': '1.7', 'identifier': True, 'srgb': True, 'pdf_tags': True}),
'pdf/a-3a': (
    partial(pdfa, version=3, variant='A'),
    {'version': '1.7', 'identifier': True, 'srgb': True, 'pdf_tags': True}),
```

malte-laukoetter:
Fixes #2574

### DIFF
--- a/weasyprint/pdf/pdfa.py
+++ b/weasyprint/pdf/pdfa.py
@@ -90,4 +90,13 @@ VARIANTS = {
     'pdf/a-4u': (
         partial(pdfa, version=4, variant='U'),
         {'version': '2.0', 'identifier': True, 'srgb': True}),
+    'pdf/a-1a': (
+        partial(pdfa, version=1, variant='A'),
+        {'version': '1.4', 'identifier': True, 'srgb': True, 'pdf_tags': True}),
+    'pdf/a-2a': (
+        partial(pdfa, version=2, variant='A'),
+        {'version': '1.7', 'identifier': True, 'srgb': True, 'pdf_tags': True}),
+    'pdf/a-3a': (
+        partial(pdfa, version=3, variant='A'),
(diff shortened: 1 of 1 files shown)


liZe:
Thank you!

When we developed our first PDF/A support, we were far from supporting PDF/A-xa, so that’s really good news to see that the latest improvements helped to achieve this.

I’ll add tests for PDF/A flavors, as we did for PDF/UA. (And fix #2573 too!)

PR #2581, opened 2025-10-14
Add option to disable protocols in url resolution

ws-ykothmeier:
It may be problematic to resolve specific protocols, such as `file://`, when the input is read from an untrusted source.

To solve this issue, this pull requests adds an option `--allowed-protocols`, a comma-separated list of protocols that are allowed when fetching urls. All other protocols will not be accessed.

If the option is not given, all protocols are implicitly allowed (matches the previous behavior)

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -576,6 +576,31 @@ def test_to_unicode_rtl():
     assert b'<00d0> <0628>' in stdout
 
 
+@assert_no_logs
+@pytest.mark.parametrize('command', [
+    '- -',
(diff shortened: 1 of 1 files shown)


liZe:
Hi!

Thanks for your pull request.

I’m usually not a big fan of new CLI options for the URL fetcher, because the needs are so specific that we’ll never cover them all, but this feature is useful, easy to understand and small in the code, so… why not!

I’ll add a few minor changes and merge the pull request.

liZe:
Thanks a lot! :heart:

PR #2593, opened 2025-10-29
Get rid of zopfli dependency

jenskutilek:
Update `fonttools` version; remove the "woff" fonttools extra to get rid of the `zopfli` dependency and add `brotli` dependencies directly to keep WOFF2 font support.

The `zopfli` dependency is not needed to decompress WOFF fonts and is problematic because it lags behind in supplying wheels for new Python versions. It can be removed because WOFF (v1) fonts can be decompressed without it, and WOFF (v1) is a legacy format anyway.

`brotli` is required to decompress WOFF2 fonts, so I added the brotli dependencies directly to weasyprint.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,9 @@ dependencies = [
   'cssselect2 >=0.8.0',
   'Pyphen >=0.9.1',
   'Pillow >=9.1.0',
-  'fonttools[woff] >=4.59.2',
+  'fonttools >=4.60.1',
+  'brotli >= 1.0.1; platform_python_implementation == "CPython"',
+  'brotlicffi >= 0.8.0; platform_python_implementation != "CPython"',
 ]
 classifiers = [
   'Development Status :: 5 - Production/Stable',


liZe:
Hi!

Thanks for your pull request. :pray: 

I understand your problem and would be really, really happy to drop the zopfli dependency too (see fonttools/py-zopfli#7, fonttools/py-zopfli#11, fonttools/py-zopfli#18 :smile:).

But the fix would be better in fonttools, not in WeasyPrint: we do want the woff support in fonttools, `fonttools[woff]` seems to be the right choice. Handling the dependencies of fonttools here would just be a temporary workaround until fonttools/fonttools#3960 is fixed, it would not help more than a real fix at the root of the problem.

In my opinion, the first step is to open a pull request on py-zopfli to get wheels for 3.14. We can then take some time to fix fonttools/fonttools#3960 and update WeasyPrint dependencies if needed.

Don’t hesitate to ping me if I can help!

PR #2594, opened 2025-10-30
Add missing `__init__.py` files

jodal:
...to make `weasyprint.formatting_structure` and `weasyprint.text` proper Python packages.


liZe:
Hi,

Thanks for your proposal.

What’s the goal of this pull request? Python doesn’t require `__init__.py` files for modules, why would this be better for you?

jodal:
I found this when trying to get the type checker ty to pass on our code base, and it failed to resolve some imports from weasyprint because of these missing init files.

Of course, not having init files is valid and something that ty should support eventually. Not having init files is also most probably an oversight, unless the intention is to make a namespace package instead of a regular module.

Given that you have init files in several other modules, and this didn't look like actual namespace packages, I figured this was probably a wanted fix.

liZe:
> I found this when trying to get the type checker ty to pass on our code base, and it failed to resolve some imports from weasyprint because of these missing init files.

OK, thanks for explaining! The bug is probably in ty then. :smile: 

> Given that you have init files in several other modules, and this didn't look like actual namespace packages, I figured this was probably a wanted fix.

It makes sense.

We currently have `__init__.py` files only when we want to have code shared at the top level of the module, and omit it when we don’t. I try hard to avoid "useless" files in our repositories to keep it clean and readable, I’m for example pretty proud of the root of the repository. Some of my friends think it’s a kind of strange obsession, and well, they’re probably right :smile:. So we’ll keep the current structure, even if the packages are not really namespace packages.

Have fun with WeasyPrint!

PR #2595, opened 2025-10-30
fix: incomplete document generated because of an overflow issue

michalfapso:
I don't know if this is the correct way to fix it, but it works for my mkdocs project. Without this patch, my document stopped at page 159 of 250, but a similar issue occurred on other pages too, usually when changing a font size or something else which shifted page layouts.

### DIFF
--- a/weasyprint/layout/block.py
+++ b/weasyprint/layout/block.py
@@ -804,9 +804,12 @@ def block_container_layout(context, box, bottom_space, skip_stack, page_is_empty
         elif stop:
             if box.height != 'auto':
                 if context.overflows(box.position_y + box.border_height(), position_y):
-                    # Box heigh is fixed and it doesn’t overflow page, forget
-                    # overflowing children.
-                    resume_at = None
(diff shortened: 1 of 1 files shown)


liZe:
Hi!

Thanks for your pull request.

Do you have a sample that shows the problem? That would help to see what’s wrong, check that this is the fix we want, and add a test to avoid regressions.

michalfapso:
Hi @liZe,

Sure, I'll try to prepare a minimal example of the issue. It may take a few days. I'll send it here when it's ready.

michalfapso:
I'm sorry it took more time than I thought, @liZe, but here is an example repo where you can reproduce the bug:
https://github.com/michalfapso/bug_mkdocs-to-pdf_weasyprint_layout_failed

liZe:
> I'm sorry it took more time than I thought, @liZe, but here is an example repo where you can reproduce the bug:

Thanks!

That’s a dirty bug of collapsing margins, I’ll fix it in another pull request.

PR #2600, opened 2025-11-08
Fix SVG image aspect ratio when only width or height specified

ernestchu:
SVG `<image>` elements with only width or height specified were not preserving aspect ratio, causing images to stretch or squash.

## Root Cause

When only one dimension was specified, the code fell back to intrinsic dimensions without accounting for the provided dimension:

```python
# Before: width=0 when unspecified, then width = 0 or intrinsic_width
width = width or intrinsic_width  # Uses full intrinsic width
height = height or intrinsic_height  # Uses specified height → incorrect aspect ratio
```

For a 4×4 image with `height="3px"` unspecified width, this produced 4×3 instead of 3×3.

## Changes

**weasyprint/svg/images.py**
- Calculate missing dimension from intrinsic aspect ratio when only one is specified:
  ```python
  if width and not height:
      height = width / intrinsic_ratio
  elif height and not width:
      width = height * intrinsic_ratio
  ```

**tests/draw/svg/test_images.py**
- Added tests for `<image>` with only height and only width specified

### DIFF
--- a/tests/draw/svg/test_images.py
+++ b/tests/draw/svg/test_images.py
@@ -327,18 +327,22 @@ def test_image_image_wrong(assert_pixels):
 def test_image_in_g_height_only(assert_pixels):
     """Test that image inside g with only height set preserves aspect ratio."""
     assert_pixels('''
-        rBBB
-        BBBB
-        BBBB
-        BBBB
+        rrBBBBBB
+        rrBBBBBB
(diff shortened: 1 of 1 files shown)


liZe:
Thank you!

I’ll update the tests (they pass even without the fix) and merge your pull request. :heart:

Issue #2603, opened 2025-11-12
Allow URL fetchers to raise uncaught exceptions/refactoring API


svbnet:
I have been using Weasyprint for a while and it's been great. However, one area I've found that could be improved on is the implementation of custom URL fetchers.

## Error handling

I believe I have a use case that's similar to a lot of peoples', which is that PDFs are generated largely using local assets, and indeed, the relevant Django and Flask packages are built to help with this. I would say the other use case is converting any random HTML document to a PDF, either through the CLI or other means.

For the former use case, I would consider it critical that all external URLs are fetched successfully. Since I'm able to test rendering with a HTML template and assets, and all assets should be present at render time, my preference would be for Weasyprint to not handle any exceptions raised when calling the URL fetcher. For example, in https://github.com/Kozea/WeasyPrint/blob/b6fea501019e709f018a6d0b3d5223bc45011162/weasyprint/images.py#L300 the exception is swallowed and logged. This makes it difficult to track down issues within the URL fetcher (without a debugger attached), or even just missing assets, since those log messages tend to get lost in the noise of the other logs.

For the latter case, I could see the current behaviour being acceptable - in the browser if an external asset isn't found then you see a fallback font, image broken icon and so on.

I have a proposed solution that I would be willing to implement in a PR:

* The `HTML` class accepts a new argument, `url_fetcher_error_mode`, that is an enum-like string of the following values:
  * `suppress`: (default) Silence and log exceptions raised in a URL fetcher.
  * `suppress_http_errors`: Do not silence exceptions raised in a fetcher except for `HTTPError`.
    * TODO: would it be better to implement a custom error class?
  * `raise`: Do not silence any exceptions.
* The `URLFetchingError` class features a new attribute, `cause`, that contains the inner exception. The `try...except` block in `urls.fetch` will pass this in the constructor.
* The `url_fetcher` function passed to dependent classes like `SVGImage` is wrapped depending on the `url_fetcher_error_mode`.
  * `suppress` will catch all exceptions and log them.
  * `suppress_http_errors` will catch `HTTPError` only and log it.
  * `raise` will not catch anything (`url_fetcher` is passed as-is).
* The `try...except` blocks where the URL fetcher is invoked are removed.

These changes aim to break existing code as little as possible, and solve for both of the above use cases. In particular, the `url_fetcher_error_mode` enum will allow existing users to opt-in to this new behaviour, and also retain the existing behaviour while being able to more easily debug logic errors in custom fetchers.

[...]

Issue #351, opened 2016-08-26
Add a "strict mode" flag that makes any error fatal


liZe:
See #124.

liZe:
Fixed by #2609.

PR #2609, opened 2025-11-18
#2603 Refactor URL fetcher API

svbnet:
* Introduce URLFetcherResource class that fetchers can return
* Update implementation and doc on `default_url_fetcher`
* Update resource fetching code for HTML and other resource modules
* Update unit test to test compat with existing fetcher functions

### DIFF
--- a/docs/first_steps.rst
+++ b/docs/first_steps.rst
@@ -507,7 +507,7 @@ fetcher:
             if url.startswith('graph:'):
                 graph_data = [float(value) for value in url[6:].split(',')]
                 string = generate_graph(graph_data)
-                return URLFetcherResponse(url, string=string, mime_type='image/png')
+                return URLFetcherResponse(url, string, {'Content-Type': 'image/png'})
             return super().fetch(url, headers)
 
     source = '<img src="graph:42,10.3,87">'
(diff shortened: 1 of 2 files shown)


liZe:
Thanks for the pull request, we’ll take a look at it after version 67 is released (we’re already late…).

liZe:
@svbnet @Alex-Sichkar I’ve updated the pull request to use a class for URL fetchers. It gives much more possibilities for users that can now use inheritance to create their fetchers.

A new error type has been added to stop the rendering, as asked in #2603 (and #351).

Old tests have been restored as they were, to be sure that we didn’t break retrocompatibility. New tests have been added for the new fetcher class.

Examples have also been added to the documentation. Feedback is welcome!

jclab-joseph:
I was facing the same problem, thanks for posting the PR!
I will wait for approval.

liZe:
@svbnet @Alex-Sichkar There are new `--no-http-redirects` and `--fail-on-http-errors` available with both the CLI and the Python APIs. I’m happy with the current result, it’s ready to be merged for me.

liZe:
Will fix #2603 and #351.

PR #2612, opened 2025-11-22
Fix corner cases and clean block layout

liZe:
Fix #2595.

@michalfapso this should solve your use case, feedback is welcome!


michalfapso:
Thanks a lot, @liZe. Your fix works well for my minimal example, but it still fails on my full project. I'll try to create another minimal example where this issue is present.

Issue #2617, opened 2025-11-30
Children of flex and grid elements can't be relatively positioned.


different55:
Child elements of flex and grid parents can't be relatively positioned away from where the flex/grid algorithm stashes them.

Test page:

```html
<style>
  * {
      box-sizing: border-box;
      padding: 0;
      margin: 0;
  }
  .container {
    outline: 1px solid black;
    outline-offset: -1px;
    height: 50px;
    width: 100px;
  }
  .flex {
    display: flex;
  }
  .grid {
    display: grid;
    grid-template-columns: 1fr 1fr;
  }
  .child {
    width: 50px;
    height: 50px;
    background: red;
    position: relative;
    left: 25px;
  }
</style>
<div class="flex container">
  <div class="child"></div>
</div>
<div class="grid container">
  <div class="child"></div>
</div>
```

Expectation:

<img width="75" height="75" alt="Two red boxes are centered within black outlines." src="https://github.com/user-attachments/assets/3ed43341-cc0f-436f-97e3-cb35027d4048" />

Actual behavior:

<img width="75" height="75" alt="Two red boxes sit on the far left side within black outlines." src="https://github.com/user-attachments/assets/a2f5b443-79a6-4a15-8877-694b625e5a7f" />

I'm working on a fix but I'm new to this codebase and struggling to get my bearings. Patience is appreciated!

liZe:
Closed by #2618.

Issue #2400, opened 2025-03-06
Flex support enhancement


liZe:
This issue is for grouping flex features requests and bugs.

If you’d like to have a new feature regarding flex properties or encounter an issue related to flex, you can check here if an issue isn’t already opened. If not, please open an issue and we’ll link it there :)

- [ ] support `min-content` and `max-content` sizes
- [ ] handle `break-before` and `break-after` for flex row items
- [ ] fragmentation in rows and `break-inside: avoid` for flex items 
- [ ] handle flex items with intrinsic size (images)
- [ ] handle flex items with `content`
- [ ] define auto size of the flex container
- [ ] avoid useless layouts for flex items – #1166
- [ ] support `visibility: collapse`
- [ ] take care of `min-*` and `max-*` properties for the flex container’s cross size
- [ ] seriously support `display: inline-flex`

moreaki:
I opened a focused PR for the flex row fragmentation item:

- #2741 targets the #2440 case only. It keeps per-item resume state for flex rows so side-by-side row items fragment together across pages.

For #2674, I tried a distilled version of the shape described in the issue (`row` flex container, child flex container with `flex-direction: column` and `flex-wrap: wrap`), but that reduced case already behaves correctly on current `main`. I think a fuller reduced example is needed to catch the auto-size bug shown in the screenshot.

For #1166, I added some implementation ideas here: https://github.com/Kozea/WeasyPrint/issues/1166#issuecomment-4294957391. The short version is that a safe path probably starts with side-effect-free flex item size measurement, then optionally caches those measurement results, instead of caching full layout output.

moreaki:
I opened #2742 for a small part of the flex min/max topic: auto-height row flex containers now clamp their computed cross size with `min-height` / `max-height`, including the `align-content: stretch` interaction for wrapped rows.

The remaining implementation gap for this #2400 bullet is broader than that PR:

- Column flex cross size (`min-width` / `max-width`) still needs dedicated coverage, especially auto-width and wrapped-column cases.
- Inline flex cross-size clamping and baseline interactions are separate.
- Definite cross-size cases with conflicting min/max constraints need tests.
- Pagination may need separate handling when min/max cross-size changes the flex container height near a page break.
- Intrinsic keywords (`min-content`, `max-content`, `fit-content()`) are adjacent but should probably be handled separately.

So #2742 is deliberately a narrow row auto-height slice, not complete flex min/max support.

moreaki:
[...]

Issue #2145, opened 2024-05-02
Grid support enhancements


grewn0uille:
This issue is for grouping grid features requests and bugs.

If you’d like to have a new feature regarding grid properties or encounter an issue related to grid, you can check here if an issue isn’t already opened.
If not, please open an issue and we’ll link it there :)

- [ ] `display: inline-grid`
- [ ] auto content size for grid containers
- [ ] `grid-auto-flow: column` − #2142 
- [ ]  subgrids
- [ ] `repeat(auto-fill, *)` and `repeat(auto-fit, *)`
- [ ] `auto` margins for grid items
- [ ] `span` with line names
- [ ] `span` for flexible tracks
- [ ] `safe` and `unsafe` alignments
- [ ] baseline alignment
- [ ] grid items with intrinsic size (images)
- [ ] distribute space beyond limits
- [ ] grid items larger than grid containers
- [ ] `min-width`, `max-width`, `min-height`, `max-height` on grid items
- [ ] complex `min-content` and `max-content` cases
- [ ] absolutely positioned and floating grid items
- [ ] fragmentation in rows

charliesneath:
Great to see this list! One clarification to see if this relates to a problem I've noticed:

> auto content size for grid containers

Does this mean that, for example, rows within a `display: grid` container layout would automatically resize their height to fit the size of their content? If so, then I can file a more specific issue to detail the behavior I'm looking for because this doesn't work now.

(My usual process is to get something working in Chrome, and then compare to the output of Weasyprint; Chrome is more forgiving but I want to make sure I can point to specific issues where possible).

Thanks for all your work on this!

liZe:
> Does this mean that, for example, rows within a `display: grid` container layout would automatically resize their height to fit the size of their content?

No, it means for example that a grid can’t find the maximum width of its content. Here’s an example:
```html
<div style="float: left">
  <div style="display: grid; grid: 1fr / 1fr 1fr; border: 1px solid red">
    <div>abc def</div>
    <div>abc def</div>
  </div>
</div>
```

In this example, we have one row with two columns with "abc def", so its maximum width is the width of "abc defabc def". WeasyPrint can’t find this width yet, it assumes that it is the width of "abc def", just as if it was a block containing other blocks.

> If so, then I can file a more specific issue to detail the behavior I'm looking for because this doesn't work now.

Yes, you can open another issue, and give us an example so that we can reproduce the problem.

PR #2618, opened 2025-11-30
Calculate relative offsets for grid and flex children.

different55:
My attempt at fixing #2617.

It's working, but I'm not sure if there's a better place to have slipped in the calls to block.relative_positioning(). Both grid_layout() and flex_layout() are pretty long and I started getting a little lost when following along with them.

I also added a new pair of tests to test_position.py, one for grid and one for flex. I also tested it on my HTML Resume I've been trying to faithfully PDF-ify.

### DIFF
--- a/tests/layout/test_position.py
+++ b/tests/layout/test_position.py
@@ -1,5 +1,7 @@
 """Tests for position property."""
 
+import pytest
+
 from ..testing_utils import assert_no_logs, render_pages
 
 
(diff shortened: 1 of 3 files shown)


liZe:
Hi @different55,

Thanks a lot for this pull request! This is a bit embarrassing, but we may have totally forgotten relative boxes in these layouts…

> It's working, but I'm not sure if there's a better place to have slipped in the calls to block.relative_positioning(). Both grid_layout() and flex_layout() are pretty long and I started getting a little lost when following along with them.

Then it’s time for us to improve the readability of these modules. :smile:

It seems to be pretty safe to put these lines here, that’s great :heart:, but I think I’ll move them for consistency (if we ever want to deduplicate the duplicated code from `block_container_layout`). For regular blocks, we call the function after the absolute children layout, so I think that I’ll move there too for the flex layout. For the grid layout, we may have forgotten the absolute children too :cry:, so it may require a few more lines too.

> I also added a new pair of tests to test_position.py, one for grid and one for flex. I also tested it on my HTML Resume I've been trying to faithfully PDF-ify.

That’s so cool, thanks a lot!

liZe:
I’ve added tests for absolute children… and they fail. For the flex layout, we’re close to a clean solution, but for grid there’s a lot left to do. I’ve added list items in the corresponding meta issues (#2145 and #2400), let’s keep them for later!

Thanks a lot for the report and for the code.

Issue #2090, opened 2024-03-08
Temporary directory does not get removed


braggpeaks:
weasyprint apparently creates a temporary directory but does not remove the directory afterwards.

Windows folder location: C:\Users\daniel.kellner\AppData\Local\Temp
Subfolder naming example: weasyprint-0qjsm58z

The folders are empty but I believe they should be removed to not leave unnecessary folders (I currently have 8000 of these weasyprint-* folders, don't ask 😂).

liZe:
Hi!

The folders is supposed to be removed [here](https://github.com/Kozea/WeasyPrint/blob/4d93769a06e2a5a9d158daf2b825bbcff143ce4f/weasyprint/text/fonts.py#L256-L258), but:
- It is not guaranteed that `__del__()` methods are called for objects that still exist when the interpreter exits (quoting the Python documentation), and
- Windows likes to keep references to folders/files even when they’re not used anymore, preventing them to be deleted.

The "real" solution would be to only use data in memory (fonts are not that big). It should be possible to keep config files in memory thanks to [`FcConfigParseAndLoadFromMemory`](https://fontconfig.pages.freedesktop.org/fontconfig/fontconfig-devel/fcconfigparseandloadfrommemory.html) but there’s nothing to replace `FcConfigAppFontAddFile`.

If someone finds a better solution, I’ll be happy to try!

seidnerj:
This is pretty annoying, it is cluttering the /tmp folder. This could also be the case if there was an errors when calling rmtree (as ignore_errors is set tot True).

liZe:
> This is pretty annoying, it is cluttering the /tmp folder.

We now use `FcConfigParseAndLoadFromMemory`, and I’ve just added a commit that avoids creating useless temporary folders. It means that you’ll only have the folders if you use `@font-face` rules, but the problem is not solved.

Unfortunately, there’s no function in Fontconfig to add fonts from memory (see almost [11-year old Fontconfig issue](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/issues/12)). Without such a function, as explained above, I didn’t find a reliable solution, ideas are welcome.

> This could also be the case if there was an errors when calling rmtree (as ignore_errors is set tot True).

As far as I can remember, Windows was randomly trying hard to keep references to this folder, preventing `rmtree` from removing the folder.

Again, the real solution would be to avoid temporary folders, and it would require a new function in Fontconfig.

seidnerj:
[...]

PR #2625, opened 2025-12-09
Add public method to cleanup font configuration temporary directory

max-blau:
The temp directory used store some downloaded bytes for fonts is sometimes not being cleaned up automatically via the `__del__` dunder method. This commit exposes a `cleanup` method to allow users to handle this cleanup themselves if it affects their workflow.

tests:
```
➜ ~/work/fork/WeasyPrint (expose-font-cleanup-method) [weasyprint] $ pytest tests/test_fonts.py
================================================================================================================ test session starts ================================================================================================================
platform darwin -- Python 3.10.18, pytest-9.0.2, pluggy-1.6.0
rootdir: /Users/maxblau/work/fork/WeasyPrint
configfile: pyproject.toml
collected 8 items

tests/test_fonts.py ........                                                                                                                                                                                                                  [100%]

================================================================================================================= 8 passed in 0.14s =================================================================================================================
```

### DIFF
--- a/tests/test_fonts.py
+++ b/tests/test_fonts.py
@@ -1,5 +1,5 @@
 """Test the fonts features."""
-
+from weasyprint.text.fonts import FontConfiguration
 from .testing_utils import assert_no_logs, render_pages
 
 
@@ -166,3 +166,12 @@ def test_woff_simple():
     assert span1.width == span3.width
     # the default font does not match the loaded fonts
     assert span1.width != span4.width
+
+
+@assert_no_logs
+def test_font_cleanup():
+    font_config = FontConfiguration()
+    font_folder = font_config.font_folder
+    assert font_folder.exists()
+    font_config.cleanup()
+    assert not font_folder.exists()
--- a/weasyprint/text/fonts.py
+++ b/weasyprint/text/fonts.py
@@ -112,15 +112,28 @@ class FontConfiguration:
         self.strut_layouts = {}
         self.font_features = {}
 
+    @property
+    def font_folder(self) -> Path:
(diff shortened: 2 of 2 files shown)


max-blau:
@liZe I don't have permissions to add reviewers, here's a PR for the workaround we discussed in https://github.com/Kozea/WeasyPrint/issues/2090

liZe:
@max-blau Why did you close this pull request?

max-blau:
@liZe sorry, meant to post a comment.
It's not needed - you can just do something like this instead:
```
font_config = FontConfiguration()
del font_config
```

max-blau:
@liZe nevermind - probably due to other implementation details in `weasyprint`, just subtracting the ref count of the `FontConfiguration` instance isn't enough to get the garbage collector to delete it.
Re-opening this - although its also possible to just call `__del__` directly as a user - if thats OK with you, then this change could be dropped.

liZe:
Hi!

Thanks again for your pull request.

> Re-opening this - although its also possible to just call `__del__` directly as a user - if thats OK with you, then this change could be dropped.

I’ve spent some time trying to play with your pull request, and… I think you’re right.

No matter how hard we will try, there’s no reliable way to remove this folder because of how Fontconfig and Windows work, so even your test may fail if we actually use `@font-face` fonts in it.

And we don’t want to support and maintain public API that should be an implementation detail.

If you’re not on Windows, calling `__del__` should reliably work, and that’s the best workaround in my opinion.

Sorry for taking so much time to come to this conclusion too :pray:, but I guess we had to go through these questions.

PR #2627, opened 2025-12-10
Add font-face support to SVG

jurriaan:
Not sure if this is how it should be implemented, but this fixes the issue where `@font-face` declarations within a SVG `<style>` where not parsed (and thus the font was not available for the SVG).

It does add the font to the font_config, which means the font will be available to the rest of the document, however this seems to be the case in browsers as well.

### DIFF
--- a/weasyprint/svg/__init__.py
+++ b/weasyprint/svg/__init__.py
@@ -839,7 +839,7 @@ class SVG:
 class Pattern(SVG):
     """SVG node applied as a pattern."""
     def __init__(self, tree, svg):
-        super().__init__(tree._etree_node, svg.url, svg.font_config)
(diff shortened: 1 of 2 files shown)


liZe:
Hi!

Thanks a lot for the pull request, it’s clean! I’ll just add some tests and clean a few things (we should keep the URL fetcher from `__init__` to use it in `draw`, for example) before merging.

jurriaan:
> Hi!
> 
> Thanks a lot for the pull request, it’s clean! I’ll just add some tests and clean a few things (we should keep the URL fetcher from `__init__` to use it in `draw`, for example) before merging.

@liZe Ah you're right, no need to pass the `url_fetcher` in the `draw` as well, cleaned it up and added some tests (that fail on main)

liZe:
That’s perfect, thanks a lot!

PR #2638, opened 2025-12-26
Update Python command for Windows installation steps

aoi2010:


liZe:
Hello,

Why do you need this change? Python from the store does install the `python3` command, doesn’t it?

aoishik:
Most of the user uses the one downloaded from https://www.python.org/ and it has python and py command and not python3 command. I think Microsoft Store one also uses the python command only.

liZe:
Well, why not. And let’s propose the Python Install Manager, as it’s the new way given in Python’s doc.

liZe:
Thank you!

PR #2640, opened 2025-12-29
PDF/A compability

dedalusMohantyMa:
I checked again with verapdf and it seems we do not really need subject but description to exist.
Changing how the xml elements are created should be sufficient.

### DIFF
--- a/weasyprint/urls.py
+++ b/weasyprint/urls.py
@@ -1,6 +1,5 @@
 """Various utility functions and classes for URL management."""
 
-import codecs
 import contextlib
 import os.path
 import re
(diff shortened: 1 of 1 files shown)


liZe:
Hi!

We don’t have problems with [our tests launched with VeraPDF](https://github.com/Kozea/WeasyPrint/actions/workflows/test_pdfa.yml), even with `subject` set instead of `description`. Adobe Acrobat seems to be OK too.

Do you have a specific problem you’d like to solve?

dedalusMohantyMa:
> Hi!
> 
> Thank you for this pull request.
> 
> We don’t have problems with [our tests launched with VeraPDF](https://github.com/Kozea/WeasyPrint/actions/workflows/test_pdfa.yml), even with `subject` set instead of `description`. Adobe Acrobat seems to be OK too.
> 
> Do you have a specific problem you’d like to solve?

Hi, I will retest it. I think it got solved by you guys already ^^

PR #2646, opened 2026-01-05
SVG: Add font shorthand support for text elements (fix #2255)

### LINKED ISSUE #2255: 'font' CSS shorthand property ignored in SVG 'text' elements
The font properties set via a 'font' CSS shorthand property seem to be totally ignored when rendering text elements inside SVG images.

Here is a minimal SVG to reproduce the issue:
```
<svg xmlns="http://www.w3.org/2000/svg"  viewBox="0 0 400 400">

<style>
   .octl text {
         font: cursive;
          fill: purple;
          stroke: white;
          stroke-width: 0.1mm;
          stroke-linecap: butt;
          stroke-linejoin: miter;
   }

   	line {
	    stroke: black;
	}
</style>
<g class="octl">
<text class="octl" x="7mm" y="7mm">Text 14mm in cursive.</text>
<line x1="7mm" y1="7mm" x2="7mm" y2="0mm" />
<line x1="7mm" y1="10mm" x2="14mm" y2="10mm" />
</g>
</svg>
```

It renders well in the PDF from Firefox:

screenshot_firefox

But Weasyprint uses default font family and size:

screenshot_weasyprint

font1
[font1-firefox.svg.pdf](https://github.com/user-attachments/files/17056914/font1-firefox.svg.pdf)
[font1-weasyprint.svg.pdf](https://github.com/user-attachments/files/17056916/font1-weasyprint.svg.pdf)

a-usatenko:
## Description
Implements minimal support for SVG `font` shorthand property as per SVG 1.1 specification.

## Changes
- Added `_parse_simple_font_shorthand` method to `Node` class
- Supports basic `<font-size> <font-family>` format
- Handles common CSS units (px, em, rem, pt, %, mm, cm)
- Returns empty dict for complex/invalid shorthand values
- Added comprehensive unit and integration tests

## Testing
- Unit tests for parser edge cases
- Integration tests for rendering and inheritance  
- Follows existing test patterns
- No regression in existing functionality

## Notes
- This is a temporary minimal implementation
- Complex CSS font shorthand (with style/weight/line-height) is ignored

Fixes #2255


liZe:
The overall logic is cool, thanks a lot! :heart: 

I’ll change a few things:
- separate the expander and `style` attribute logic, so that we can set `font` as an XML attribute too, :heart_eyes: 
- let the validators validate the input (and don’t duplicate the units detection for example),
- simplify the code and remove the separate function (as it’s only 4 lines without the validation),
- merge the unit and draw tests, now that we don’t handle units in our code.

If you’re OK with these changes, it’s time to merge!

a-usatenko:
@liZe Thank you for the review and improvements! 

I really like how you simplified the code — using `.pop()` and moving the logic directly into cascade is much cleaner than my separate function.

The test integration into test_text.py also makes perfect sense.

I'm excited to see this merged for v68.0! 🚀

Issue #1698, opened 2022-07-31
Add support for CSS Shape Module


citorva:
This module allows a designer to align text with a custom shape or path and set a space between the object and the text.
This module is defined as a Candidate Release [here](https://www.w3.org/TR/css-shapes/)
It provides three properties:
- [`shape-outside`](https://www.w3.org/TR/css-shapes/#propdef-shape-outside): The shape that the text area should follow
- [`shape-margin`](https://www.w3.org/TR/css-shapes/#propdef-shape-margin): The margin between the shape’s border and the text area.
- [`shape-image-threshold`](https://www.w3.org/TR/css-shapes/#propdef-shape-image-threshold): The alpha channel threshold used to extract the shape using an image. A value of 0.5 means that the shape will enclose all the pixels that are more than 50% opaque.

liZe:
That would be a nice feature to have!

itlackey:
+1 for this feature!

itlackey:
This is the biggest missing feature and what is keeping browser based tools as my go to. If there is guidance on how to achieve similar behavior, please point me in the right direction. I would love to switch over but need text wrapping around image shapes.

liZe:
> If there is guidance on how to achieve similar behavior, please point me in the right direction. I would love to switch over but need text wrapping around image shapes.

Thanks for your help!

This issue is really, really, really complex. You’ll need solid knowledge in Python, CSS, geometry. And probably a few weeks. 😁

The first step is to support the CSS properties validation. This is done by [adding validators](https://github.com/Kozea/WeasyPrint/blob/main/weasyprint/css/validation/properties.py) and [initial values](https://github.com/Kozea/WeasyPrint/blob/main/weasyprint/css/properties.py). `shape-margin` and `shape-image-threshold` are straightforward, but `shape-outside` is… well… a nightmare.

You then have to add [computed values](https://github.com/Kozea/WeasyPrint/blob/main/weasyprint/css/computed_values.py).

When this is done, the management of floats during the layout has to be adapted to handle shapes.

[...]

PR #2668, opened 2026-01-21
Shape outside support

itlackey:
* #1698 

## Add CSS Shapes Level 1 documentation to API reference

Documents the newly implemented shape-outside feature including:
- Box keywords (margin-box, border-box, padding-box, content-box)
- Shape functions (circle, ellipse, polygon, inset)
- Shape-margin property and reference box combinations
- Example CSS usage
- List of unsupported features (shape-image-threshold, url(), path())

## Enhance shape boundary handling for text wrapping and border-radius support

- Update `test_inset_with_margin_text_wrap` to clarify shape boundary usage.
- Improve `avoid_collisions` to sample bounds for curved shapes at multiple Y positions.
- Modify `_create_base_boundary` to create `InsetBoundary` for elements with border-radius.
- Add `_get_border_radii` function to extract border-radius values for shape calculations.
- Add shape-image-threshold property and implement ImageBoundary class for alpha channel extraction
- Add CSS Shapes Level 1 features and documentation to changelog

## Tests: 117 tests

- Fix critical edge cases in shape boundary classes
- CircleBoundary: Return None for zero/negative radius
- EllipseBoundary: Return None for zero/negative rx or ry
- PolygonBoundary: Mark polygons with < 3 points as degenerate
- InsetBoundary: Guard sqrt operations with max(0, ...) for floating-point precision


itlackey:
Pushed a branch with a new book example I used for testing: https://github.com/itlackey/weasyprint-samples/tree/dev/shape-outside

liZe (changes requested):
Thanks for the PR, it’s pretty cool!

I like the choice to avoid `path()` (and `shape()`?). Your comment and the documentaton include `shape-image-threshold` and `url()`, but they’re supported, right? I think that SVGs are also unsupported, and that’s a good idea if we don’t want to actually spend weeks on that. 😄 That’s a good way to cover most of the common use cases fast.

Many comments are just small details.

The main problem for me is the boundary detection logic. I think that it would be better to let the box handle this complexity (and have not much in float), and that the box API should give the max boundaries for a range instead of a single y value. Do you agree with that?

For the tests, it may be interesting to dispatch them into different files, and to add drawing tests. If you don’t want to spend more time on them, tell me, I’ll do that later.

I’ll review `shapes.py` and the parsing details later.

itlackey:
Just checking in. I found a memory/performance regression when working on large documents with high resolution images. I hope to get back to debugging this weekend.

*I will also address the comments/issues highlighted above

liZe:
Hi @itlackey, did you find some time to take a look at the comments?

grewn0uille:
Hi,

I close that PR for now. We’ll reopen it if you find time to take a look at it!

PR #2671, opened 2026-01-27
Improve API compatibility between URLFetcherResponse and addinfourl

liZe:
Adding deprecated functions and using an integer for `status` in `URLFetcherResponse` increases the compatibility with `urllib.response.addinfourl` and external code that depends on that.


jurriaan:
Thanks a lot!

PR #2681, opened 2026-02-11
Add dc:description to pdf/a meta data

dedalusMohantyMa:
Bildschirmfoto vom 2026-02-11 16-30-36

As you can see the current version is not compliant as apparently `dc:description` is missing.
I added this to the current implementation, so that it is not missing anymore and veraPDF does not complain.


liZe:
@dedalusMohantyMa Hi!

Thanks for the pull request, we finally could reproduce the problem.

Do you know if we could just replace `subject/Bag` by `description/Alt`, instead of using both? It seems to work for PDF/A validation, and the field description in the XMP specification seems to match what we want ("textual description of the content of the resource").

liZe:
I’ve replaced the old field by the new one. Feel free to add a comment if there’s something wrong!

Issue #412, opened 2017-01-18
Weasyprint logging is missconfigured


PamelaM:
From the python docs https://docs.python.org/2/howto/logging.html#library-config : 
```
Note It is strongly advised that you do not add any handlers other than NullHandler to your 
library’s loggers. This is because the configuration of handlers is the prerogative of the 
application developer who uses your library. The application developer knows their target 
audience and what handlers are most appropriate for their application: if you add handlers 
‘under the hood’, you might well interfere with their ability to carry out unit tests and 
deliver logs which suit their requirements.
```

The default handler should be the NullHandler.

liZe:
I agree. Silent logs have been asked before (at least #300 and #312) and that should be the default when WeasyPrint used as a library. We should log to stderr only when it's the "normal" way to handle logs, i.e. when WeasyPrint is launched from the CLI.

PamelaM:
For those hitting this problem before it's fixed, here's how I've disabled the weasyprint logging handler - the logger needs to be set up before the first import of weasyprint

```
import logging

logger = logging.getLogger("weasyprint")
if not logger.handlers:
    logger.addHandler(logging.NullHandler())

from weasyprint import HTML
```

erichonkanen:
possibly related, I think all of the `WARNING` messages are coming from this package?

![image](https://cloud.githubusercontent.com/assets/230757/24126902/f8e7956a-0d8d-11e7-895f-3ee79d24ccb3.png)

alacret:
This hack doesn't work for Django applications. Neither in the settings.py nor before the import of the library

oleh-smediuk:
For those who use Django internal LOGGING dict-like config:

```python
LOGGING = {
    'version': 1,
    ...
    'handlers': {
        ...
        'null_handler': {
            'level': 'ERROR',
            'class': 'logging.NullHandler',
        },
    },
    'loggers': {
        ...
        'weasyprint': {
            'level': 'ERROR',
            'handlers': ['null_handler'],
        },
    }
}
```

conformist-mw:
For those who use Django internal LOGGING dict-like config:

```python
LOGGING = {
    'version': 1,
    ...
    'handlers': {
        ...
        'null_handler': {
            'level': 'ERROR',
            'class': 'logging.NullHandler',
        },
    },
    'loggers': {
        ...
        'weasyprint': {
            'level': 'ERROR',
            'handlers': ['null_handler'],
        },
    }
}
```

[...]

PR #2704, opened 2026-03-16
Removed fontTools logging capture

arjenzorgdoc:
https://github.com/Kozea/WeasyPrint/issues/412#issuecomment-2830795133

> Remove capture_logs function completely [...]

and

https://github.com/Kozea/WeasyPrint/issues/412#issuecomment-2830940404

> Could you please open a PR so that we can discuss this topic over there?

### DIFF
--- a/weasyprint/pdf/fonts.py
+++ b/weasyprint/pdf/fonts.py
@@ -3,7 +3,6 @@
 import io
 import re
 from hashlib import md5
-from logging import WARNING
 from math import ceil
 
 import pydyf
@@ -11,7 +10,7 @@ from fontTools import subset
 from fontTools.ttLib import TTFont, TTLibError, ttFont
 from fontTools.varLib.instancer import instantiateVariableFont
 
-from ..logger import LOGGER, capture_logs
+from ..logger import LOGGER
 from ..text.constants import PANGO_STRETCH_PERCENT
 from ..text.ffi import FROM_UNITS, ffi, harfbuzz, harfbuzz_subset, pango
 from ..text.fonts import get_hb_object_data, get_pango_font_hb_face
@@ -280,12 +279,7 @@ class Font:
         # Subset font.
         try:
             ttfont = TTFont(full_font, fontNumber=self.index)
-            with capture_logs('fontTools', level=WARNING) as logs:
-                subsetter.subset(ttfont)
-            for log in logs:
-                LOGGER.warning(
-                    'fontTools warning when subsetting '
-                    f'"{self.family}": {log}')
+            subsetter.subset(ttfont)
         except TTLibError as exception:
             LOGGER.warning(f'Unable to subset "{self.family}" with fontTools')
             LOGGER.debug('Original exception:', exc_info=exception)


liZe:
Hi!

Thanks for this pull request, that’s a better place than the original issue to discuss about this topic!

The problem we have is that users complain when they get fontTools’ logs: the current code has been added to fix #1585, where different people complained about these logs.

WeasyPrint decided some time ago not to display logs to stderr when used as a library (that’s the original topic of #412). I think that it’s a good idea. But fontTools writes these messages by default. So, either we’re inconsistent and write fontTools logs while hiding WeasyPrint logs, or we capture fontTools’ logs.

The [solution](https://github.com/fonttools/fonttools/issues/3538#issuecomment-2135700552) proposed by the fonttools devs was:

> you can register a custom handler that intercepts only the messages which come from loggers whose name starts with "fontTools", and then do whatever you want with them

We had actually come to the same solution without their advice. 😁 It’s also clear that these logs come from fontTools, and not WeasyPrint directly. But maybe the solution was to ask Fonttools devs not to display logs by default.

Merging your pull request would lead back to the #1585 complains, so we won’t merge it right now, but we can try to find another solution. What’s your current use case, and what does the logging capture break for you?

arjenzorgdoc:
My problem is this:

I want to use python built in logging simplified to this:

```python
import logging

import weasyprint

handler = logging.StreamHandler()
logging.basicConfig(handlers=[handler], level="INFO")


weasyprint.HTML("https://weasyprint.org/").write_pdf("./weasyprint-website.pdf")
```

```
$ uv run --with weasyprint huu.py 
WARNING:weasyprint:Ignored `aspect-ratio: 1.5` at 293:3, unknown property.
WARNING:weasyprint:Ignored `filter: grayscale(10%)` at 358:3, unknown property.
DEBUG:fontTools.ttLib.ttFont:Reading 'maxp' table from disk
DEBUG:fontTools.ttLib.ttFont:Decompiling 'maxp' table
DEBUG:fontTools.subset.timer:Took 0.000s to load 'maxp'
DEBUG:fontTools.subset.timer:Took 0.000s to prune 'maxp'
INFO:fontTools.subset:maxp pruned
DEBUG:fontTools.ttLib.ttFont:Reading 'cmap' table from disk
DEBUG:fontTools.ttLib.ttFont:Decompiling 'cmap' table
DEBUG:fontTools.ttLib.ttFont:Reading 'post' table from disk
...
```

I get all sorts of fontTools logs I didn't ask for.

With this patch I get fewer:

```
$ python huu.py 
WARNING:weasyprint:Ignored `aspect-ratio: 1.5` at 293:3, unknown property.
WARNING:weasyprint:Ignored `filter: grayscale(10%)` at 358:3, unknown property.
INFO:fontTools.subset:maxp pruned
INFO:fontTools.subset:cmap pruned
INFO:fontTools.subset:fpgm dropped
INFO:fontTools.subset:prep dropped
INFO:fontTools.subset:cvt  dropped
...
```

And now I can filter it:

```python
handler = logging.StreamHandler()
logging.basicConfig(handlers=[handler], level="INFO")
logging.getLogger("fontTools").setLevel(logging.WARNING)
```

```
$ python huu.py 
WARNING:weasyprint:Ignored `aspect-ratio: 1.5` at 293:3, unknown property.
WARNING:weasyprint:Ignored `filter: grayscale(10%)` at 358:3, unknown property.
$
```

That doesn't work at all without this patch.

BTW: when I don't configure logging at all, it doesn't log anything with or without this patch.

arjenzorgdoc:
BTW: this whole capturing the log idea is not something that can be done in a thread safe way. The logger config is a global state. The are always side effects.

liZe:
> BTW: this whole capturing the log idea is not something that can be done in a thread safe way. The logger config is a global state. The are always side effects.

Of course, that’s just a dirty workaround originally created for tests. Hopefully, pytest-xdist spawns processes…

We could put the function back to `testing_utils.py` where it belongs, if we remove the call for fontTools.

> My problem is this: […]

Thanks a lot for sharing this. That’s definitely a problem.

> BTW: when I don't configure logging at all, it doesn't log anything with or without this patch.

They are printed, that’s Python’s default behavior when nothing is configured. You don’t see anything because the default level only displays warnings. Unfortunately, many fonts have minor problems that log warnings, the timestamp problem from #1585 is a common one.

WeasyPrint avoids this with:
https://github.com/Kozea/WeasyPrint/blob/c3b780ec328fe58528b57240b31fe638881aec23/weasyprint/logger.py#L19-L22

Best practice from the [Python logging tutorial](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) is a bit hard to decipher:
- "It is strongly advised that you do not add any handlers other than [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) to your library’s loggers."
- "[…] events of severity WARNING and greater will be printed to sys.stderr. This is regarded as the best default behaviour."

I’m convinced by the first part, but not the second part.

From there, I see two solutions:
- Ask the fontTools devs to hide logs by default, just like WeasyPrint. I’m not sure that it’s the best idea, as there’s no strong consensus about what logs should do by default, but why not.
- Apply the code adding the default null handler to the fontTools logger when WeasyPrint’s logging module is imported.

What do you think?

septatrix:
My thoughts:

- For library usage it should never ever touch another packages logging config. If I want logs I should get them. If they annoy me, I, as the developer, can manually change that packages logging level.
- For CLI usage, you are free to override logs how you see fit. E.g. set the logging level to `ERROR` on the root logger.

Regarding the code you linked I even think that it might be wrong, too. At least the `setLevel` call. If I enable logging I should get all logs, and configure the level myself. That's not the job of a library

liZe:
Thanks a lot.

> * For library usage it should never ever touch another packages logging config. If I want logs I should get them. If they annoy me, I, as the developer, can manually change that packages logging level.

I understand, and I agree with the theory.

The problem is that WeasyPrint users complain when they get fontTools error messages printed on their terminal when they use WeasyPrint as a library. That’s what happens if we don’t touch fontTools’ logging config. Users can’t even know that the messages come from fontTools, so they blame WeasyPrint.

So, what can we do to avoid issues like #1585?

> * For CLI usage, you are free to override logs how you see fit. E.g. set the logging level to `ERROR` on the root logger.

I agree. With the `--debug` option at least, we could set the level of the root logger in `__main__.py`, and add the name of the logger in the formatter.

> Regarding the code you linked I even think that it might be wrong, too. At least the `setLevel` call. If I enable logging I should get all logs, and configure the level myself. That's not the job of a library

You’re right. The line setting the level has been added in 2017 to work around a bug in a new version of Pytest, it now seems useless and can be removed.

septatrix:
I made a general cleanup in #2711

liZe:
Replaced by #2711.

PR #2710, opened 2026-03-18
Try system mimetype databases as a fallback

septatrix:
Follow-up for #2708

I hope the comment is sufficiently clear. I was not really sure as to what terminology to use...


liZe:
Thank you!

PR #2713, opened 2026-03-24
Remove fontTools subsetter

liZe:


liZe:
For the record: we can’t do that. Harfbuzz-Subset is not installed by default on GitHub Actions’ Ubuntu, and the library seems to be packaged differently depending on the Ubuntu version. We’ll keep this fallback for some time, and add a deprecation warning to get feedback.

liZe:
Ubuntu 22.04 (supported until April 2027) and Debian 11 (supported until August 2026) don’t seem to package Harfbuzz-Subset.

Note that uharfbuzz does provide mapping for Subset.

PR #2716, opened 2026-03-28
Fix PDF text extraction for outside list markers

### LINKED ISSUE #2715: Outside list markers are emitted as separate PDF text runs, producing stray bullet-only spans in extraction
## Summary

WeasyPrint emits native outside list markers (`ul/li`) as separate PDF text runs from the corresponding list item text. As a result, PDF text extractors such as PyMuPDF see standalone bullet spans like `• ` instead of a single extracted line such as `• One short bullet`.  Visually the PDF looks fine, but when it's interpreted by tooling it finds empty bullets.  I'm running into this when I get AI to inspect documents, it keeps pointing out these problems to me.  Which means other consumers of these PDF's would also get confused.

This is not just a styling quirk. It changes the PDF text layer and can affect downstream extraction, indexing, review tools, and any workflow that expects list markers to stay associated with their list item text.

## Status

This behavior was reproduced inside the WeasyPrint repo and traced to PDF drawing / stacking behavior for outside list markers, not to CSS parsing or HTML structure.

A repo-level fix was validated:

- native list markers were being painted as separate positioned text runs
- after the fix, each bullet and its item text are emitted in the same PDF text object for a given line
- PyMuPDF then extracts:
  - `• One short bullet`
  - `• Two short bullet`
  - `• Three short bullet`

## Environment

- OS: Windows
- WeasyPrint 68.1
- Also reproduced with WeasyPrint 67.0
- PDF inspection tool: PyMuPDF

## Minimal Reproduction

HTML:

```html
<!doctype html>
<html lang="en">
  <body>
    <ul>
      <li>One short bullet</li>
      <li>Two short bullet</li>
      <li>Three short bullet</li>
    </ul>
  </body>
</html>
```

Python:

```python
from weasyprint import HTML

HTML(filename="repro.html").write_pdf("repro.pdf")
```

## Inspection Method

I inspected the generated PDF with PyMuPDF using:

```python
page.get_text("text")
page.get_text("dict")
```

and checked for spans whose extracted text was only `•` or `• `.

## Actual Result

PyMuPDF extracts the list item text and the bullets separately.

Example:

```text
One short bullet
Two short bullet
Three short bullet
• 
• 
• 
```

Example extracted spans:

```text
SPAN 'One short bullet' (...)
SPAN 'Two short bullet' (...)
SPAN 'Three short bullet' (...)
SPAN '• ' (...)
SPAN '• ' (...)
SPAN '• ' (...)
```

## Expected Result

Native list rendering should keep each bullet associated with its corresponding list item text in the PDF text layer.

For example, extraction should look like:

```text
• One short bullet
• Two short bullet
• Three short bullet
```

or at minimum should not produce stray standalone bullet-only spans for ordinary semantic lists.

## Root Cause

This appears to happen specifically for outside list markers.

Inside WeasyPrint, the `li::marker` bullet is built as an outside marker box and then treated as a separately positioned drawn object. That causes the marker text to be emitted in a different PDF text run from the list item text, even though visually they belong to the same line.

In the WeasyPrint repo, the issue was fixed by changing the stacking/drawing behavior for outside markers so they are no longer split into a separate positioned stacking path during PDF emission.

## Why This Matters

- It pollutes the PDF text layer.
- It makes extraction inconsistent with the visual reading order.
- It can make native semantic lists look broken to downstream tools.
- It pushes users toward non-semantic workarounds like manually writing `• text`.

## Workaround

As a workaround, rendering bullets as a single inline text run such as `• Item text` avoids the problem, because the bullet and text are emitted together in the same extracted text run.

## Suggested Regression Test

A useful regression test is:

- render a minimal `ul/li` document to PDF
- inspect the emitted PDF text stream or extracted text structure
- assert that marker and item text are not emitted as separate text objects for the same list line

## Solution
I've got a proposed solution that I will put a PR out there for.

AndyBevan:
Fixes https://github.com/Kozea/WeasyPrint/issues/2715

Outside list markers were being emitted as separate positioned text runs, which caused extractors like PyMuPDF to see stray bullet-only spans.

Avoid treating outside markers as positioned stacking contexts during PDF painting, and add regression coverage for list marker text-object emission.

## Summary

This fixes PDF text extraction for outside list markers.

Previously, native ul/li markers could be emitted as separate positioned PDF text runs from their corresponding list item text. In practice, tools like PyMuPDF could then extract stray bullet-only spans such
  as •  instead of a single line like • One short bullet.

This change keeps outside list markers out of the separate positioned stacking-context path during painting, so marker and item text are emitted together in the same PDF text object for a given line.

## What Changed

  - Update C:/DevGit/AndyBevan/Github/WeasyPrint/weasyprint/stacking.py so is_outside_marker boxes are not treated as positioned stacking contexts for painting.
  - Add regression coverage in C:/DevGit/AndyBevan/Github/WeasyPrint/tests/test_pdf.py for:
      - basic outside list markers
      - nested lists
      - RTL lists
      - lists spanning multiple pages

  ## Why This Fix

  The issue was not in CSS parsing or marker box creation. The marker boxes were laid out correctly, but during painting they were handled as separate positioned content, which split marker and item text into
  different PDF text runs.

  By keeping outside markers in normal paint order, the emitted PDF stream no longer splits the bullet from the line’s text, and extraction matches the visual reading order.

  ## Verification

  Focused checks:

  - tests/test_pdf.py -k list_markers
  - tests/draw/test_list.py

  Full suite on this machine:

  - 4022 passed
  - 6 failed
  - 41 xfailed

 The remaining 6 failures are pre-existing local platform/font pixel-test failures unrelated to this change:

  - C:/DevGit/AndyBevan/Github/WeasyPrint/tests/draw/test_text.py
  - C:/DevGit/AndyBevan/Github/WeasyPrint/tests/test_acid2.py

  ## Issue

  Fixes #2715

### DIFF
--- a/tests/test_pdf.py
+++ b/tests/test_pdf.py
@@ -1,5 +1,6 @@
 """Test PDF-related code, including metadata, bookmarks and hyperlinks."""
 
+from collections import defaultdict
 import hashlib
 import io
 import re
@@ -9,6 +10,7 @@ import pytest
 
 from weasyprint import Attachment
 from weasyprint.document import Document, DocumentMetadata
+from weasyprint.pdf.stream import Stream
 from weasyprint.text.fonts import FontConfiguration
 from weasyprint.urls import path2url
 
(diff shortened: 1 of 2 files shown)


liZe:
Closing, as we don’t want to merge the stacking contexts.

PR #2720, opened 2026-04-04
Parse HTML integer attributes per spec, ignoring trailing non-digits

### LINKED ISSUE #2636: Ignore text after digits when parsing HTML integers
Yeah, I know ! It's not valid HTML but it has worked "forever" in browsers.
My understanding is that browsers actually coerce colspan="100%" to colspan="100" ?
Works for me !
Thank you ! :-)

danfitz36:
## Summary

- Add `parse_html_integer()` utility following the [HTML spec rules for parsing integers](https://html.spec.whatwg.org/#rules-for-parsing-integers)
- Replace `int()` calls with `parse_html_integer()` in all HTML integer attribute parsing locations: `colspan`, `rowspan`, `col`/`colgroup` `span`, `<font>` `size`, and `<hr>` `size`
- Fixes values like `colspan="100%"` which browsers parse as `100` but WeasyPrint previously rejected

## Context

Browsers ignore non-digit characters after digits when parsing HTML integer attributes (e.g. `colspan="100%"` is treated as `colspan="100"`). WeasyPrint was using Python's `int()` which rejected such values entirely, causing layout differences vs browsers.

The maintainer outlined the fix in issue comments: write a function following the [integer parsing rules](https://html.spec.whatwg.org/#rules-for-parsing-integers) and use it everywhere HTML integer attributes are parsed.

### Call sites updated

- `TableCellBox.__init__` — `colspan`, `rowspan`
- `TableColumnBox.span` property
- `TableColumnGroupBox.span` property
- `find_style_attributes` — `<font size>` and `<hr size>`

## Test plan

- [x] Unit tests for `parse_html_integer`: basic values, signs, whitespace, trailing non-digits, error cases, non-ASCII digits
- [x] Integration tests for `colspan="100%"`, `rowspan="2px"`, `span="2px"` in table layout
- [x] Enhanced existing `test_table_bad_int_td_th_span` with explicit value assertions
- [x] All existing tests pass (2690 passed, 0 regressions)
- [x] `ruff check` passes

Fixes #2636


danfitz36:
I used weasyprint a few years ago at a previous job to create insurance documents, quotes, policies, etc. I now work at a new company and we're using weasyprint again to generate estimates/invoices for builders, remodelers, and contractors. we'll want to start contributing when we can, so this is hopefully the first of many PRs, but I'll wait for a review and any comments before starting on anything new.

weasyprint/html_utils.py
```diff
@@ -0,0 +1,43 @@
+"""Utility functions for HTML microsyntax parsing."""
```
liZe: We should put this in `weasyprint/html.py` instead. You’ll find that `HTML_WHITESPACE` is already defined!

tests/test_html_utils.py
```diff
@@ -0,0 +1,48 @@
+"""Tests for HTML microsyntax parsing utilities."""
```
liZe: I think that we can put these tests in `test_presentational_hints.py` and rename it to `test_html.py` for consistency with the `weasyprint` folder.

tests/test_html_utils.py
```diff
@@ -0,0 +1,48 @@
...
+
+from weasyprint.html_utils import parse_html_integer
+
+
+def test_parse_html_integer_basic():
```
liZe: For this kind of tests, we prefer to use `@pytest.mark.parametrize` so that we have independent tests, with a function for right values and a function for wrong values. You can take a look at `tests/css/test_validation.py` that has a lot of tests like this.

weasyprint/formatting_structure/boxes.py
```diff
@@ -724,14 +721,10 @@ def __init__(self, *args, **kwargs):
...
-        except (AttributeError, ValueError):
-            self.colspan = 1
-        try:
-            self.rowspan = max(int(self.element.get('rowspan', '').strip()), 0)
-        except (AttributeError, ValueError):
-            self.rowspan = 1
+        colspan = parse_html_integer(self.element.get('colspan', ''))
+        self.colspan = max(colspan, 1) if colspan is not None else 1
+        rowspan = parse_html_integer(self.element.get('rowspan', ''))
+        self.rowspan = max(rowspan, 0) if rowspan is not None else 1
```
liZe: Shouldn’t that be `max(rowspan, 1)` instead?

weasyprint/css/__init__.py
```diff
@@ -488,10 +490,13 @@ def check_style_attribute(element, style_attribute):
...
             if element.get('size'):
-                try:
-                    size = int(element.get('size'))
-                except ValueError:
-                    LOGGER.warning('Invalid value for size: %s', size)
+                parsed = parse_html_integer(element.get('size'))
+                if parsed is not None:
+                    size = parsed
+                else:
+                    LOGGER.warning(
```
liZe: One line is enough. 😊 (Same for line 378.)

weasyprint/css/__init__.py
```diff
@@ -367,15 +368,16 @@ def check_style_attribute(element, style_attribute):
                 yield specificity, check_style_attribute(
                     element, f'font-family:{element.get("face")}')
             if element.get('size'):
-                size = element.get('size').strip()
-                relative_plus = size.startswith('+')
-                relative_minus = size.startswith('-')
+                size_attr = element.get('size').lstrip(' \t\n\f\r')
```
liZe: You can use `HTML_WHITESPACE`.

weasyprint/html_utils.py
```diff
@@ -0,0 +1,43 @@
...
+
+# https://html.spec.whatwg.org/multipage/#space-character
+HTML_WHITESPACE = ' \t\n\f\r'
+ASCII_DIGITS = frozenset('0123456789')
+
+
+def parse_html_integer(string):
```
liZe: That would be a good exercise for regular expression lovers, but we’ll save that for later! 😁

liZe (changes requested):
There’s a lot of small comments, but the overall feature is cool. Thanks a lot!

liZe:
Hi!

Thanks for this pull request.

> we'll want to start contributing when we can, so this is hopefully the first of many PRs, but I'll wait for a review and any comments before starting on anything new.

That’s good news!

weasyprint/formatting_structure/boxes.py
liZe: I was wrong, 0 is allowed for rowspan. I’ll fix that!

weasyprint/html_utils.py
liZe: I’ll add a commit for that. 😁

liZe:
Thank you!

PR #2722, opened 2026-04-04
Replace .isdigit() with ASCII-only digit check

### LINKED ISSUE #2721: Replace .isdigit() with ASCII-only digit check in presentational hints
Python's `str.isdigit()` returns `True` for non-ASCII digit characters (e.g. Arabic-Indic digits `٠١٢٣`), but the HTML spec defines digits as [ASCII digits](https://infra.spec.whatwg.org/#ascii-digit) only (`0-9`).

Several places in `find_style_attributes` use `.isdigit()` to decide whether to append `px` to attribute values:

- `cellpadding.isdigit()` (line ~402 in `css/__init__.py`)
- `element.get('width').isdigit()` for `<hr>` (line ~507)
- Similar checks for `hspace`, `vspace`, `width`, `height` on various elements

These should use an ASCII-only digit check to match browser behavior.

Related to #2636, which fixed the same class of issue for integer parsing.

danfitz36:
## Summary

- Add `is_ascii_digits()` utility that only matches ASCII digits (`0-9`), unlike Python's `str.isdigit()` which also matches non-ASCII digit characters (e.g. Arabic-Indic `٠١٢٣`)
- Replace all `.isdigit()` calls in presentational hint parsing (`css/__init__.py`) and PDF form field handling (`pdf/anchors.py`) with the new function
- Affected attributes: `cellpadding`, `hspace`, `vspace`, `width`, `height`, `maxlength`

## Context

Python's `str.isdigit()` returns `True` for non-ASCII digit characters, but the [HTML spec](https://infra.spec.whatwg.org/#ascii-digit) defines digits as ASCII digits only (`0-9`). This could cause incorrect unit appending (e.g. treating Arabic-Indic digits as a bare number and appending `px`).

Note: the two `.isdigit()` calls in `svg/bounding_box.py` are intentionally left unchanged — they check individual characters in SVG path data, a different context.

Related to #2636.

## Test plan

- [x] Unit tests for `is_ascii_digits`: basic values, empty string, non-digits, non-ASCII digits
- [x] All existing tests pass (3985 passed, 0 regressions)
- [x] `ruff check` passes

Fixes #2721

### DIFF
--- a/tests/test_html_utils.py
+++ b/tests/test_html_utils.py
@@ -1,0 +1,30 @@
+"""Tests for HTML microsyntax parsing utilities."""
+
+from weasyprint.html_utils import is_ascii_digits
+
+
+def test_is_ascii_digits_basic():
+    assert is_ascii_digits('123')
+    assert is_ascii_digits('0')
+    assert is_ascii_digits('007')
+
+
+def test_is_ascii_digits_rejects_empty():
+    assert not is_ascii_digits('')
+
+
+def test_is_ascii_digits_rejects_non_digits():
+    assert not is_ascii_digits('abc')
+    assert not is_ascii_digits('12px')
+    assert not is_ascii_digits('100%')
+    assert not is_ascii_digits(' 5')
+    assert not is_ascii_digits('+3')
+    assert not is_ascii_digits('-1')
+
+
+def test_is_ascii_digits_rejects_non_ascii_digits():
+    # Python's str.isdigit() returns True for these, but HTML spec
+    # requires ASCII digits only.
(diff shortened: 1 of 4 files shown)


liZe:
Thanks for the pull request!

We’ll merge #2720 first and see what we can do for this. I think that it may be better to include these changes in a larger fix that can handle min/max values from the HTML specification and checks input values more seriously. 

There’s no need to work more on this PR now, we’ll see after #2720.

liZe:
It’s now included in #2773, with better parsers for HTML attributes.

Issue #2718, opened 2026-04-02
Providing Windows executable also as 'onedir' compilation


isenmann:
Hi,

at the moment the Windows executable is built as a standalone all-in-one executable, which is fine and is working for most use cases. The downside of this approach is that every time you execute the exe it will first unpack about 50 MB before doing it's job. That's fine if you just generate PDFs occasionally, but if you generate PDFs often and even in parallel, all the processes unpack those 50 MB all the time which leads to heavy disk I/O and slow downs on that machine. So changing this to the onedir compilation it would also be faster because the unpack step will be not needed.

The only thing which must be changed is the exe workflow file which must be changed to this:

```
- name: Generate executable
   run: python -m PyInstaller weasyprint/__main__.py -n weasyprint --onedir
 - name: Test executable
   run: dist/weasyprint/weasyprint --info
 - name: Store executable
   uses: actions/upload-artifact@v4
   with:
      name: weasyprint-windows
      path: |
        dist/weasyprint
        README.rst
        LICENSE
``` 

Would it be possible to also provide the Windows exe as a onedir solution **additionally** to the all-in-one exe? 

Cheers
Daniel

liZe:
Hi!

From [PyInstaller’s documentation]()https://pyinstaller.org/en/stable/usage.html#cmdoption-D:
> `-D, --onedir`
>   Create a one-folder bundle containing an executable (default)

This option seems to be set by default. Did I miss something?

isenmann:
Yeah, because at the moment in line 23 in the exe.yml it is set to -F which means, that it should generate a single exe file.

liZe:
> Yeah, because at the moment in line 23 in the exe.yml it is set to -F which means, that it should generate a single exe file.

Of course. 😊

Interested in opening a PR that updates `exe.yml` to generate both?

isenmann:
Sure, I can do that. I will provide one and link it to this issue in the next couple of days.

isenmann:
Like I said before, I have open the PR for this and linked it.

PR #2727, opened 2026-04-15
Generate the WIndows package also as a one dir solution

isenmann:
Like described in #2718, this will be the PR to have the Windows exe as a single file self extracted like before and additionally as a single directory executable to avoid extracting and deleting the self-contained exe all the time it is used. 

I don't know where the release zip is generated, couldn't find it in the workflow files. But the this workflow file generates both and upload it to the artifacts of the build.


liZe:
Thanks for the pull request! It’s been rebased, simplified, and pushed to `main`.

PR #2728, opened 2026-04-16
Fix: honor break-before and break-after on floated elements

### LINKED ISSUE #2277: Page break attribute not always honored
Hi,

I noticed that weasyprint does not honor the page break attributes when generating a pdf. Looks like a bug to me as it happens with a specific mix of div / css . See example below :

`index.html ` 
```html
<html>
<body class="dark">
<main class="page-content" aria-label="Content">

<div class="container experience-container">
  <h3>Title h3</h3>
  <div class="row clearfix layout layout-left" style="" >
    <div class="col-xs-12 col-sm-4 col-md-3 col-print-12 details" > <h4>Title h4</h4> </div>
    <div class="col-xs-12 col-sm-8 col-md-9 col-print-12"> IN DIV page break at end <p style="page-break-after: always;"></p>  </div>
  </div>


  
  <div class="row clearfix layout layout-left" style="" >
    <div class="col-xs-12 col-sm-4 col-md-3 col-print-12 details" ><h4>Title h4 II</h4> </div>
    <div class="col-xs-12 col-sm-8 col-md-9 col-print-12"> SHOULD BE ON NEXT PAGE <br /> </div>
  </div>

</div>

</main>
</body>
</html>

```

[merged-css.txt](https://github.com/user-attachments/files/17447940/merged-css.txt)

```bash
mv merged-css.txt merged.css
weasyprint -s merged.css -q index.html index.pdf && evince index.pdf

```

You should see no page break, and everything in the same pdf page.

I'm not a css guy, the file comes from a merge of bootstrap and other css files. I did not write them.

I noticed 2 things : 
* If I remove both `col-xs-12` and `col-sm-8` on the line 5 (where the page break is). It works 
* If I move the page break out of the `row clearfix layout layout-left` div scope. It works.

It is weird to me because all those classes don't mention page break avoid (even if there are page break avoid in the css). 

Let me know if you see some obvious workaround for me. The html is generated but I can change the text inside the div (" IN DIV page break at end"). As of now I use the print to pdf from firefox that works, but not very fun to automate.

Thanks

danfitz36:
## Summary
- Floated elements bypassed break property handling because they are routed through `_out_of_flow_layout` instead of `_in_flow_layout`
- Added forced page break detection for floats following the CSS Fragmentation spec: *"User agents should also apply these properties to floated boxes whose containing block is in the normal flow of the root fragmented element."*
- Handles both `break-before` and `break-after` on floated elements

## Test plan
- [x] Added `test_float_break_after_page` — verifies `break-after: page` on a float forces a page break
- [x] Added `test_float_break_before_page` — verifies `break-before: page` on a float forces a page break

Fixes #2277

### DIFF
--- a/tests/layout/test_float.py
+++ b/tests/layout/test_float.py
@@ -840,3 +840,46 @@ def test_first_letter_float():
     assert first_letter.position_x == 0
     # TODO: fix problem described in #1859.
     # assert text.position_x == 20
+
+
+@assert_no_logs
+def test_float_break_after_page():
+    # https://github.com/Kozea/WeasyPrint/issues/2277
+    # break-after: page on a floated element should force a page break.
+    pages = render_pages('''
+      <style>
+        @page { size: 100px; margin: 10px }
+      </style>
+      <div style="break-after: page; float: left">float</div>
+      <div>end</div>
+    ''')
+    assert len(pages) == 2
+    html1, = pages[0].children
+    body1, = html1.children
+    float_div, = body1.children
+    assert float_div.position_y == 10
+    html2, = pages[1].children
+    body2, = html2.children
+    end_div, = body2.children
+    assert end_div.position_y == 10
+
+
+@assert_no_logs
+def test_float_break_before_page():
+    # break-before: page on a floated element should force a page break.
+    pages = render_pages('''
+      <style>
+        @page { size: 100px; margin: 10px }
+      </style>
+      <div>start</div>
+      <div style="break-before: page; float: left">float</div>
+    ''')
+    assert len(pages) == 2
+    html1, = pages[0].children
+    body1, = html1.children
+    start_div, = body1.children
+    assert start_div.position_y == 10
+    html2, = pages[1].children
+    body2, = html2.children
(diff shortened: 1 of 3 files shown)


liZe:
Hi!

Thanks for this pull request, here’s a short review.

- As we want to check breaks between blocks, there’s no need to check before and after each block: checking once is enough, as we do for in-flow blocks.
- Forced or prevented breaks are determined using the `break-after` of the first block and the `break-before` of the second block, with a possible combination. We never rely on only one value, as you did with `break-after`.
- Your second check is done before breaking or aborting the float rendering. It’s wrong if the float is split between two pages, as we don’t want to honor `break-after` value if the end of the float is not rendered.
- Your first check is the same as the one done for in-flow blocks. Instead of duplicating the code, it’s possible to move it at the beginning of the child layout instead.
- As we want to break between floats, the `find_last_in_flow_child` function is not adapted to find the previous block that creates a class A break. We have to find the last child that is either an in-flow block or a float.

I updated your code to fix these issues, and added more tests.

PR #2733, opened 2026-04-20
Document paper size and orientation command-line usage

### LINKED ISSUE #2703: Document command-line option for papersize and orientation
It would be nice if weasyprint supported command-line flags to specify the orientation and paper size.

I'm aware that this can be specified within the document, using e.g.

```
<style>
@page {
  size: A3 landscape;
}
</style>
```

Or that we can apply a stylesheet with -s.

So, why would this be this useful?

1. simplicity.
2. weasyprint is a great replacement to, and improvement on, the (now unsupported) wkhtmltopdf, and it would be nice if it could work compatibly, for ease of drop-in replacement.  E.g.
```
cat input.html | wkhtmltopdf -O Landscape -s A3 - output.pdf
```

Thanks for your time, and an amazing product.

ElangoSubramani:
Description:
Fixes #2703

### DIFF
--- a/docs/common_use_cases.rst
+++ b/docs/common_use_cases.rst
@@ -87,14 +87,22 @@ Adjust Document Dimensions
 
 WeasyPrint does not provide support for adjusting page size or document margins
 via command-line flags. This is best accomplished with the CSS ``@page``
-at-rule. Consider the following example:
+at-rule, which can be added to the document or provided in a stylesheet.
+Consider the following example:
 
 .. code-block:: css
 
   @page {
-    size: Letter; /* Change from the default size of A4 */
+    size: A3 landscape; /* Change the page size and orientation */
     margin: 3cm; /* Set margin on each page */
   }
+
+To apply these rules from the command line without creating a separate CSS
+file, use the ``-s`` option with shell process substitution:
+
+.. code-block:: sh
+
+  $ weasyprint input.html output.pdf -s <(echo "@page { size: A3 landscape }")
 
 There is much more which can be achieved with the ``@page`` at-rule,
 such as page numbers, headers, etc. Read more about the page_ at-rule.
--- a/docs/common_use_cases.rst
+++ b/docs/common_use_cases.rst
@@ -93,7 +93,7 @@ Consider the following example:
 .. code-block:: css
 
   @page {
-    size: A3 landscape; /* Change the page size and orientation */
+    size: A3 landscape; /* Change from the default size and orientation of A4 portrait */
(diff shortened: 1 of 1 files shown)


ElangoSubramani:
Can anyone please review?

docs/common_use_cases.rst
```diff
@@ -87,15 +87,23 @@ Adjust Document Dimensions
...
+    size: A3 landscape; /* Change the page size and orientation */
     margin: 3cm; /* Set margin on each page */
   }
 
+To apply these rules from the command line without creating a separate CSS
+file, use the ``-s`` option with shell process substitution:
+
+.. code-block:: sh
+
+  $ weasyprint input.html output.pdf -s <(echo "@page { size: A3 landscape }")
```
grewn0uille: That’d be interesting to also have the margin here, so the rules with and without the stylesheet would be the same.

docs/common_use_cases.rst
```diff
@@ -87,15 +87,23 @@ Adjust Document Dimensions
...
 via command-line flags. This is best accomplished with the CSS ``@page``
-at-rule. Consider the following example:
+at-rule, which can be added to the document or provided in a stylesheet.
+Consider the following example:
 
 .. code-block:: css
 
   @page {
-    size: Letter; /* Change from the default size of A4 */
+    size: A3 landscape; /* Change the page size and orientation */
```
grewn0uille: I think the information about the default size/orientation is also important. Maybe something like "Change from the default size and orientation of A4 portrait"? What do you think?

grewn0uille (commented):
Hi,

Thanks for the PR! Here’s few comments.

docs/common_use_cases.rst
ElangoSubramani: Yes, I feel the same
I'll just add the orientation. Is that fine?

docs/common_use_cases.rst
ElangoSubramani: Okay, will add margin there
Thanks for reviewing

docs/common_use_cases.rst
grewn0uille: I think so

docs/common_use_cases.rst
ElangoSubramani: I have updated

docs/common_use_cases.rst
ElangoSubramani: I have updated

grewn0uille:
Looks good to me. I’ll merge that! Thank you :)

PR #2737, opened 2026-04-21
Support SVG transform angle units

### LINKED ISSUE #2735: SVG transform angle units such as deg are not parsed
## Summary

Reviewed base: `main` at commit `71115d931cd06c0a6906e23575619e9a4b237df7`.

SVG transform functions parse angle arguments as lengths, so CSS/SVG angle units such as `deg`, `rad`, `grad`, and `turn` are not handled. This breaks diagrams generated by tools such as Mermaid, which emit transforms like `rotate(45deg)` and `rotate(90deg)` in SVG styles.

## Problem

Unitless `rotate(90)` works, but `rotate(90deg)` does not render equivalently. The transform parser currently routes transform arguments through the length parser, which is correct for `translate(...)` but not for angle-bearing functions such as `rotate(...)` and `skew(...)`.

A reduced example:

```html
<style>
  @page { size: 40px 40px }
  svg { display: block }
</style>
<svg width="40px" height="40px" xmlns="http://www.w3.org/2000/svg">
  <style>
    .milestone { transform: rotate(45deg) scale(.7, .7); }
    .tick text { transform: translate(3px, 1px) rotate(90deg); }
  </style>
  <rect class="milestone" x="14" y="4" width="10" height="10"
        transform-origin="19px 9px" stroke="green" fill="green" />
  <g class="tick" transform="translate(12, 18)">
    <text y="3" font-family="weasyprint" font-size="3" fill="blue">
      2026-04-01
    </text>
  </g>
</svg>
```

## Why It Matters

This affects common generated SVGs. In Mermaid Gantt charts, milestones are represented with `rotate(45deg)`, and custom tick labels often use `rotate(90deg)`. In PDF output the result can be square milestones or unrotated/missing axis labels.

## Code References

- `transform()` parses all transform arguments as lengths before dispatching by transform type: https://github.com/Kozea/WeasyPrint/blob/71115d931cd06c0a6906e23575619e9a4b237df7/weasyprint/svg/utils.py#L164-L207
- Existing tests cover unitless rotation but not angle units: https://github.com/Kozea/WeasyPrint/blob/71115d931cd06c0a6906e23575619e9a4b237df7/tests/draw/svg/test_transform.py#L106-L147

## Proposed Fix Envelope

Parse angle arguments with an angle parser for `rotate`, `skew`, `skewX`, and `skewY`, while keeping length parsing for `matrix`, `translate`, and transform origins. Add regression coverage comparing unitless rotation with `deg` rotation and a Mermaid-like CSS transform sample.

## Suggested Test Scenarios

- Regression: `transform="rotate(90deg)"` renders the same as `transform="rotate(90)"`.
- Regression: Mermaid-like SVG CSS using `.milestone { transform: rotate(45deg) scale(...) }` renders the same as equivalent transform attributes.
- Control: existing unitless rotate, rotate with center, translate, scale, and skew transform tests still pass.

moreaki:
## What changed

Fixes #2735.

This adds SVG angle parsing for transform functions that take angles, so values such as `rotate(90deg)`, `rotate(.25turn)`, `rotate(1.5708rad)`, and `skew(20deg)` are treated as angles instead of lengths.

## Why

Generated SVGs commonly use explicit angle units in CSS transforms. Mermaid Gantt diagrams, for example, emit rules such as `transform: rotate(45deg) scale(...)` for milestones and `rotate(90deg)` for rotated tick labels. WeasyPrint previously parsed transform arguments as lengths before dispatching by transform function, which made angle units fail.

## Tests

- `python -m pytest tests/draw/svg/test_transform.py -q`
- `python -m pytest tests/draw/svg -q`

### DIFF
--- a/tests/draw/svg/test_transform.py
+++ b/tests/draw/svg/test_transform.py
@@ -1,5 +1,7 @@
 """Test how SVG shapes are transformed."""
 
+import pytest
+
 from ...testing_utils import assert_no_logs
 
 
@@ -147,8 +149,10 @@ def test_transform_rotate_cx_cy(assert_pixels):
     ''')
 
 
+@pytest.mark.parametrize(
+    'angle', ['90deg', '100grad', '1.5707963267948966rad', '.25turn'])
 @assert_no_logs
-def test_transform_rotate_degrees(assert_same_renderings):
+def test_transform_rotate_angle_units(assert_same_renderings, angle):
     assert_same_renderings('''
       <style>
         @page { size: 9px }
@@ -158,57 +162,14 @@ def test_transform_rotate_degrees(assert_same_renderings):
         <rect x="2" y="-7" width="4" height="5" transform="rotate(90)"
               stroke-width="2" stroke="red" fill="none" />
       </svg>
-    ''', '''
+    ''', f'''
       <style>
-        @page { size: 9px }
-        svg { display: block }
+        @page {{ size: 9px }}
+        svg {{ display: block }}
       </style>
       <svg width="9px" height="9px" xmlns="http://www.w3.org/2000/svg">
-        <rect x="2" y="-7" width="4" height="5" transform="rotate(90deg)"
+        <rect x="2" y="-7" width="4" height="5" transform="rotate({angle})"
               stroke-width="2" stroke="red" fill="none" />
-      </svg>
-    ''')
-
-
-@assert_no_logs
-def test_transform_mermaid_gantt_css(assert_same_renderings):
-    assert_same_renderings('''
-      <style>
-        @page { size: 40px 40px }
-        svg { display: block }
-      </style>
-      <svg width="40px" height="40px" xmlns="http://www.w3.org/2000/svg">
-        <rect x="14" y="4" width="10" height="10" transform-origin="19px 9px"
-              transform="rotate(45) scale(.7 .7)"
-              stroke="green" fill="green" />
-        <g transform="translate(12, 18)">
-          <text y="3" transform="translate(3 1) rotate(90)"
-                font-family="weasyprint" font-size="3" fill="blue">
-            2026-04-01
-          </text>
-        </g>
-      </svg>
-    ''', '''
-      <style>
-        @page { size: 40px 40px }
-        svg { display: block }
-      </style>
-      <svg width="40px" height="40px" xmlns="http://www.w3.org/2000/svg">
-        <style>
-          .milestone {
-            transform: rotate(45deg) scale(.7, .7);
-          }
-          .tick text {
-            transform: translate(3px, 1px) rotate(90deg);
-          }
-        </style>
-        <rect class="milestone" x="14" y="4" width="10" height="10"
-              transform-origin="19px 9px" stroke="green" fill="green" />
-        <g class="tick" transform="translate(12, 18)">
-          <text y="3" font-family="weasyprint" font-size="3" fill="blue">
-            2026-04-01
-          </text>
-        </g>
       </svg>
     ''')
 
--- a/tests/draw/svg/test_transform.py
+++ b/tests/draw/svg/test_transform.py
@@ -148,6 +148,72 @@ def test_transform_rotate_cx_cy(assert_pixels):
 
 
 @assert_no_logs
+def test_transform_rotate_degrees(assert_same_renderings):
+    assert_same_renderings('''
+      <style>
+        @page { size: 9px }
+        svg { display: block }
+      </style>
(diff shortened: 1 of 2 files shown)


weasyprint/svg/utils.py
```diff
@@ -57,6 +57,20 @@ def size(string, font_size=None, percentage_reference=None):
     return 0
 
 
+def angle(string):
```
liZe, line 61: Could it be possible to use `css.units.ANGLE_TO_RADIANS` here?

weasyprint/svg/utils.py
```diff
@@ -57,6 +57,20 @@ def size(string, font_size=None, percentage_reference=None):
...
+    """Compute an angle in degrees from an SVG transform value."""
```
liZe: We should return radians instead.

tests/draw/svg/test_transform.py
```diff
@@ -147,6 +147,72 @@ def test_transform_rotate_cx_cy(assert_pixels):
...
+      </style>
+      <svg width="9px" height="9px" xmlns="http://www.w3.org/2000/svg">
+        <rect x="2" y="-7" width="4" height="5" transform="rotate(90deg)"
+              stroke-width="2" stroke="red" fill="none" />
+      </svg>
+    ''')
+
+
+@assert_no_logs
+def test_transform_mermaid_gantt_css(assert_same_renderings):
```
liZe: It’s better to use names related to tested features, if possible. And maybe the test above is enough, as this one doesn’t add anything I think.

tests/draw/svg/test_transform.py
```diff
@@ -147,6 +147,72 @@ def test_transform_rotate_cx_cy(assert_pixels):
...
+        <rect x="2" y="-7" width="4" height="5" transform="rotate(90)"
+              stroke-width="2" stroke="red" fill="none" />
+      </svg>
+    ''', '''
+      <style>
+        @page { size: 9px }
+        svg { display: block }
+      </style>
+      <svg width="9px" height="9px" xmlns="http://www.w3.org/2000/svg">
+        <rect x="2" y="-7" width="4" height="5" transform="rotate(90deg)"
```
liZe: Could it be possible to test all units here, using `@pytest.mark.parametrize`?

tests/draw/svg/test_transform.py
```diff
@@ -147,6 +147,72 @@ def test_transform_rotate_cx_cy(assert_pixels):
...
+@assert_no_logs
+def test_transform_mermaid_gantt_css(assert_same_renderings):
+    assert_same_renderings('''
+      <style>
+        @page { size: 40px 40px }
+        svg { display: block }
+      </style>
+      <svg width="40px" height="40px" xmlns="http://www.w3.org/2000/svg">
+        <rect x="14" y="4" width="10" height="10" transform-origin="19px 9px"
+              transform="rotate(45) scale(.7 .7)"
```
liZe: Don’t we want to have a unit here?

liZe (changes requested):
Thanks a lot for this pull request. 🙏

Here’s a small list of comments, that’s only minor changes.

moreaki:
Thanks again for the careful review! I’ve addressed the angle-unit feedback in `c2f8813f`: angle parsing now uses `ANGLE_TO_RADIANS` and returns radians, rotate/skew consume radians directly, and the SVG transform test is now parametrized across `deg`, `grad`, `rad`, and `turn`.

weasyprint/svg/utils.py
moreaki: Done, thanks. `angle()` now uses `ANGLE_TO_RADIANS` for angle-unit conversion.

weasyprint/svg/utils.py
moreaki: Done, thanks. `angle()` now returns radians directly, and rotate/skew consume that value without `math.radians()`.

tests/draw/svg/test_transform.py
moreaki: Agreed, thanks. I removed the Mermaid-specific test and renamed the remaining coverage to `test_transform_rotate_angle_units`.

tests/draw/svg/test_transform.py
moreaki: Done, thanks. The angle-unit test is now parametrized over `deg`, `grad`, `rad`, and `turn`.

tests/draw/svg/test_transform.py
moreaki: Done, thanks. The comparison case now uses explicit units for every parametrized value.

liZe (approved):
Thanks.

PR #2738, opened 2026-04-21
Account for transformed SVG group bounds

### LINKED ISSUE #2736: SVG opacity group bounding boxes ignore child transforms
## Summary

Reviewed base: `main` at commit `71115d931cd06c0a6906e23575619e9a4b237df7`.

SVG group bounding boxes ignore transforms applied to child elements. When a group has opacity, WeasyPrint paints it as a Form XObject using this under-sized BBox, so transformed children can be clipped.

## Problem

For an opacity group, WeasyPrint calculates a group bounding box from child bounding boxes. Child transforms are applied while drawing, but they are not applied while extending the parent group BBox. The Form XObject therefore clips content that has moved or rotated outside the child’s original untransformed bounds.

A reduced example:

```html
<style>
  @page { size: 12px }
  svg { display: block }
</style>
<svg width="12px" height="12px" xmlns="http://www.w3.org/2000/svg">
  <g opacity="0.5">
    <rect x="0" y="0" width="4" height="4" fill="red"
          transform="translate(5 5)" />
  </g>
</svg>
```

This should render like a half-opacity rectangle at `(5, 5)`, but the opacity group’s BBox can clip it to the original untransformed location.

## Why It Matters

Generated SVGs often put opacity on groups. Mermaid Gantt charts put opacity on tick groups; when tick labels are rotated, their Form XObject BBox can be too small, so date labels are clipped in PDF output even though the same SVG renders correctly in browsers.

## Code References

- Group BBoxes extend from child BBoxes without applying child transforms: https://github.com/Kozea/WeasyPrint/blob/71115d931cd06c0a6906e23575619e9a4b237df7/weasyprint/svg/bounding_box.py#L212-L223
- Opacity groups use calculated BBoxes for Form XObjects: https://github.com/Kozea/WeasyPrint/blob/71115d931cd06c0a6906e23575619e9a4b237df7/weasyprint/svg/__init__.py#L560-L575

## Proposed Fix Envelope

When calculating a group BBox, apply each child’s SVG transform matrix to the child BBox corners before extending the parent group BBox. Add regression coverage for transformed content inside an opacity group.

## Suggested Test Scenarios

- Regression: a translated child inside `g opacity="0.5"` is not clipped to its untransformed bounds.
- Control: existing opacity and transform tests still pass.
- Real-world validation: Mermaid Gantt tick labels inside opacity tick groups remain fully visible after rotation.

moreaki:
## What changed

Fixes #2736.

This applies child transform matrices when extending an SVG group bounding box from child bounding boxes.

## Why

Opacity groups are rendered as Form XObjects whose BBox is computed from the SVG group bounds. If transformed children are measured in their original, untransformed positions, the Form XObject BBox can be too small and clip visible content. This shows up in generated SVGs such as Mermaid Gantt charts, where transformed tick labels sit inside opacity groups.

## Tests

- `python -m pytest tests/draw/svg/test_transform.py -q`
- `python -m pytest tests/draw/svg -q`

### DIFF
--- a/tests/draw/svg/test_opacity.py
+++ b/tests/draw/svg/test_opacity.py
@@ -146,6 +146,22 @@ def test_translate_opacity(assert_same_renderings):
 
 
 @assert_no_logs
+def test_group_opacity_transformed_child(assert_same_renderings):
+    assert_same_renderings(
+        opacity_source % '''
+            <g opacity="0.5">
+              <rect x="0" y="0" width="4" height="4" fill="red"
+                    transform="translate(5 5)" />
+            </g>
+        ''',
+        opacity_source % '''
+            <rect x="0" y="0" width="4" height="4" fill="rgb(255, 127, 127)"
+                  transform="translate(5 5)" />
+        ''',
+    )
+
+
+@assert_no_logs
 def test_translate_use_opacity(assert_same_renderings):
     # Regression test for #1976.
     assert_same_renderings(
--- a/tests/draw/svg/test_opacity.py
+++ b/tests/draw/svg/test_opacity.py
@@ -155,6 +155,19 @@ def test_group_opacity_transformed_child(assert_same_renderings):
             </g>
         ''',
         opacity_source % '''
+            <g opacity="0.5"><g transform="translate(3 3)">
+              <rect x="0" y="0" width="4" height="4" fill="red"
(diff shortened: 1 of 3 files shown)


liZe:
Thanks for the pull request.

I added the support for `use` tags and deduplicated the transformation logic.

Issue #2580, opened 2025-10-14
Whole table (instead of overflowed rows) is carried to the second page


nhattantran:
I use `<tfoot>` tag to define a table footer row so that all pages have that row in the end of the table

<img width="933" height="245" alt="Image" src="https://github.com/user-attachments/assets/bb219a0e-4a7e-4ebf-8b94-cfab4dffd1ba" />

However when I add one more row, the whole table is carried to the next page (the cyan part indicates the bottom limit using page margin)

When I add 2 more rows, it works normally (only the last row is carried to the next page) and the footer row is shown at both pages

I am using version 66.0

liZe:
Hi,

Could you please provide a sample, so that we can reproduce your problem?

nhattantran:
> Hi,
> 
> Could you please provide a sample, so that we can reproduce your problem?

Hi @liZe what do you mean by "sample"? html, css code?
You can see that from the above image, the whole table is on page 1. However, when I add 1 more row, the whole table is carried to the second page. I expect that only the last row is carried to the next page

grewn0uille:
Hi,

By "sample" we indeed mean the HTML/CSS leading to the issue.
So we can reproduce the problem and then find the cause!

nhattantran:
Here is my html code
```
<table>
          <thead class="table-header">
          <tr>
            <th style="min-width: 45px" class="text-left table-header-border-top">Date</th>
            <th style="min-width: 50px" class="text-left table-header-border-top">Details</th>
            <th style="min-width: 30px" class="text-right table-header-border-top">a</th>
            <th style="min-width: 30px" class="text-right table-header-border-top">b</th>
            <th style="min-width: 30px" class="text-right table-header-border-top">c</th>
            <th style="min-width: 50px" class="text-right table-header-border-top">d</th>
          </tr>
          </thead>
          <tbody>
          {% if template_data.data %} {% for row in
          template_data.data %}

          <tr>
            <td class="text-left">{{row.date}}</td>
            <td class="text-left">{{row.narrative}} </td>
            <td class="text-right">{% if row.a %}{{row.a}}{% else %}-{% endif %}</td>
            <td class="text-right">{% if row.b %}{{row.b}}{% else %}-{% endif %}</td>
            <td class="text-right">{% if row.c %}{{row.c}}{% else %}-{% endif %}</td>
            <td class="text-right">{{row.d}}</td>
          </tr>
          </tbody>
          {% endfor %}
          {% else %}
[...]

PR #2739, opened 2026-04-21
Fix table break retry after padding overflow

moreaki:
## Summary

Fixes a block layout retry case that can push a table to the next page when the table is wrapped in a block with bottom padding. Adds a regression test based on the shape reported in #2580: previous content, a padded wrapper, and a table with `thead`, `tbody`, and `tfoot`.

## Root Cause

When a child's border, padding, or margin overflowed the current page, `_in_flow_layout` removed the laid-out child and retried layout with additional `bottom_space`. The first layout attempt can mutate the original child's coordinates, for example through table margin collapsing. The retry then started from the already-shifted position, effectively double-counting vertical offset and incorrectly deciding that the table no longer fit on the page.

The retry call also passed `discard` and `max_lines` into the `first_letter_style` / `first_line_style` argument slots, which meant the second layout attempt was not preserving the intended block layout parameters.

## Changes

- Save the child's original `position_x` and `position_y` before the first layout attempt.
- Restore those coordinates before retrying layout after bottom spacing overflow.
- Pass `first_letter_style`, `first_line_style`, `discard`, and `max_lines` to the retry in the correct order.
- Add a regression test proving that a table in a padded wrapper remains on the current page when its contents fit.

## Validation

- `pytest tests/layout/test_table.py -q`
- `pytest tests/layout/test_block.py -q`
- `ruff check weasyprint/layout/block.py tests/layout/test_table.py`

### DIFF
--- a/tests/layout/test_table.py
+++ b/tests/layout/test_table.py
@@ -3175,6 +3175,46 @@ def test_table_break_children_margin():
     assert len(render_pages(html)) == 3
 
 
+def test_table_break_in_padded_block():
+    # Regression test for #2580.
+    page1, = render_pages('''
+      <style>
+        @page { size: 100px; margin: 0 }
+        body { margin: 0; font: 2px/1 weasyprint }
+        div.previous { height: 20px }
+        div.wrapper { padding-bottom: 6px }
+        table {
+          border-collapse: collapse;
+          border-spacing: 0;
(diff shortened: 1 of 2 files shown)


moreaki:
Superseded by #2740 with a neutral branch name and PR title.

Issue #2717, opened 2026-03-31
TypeError: unsupported operand type(s) for |=: 'bool' and 'dict'


elisa-aleman:
I got this error about `first_letter_style`, which could stem from this CSS:

```
h1:first-letter,h2:first-letter,h3:first-letter,h4:first-letter,h5:first-letter,h6:first-letter{
    font-size:120%
}
```

```
Traceback (most recent call last):
File ".../test.py", line 21, in <module>
weasyprint.HTML(string=html).write_pdf(filename.with_suffix('.pdf'))
File ".../weasyprint/__init__.py", line 266, in write_pdf
self.render(font_config, counter_style, color_profiles, **options)
File ".../weasyprint/__init__.py", line 222, in render
return Document._render(
^^^^^^^^^^^^^^^^^
File ".../weasyprint/document.py", line 201, in _render
[Page(page_box) for page_box in page_boxes],
^^^^^^^^^^
File ".../weasyprint/layout/__init__.py", line 131, in layout_document
pages = list(make_all_pages(context, root_box, html, pages))
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File ".../weasyprint/layout/page.py", line 1005, in make_all_pages
page, resume_at = remake_page(i, context, root_box, html)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File ".../weasyprint/layout/page.py", line 946, in remake_page
page, resume_at, next_page = make_page(
^^^^^^^^^^
File ".../weasyprint/layout/page.py", line 664, in make_page
root_box, resume_at, next_page, _, _, _ = block_level_layout(
^^^^^^^^^^^^^^^^^^^
File ".../weasyprint/layout/block.py", line 59, in block_level_layout
return block_level_layout_switch(
^^^^^^^^^^^^^^^^^^^^^^^^^^
File ".../weasyprint/layout/block.py", line 75, in block_level_layout_switch
return block_box_layout(
^^^^^^^^^^^^^^^^^
File ".../weasyprint/layout/block.py", line 126, in block_box_layout
result = block_container_layout(
^^^^^^^^^^^^^^^^^^^^^^^
File ".../weasyprint/layout/block.py", line 763, in block_container_layout
next_page, new_children, new_max_lines) = _in_flow_layout(
^^^^^^^^^^^^^^^^
[...]

PR #2740, opened 2026-04-21
Fix table break retry after padding overflow

moreaki:
## Summary

Fixes a block layout retry case that can push a table to the next page when the table is wrapped in a block with bottom padding. Adds a regression test based on the shape reported in #2580: previous content, a padded wrapper, and a table with `thead`, `tbody`, and `tfoot`.

The same bad retry call also caused #2717: `discard` and `max_lines` were passed into the `first_letter_style` / `first_line_style` argument slots, so a retry involving multicolumn layout and `:first-letter` could crash with `TypeError: unsupported operand type(s) for |=: bool and dict`. This PR adds the maintainer-provided sample as a regression test too.

Related issues: #2580, #2717

## Root Cause

When a child's border, padding, or margin overflowed the current page, `_in_flow_layout` removed the laid-out child and retried layout with additional `bottom_space`. The first layout attempt can mutate the original child's coordinates, for example through table margin collapsing. The retry then started from the already-shifted position, effectively double-counting vertical offset and incorrectly deciding that the table no longer fit on the page.

The retry call also passed `discard` and `max_lines` into the `first_letter_style` / `first_line_style` argument slots, which meant the second layout attempt was not preserving the intended block layout parameters and could crash when merging `:first-letter` styles.

## Changes

- Save the child's original position before the first layout attempt.
- Restore that position before retrying layout after bottom spacing overflow.
- Pass `first_letter_style`, `first_line_style`, `discard`, and `max_lines` to the retry in the correct order.
- Add a regression test proving that a table in a padded wrapper remains on the current page when its contents fit.
- Add a regression test for multicolumn layout with `:first-letter`.

## Validation

- `pytest tests/layout/test_column.py -q`
- `pytest tests/layout/test_table.py tests/layout/test_block.py -q`
- `ruff check weasyprint/layout/block.py tests/layout/test_table.py tests/layout/test_column.py`

### DIFF
--- a/tests/layout/test_column.py
+++ b/tests/layout/test_column.py
@@ -1113,3 +1113,25 @@ def test_columns_grid():
     body, = html.children
     div, = body.children
     column1, column2 = div.children
+
+
+@assert_no_logs
+def test_columns_first_letter():
+    # Regression test for #2717.
+    page, = render_pages('''
+      <style>
+        h1:first-letter { font-size: 120% }
+        div { columns: 2 }
+      </style>
+      <div>
+        <h1>test</h1>
+        <h1>test</h1>
+      </div>
+    ''')
+    html, = page.children
+    body, = html.children
+    div, = body.children
+    column1, column2 = div.children
+    h1_1, = column1.children
+    h1_2, = column2.children
+    assert h1_1.element_tag == h1_2.element_tag == 'h1'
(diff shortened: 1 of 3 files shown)


liZe:
Fix #2580.

liZe:
Thanks for this pull request.

PR #2742, opened 2026-04-22
Fix flex cross size min/max

moreaki:
## Summary

This PR makes auto-height row flex containers respect cross-size `min-height` and `max-height` constraints when their height is computed from wrapped flex lines.

It also makes the clamped auto height available before `align-content: stretch`, so wrapped row flex lines can stretch into a `min-height`-expanded container.

Part of #2400.

## Root Cause

For multi-line flex containers, the final auto cross size was computed as the sum of flex line cross sizes plus gaps, but it was not clamped by the flex container's cross-axis min/max properties. In row flex containers, this meant wrapped containers ignored `min-height` and `max-height` during final height computation.

The computed auto height was also set after the `align-content: stretch` step, so stretch could not distribute extra height introduced by `min-height`.

## Changes

- Add a small helper for clamping computed cross sizes with the container's `min-*` and `max-*` cross-axis properties.
- Reuse it for the existing single-line cross-size clamp.
- Clamp multi-line auto cross sizes before final assignment.
- For auto-height row flex containers, set the clamped height before `align-content: stretch` runs.
- Add tests for wrapped row flex containers with `min-height`, `max-height`, and stretch into a `min-height` container.

## Known Remaining Scope

This intentionally covers a small, reviewable slice of the larger flex min/max topic. Remaining gaps include:

- Column flex cross size (`min-width` / `max-width`) in all auto-width and wrapped-column cases.
- Inline flex cross-size clamping and baseline interactions.
- Definite cross-size cases with conflicting min/max constraints.
- Pagination interactions when min/max cross-size changes available page height.
- Broader intrinsic keyword support such as `min-content`, `max-content`, and `fit-content()`.

## Validation

- `venv/bin/python -m pytest tests/layout/test_flex.py`
- `venv/bin/python -m ruff check weasyprint/layout/flex.py tests/layout/test_flex.py`

### DIFF
--- a/tests/layout/test_flex.py
+++ b/tests/layout/test_flex.py
@@ -53,6 +53,55 @@ def test_flex_direction_row_min_height():
 
 
 @assert_no_logs
+def test_flex_direction_row_wrap_min_height():
+    page, = render_pages('''
+      <article style="display: flex; flex-wrap: wrap; min-height: 100px; width: 4px">
+        <div style="height: 10px; width: 4px"></div>
+        <div style="height: 10px; width: 4px"></div>
+      </article>
+    ''')
+    html, = page.children
+    body, = html.children
+    article, = body.children
+    assert article.height == 100
+    div1, div2 = article.children
+    assert div1.position_y == 0
+    assert div2.position_y == 50
+
+
+@assert_no_logs
+def test_flex_direction_row_wrap_min_height_stretch():
+    page, = render_pages('''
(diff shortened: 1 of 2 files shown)


liZe:
Thanks for the pull request.

Having step 15 between steps 8.3 and 9 makes me think that this is a hack. The fix is interesting, but we have to stay close to the algorithm of the specification.

PR #2743, opened 2026-04-22
Fix forced breaks in flex rows

moreaki:
## Summary

This PR adds a narrow forced-break path for row flex containers. When a row flex item requests a forced page break before or after another item, the flex line is split at that item boundary and the remaining items resume on the next page.

Part of #2400.

## Root Cause

Flex layout already delegates final child layout to the normal block layout machinery, but it did not inspect break opportunities between flex items. As a result, `break-before: page` and `break-after: page` on row flex items were ignored, even though block and table layout already honor the same break values between sibling boxes.

## Changes

- Reuse `block_level_page_break` and `force_page_break` to detect forced page breaks between row flex items.
- Split the current flex line at the forced break boundary.
- Resume layout from the first item after the forced break on the next page.
- Add coverage for `break-before: page`, `break-after: page`, and a wrapped-row boundary.

## Known Remaining Scope

This intentionally handles a small, reviewable slice of flex breaks. Remaining work includes:

- `break-before` on the first flex item, which needs parent/container-level coordination to avoid pointless blank pages.
- Full `left` / `right` / `recto` / `verso` behavior and page-name selection tests.
- `avoid` values, which require finding earlier/later legal flex-line break opportunities.
- Column flex direction interactions without regressing existing column fragmentation behavior.
- `break-inside: avoid` for row flex items, especially when one item fragments and another should remain unbroken.
- Interactions between forced sibling breaks and already-fragmented flex items.
- Nested table/grid/flex children whose own break signals need to compose with flex-line breaks.

## Validation

- `venv/bin/python -m pytest tests/layout/test_flex.py`
- `venv/bin/python -m ruff check weasyprint/layout/flex.py tests/layout/test_flex.py`

### DIFF
--- a/tests/layout/test_flex.py
+++ b/tests/layout/test_flex.py
@@ -1762,6 +1762,103 @@ def test_flex_auto_break_before():
 
 
 @assert_no_logs
+def test_flex_direction_row_break_before_page():
+    page1, page2 = render_pages('''
+      <style>
+        @page { size: 20px; margin: 0 }
+        body { font: 2px/1 weasyprint; margin: 0 }
+        article { display: flex; width: 20px }
+        div { width: 4px }
+      </style>
+      <article>
+        <div>A</div>
+        <div style="break-before: page">B</div>
+        <div>C</div>
+      </article>
+    ''')
+    html, = page1.children
+    body, = html.children
+    article, = body.children
+    div, = article.children
+    assert div.children[0].children[0].text == 'A'
+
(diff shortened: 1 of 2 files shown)


liZe:
From [the specification](https://drafts.csswg.org/css-flexbox/#pagination):

> In a row flex container, Class A break opportunities occur between sibling flex lines

It means that forced breaks are allowed between lines, not between items.

PR #2744, opened 2026-04-22
Support context paint in SVG markers

### LINKED ISSUE #2667: Support context-stroke and context-fill SVG painting values
When a SVG marker is used, and the marker uses the "context-stroke" to inherit the stroke color, instead other properties like dash-array are taken instead. Only the color should be taken (as per spec, and other renderer like Firefox).

Minimal example:

```
<svg viewBox="0 0 100 20" xmlns="http://www.w3.org/2000/svg">
  <marker
    id="circle"
    markerWidth="6"
    markerHeight="6"
    refX="3"
    refY="3"
    markerUnits="strokeWidth">
    <circle cx="3" cy="3" r="2" stroke="context-stroke" fill="context-fill" />
  </marker>
  <path d="M 10,10 h 50"  style="marker: url(#circle);stroke:#ff0000;stroke-dasharray:6,6;stroke-dashoffset:0" />
</svg> 
```

in Firefox, that results in the marker color take the the line definition

Image

in Weasyprint on the contrary the color stay the sane, while it inherit the dash style:

Image

moreaki:
Fixes #2667.

## Context

SVG 2 adds `context-fill` and `context-stroke` paint values so marker contents can use the fill or stroke paint of the element referencing the marker. The motivating issue uses a marker circle with `stroke="context-stroke"` and `fill="context-fill"` on a dashed red path. In browsers such as Firefox, the marker takes the path paint values, but WeasyPrint did not recognize these keywords.

While investigating this, there was also an older rendering problem exposed by the same example: the dashed PDF graphics state from the referencing path could leak into marker contents. That meant a marker stroke could become dashed even if the marker child itself had no `stroke-dasharray`. This was not specific to the new SVG 2 keywords; it could happen with a plain `stroke="red"` marker too.

## Changes

- Track the marker-referencing SVG node while marker children are rendered.
- Resolve `context-fill` and `context-stroke` paint values from that context node.
- Keep the implementation scoped to marker rendering, which is the context element defined by the reported issue and the SVG 2 paint spec.
- Reset the SVG dash state to a solid stroke when the current node has no valid `stroke-dasharray`, preventing marker strokes from inheriting a dashed graphics state from the referencing path.
- Add regression tests for both the SVG 2 context-paint behavior and the pre-existing dashed-marker state leak.

## Scope and follow-up

This is intentionally a first, narrow slice of SVG 2 context paint support. It does not try to implement every context-paint case from the spec. In particular, `use`-element context paint and paint-server coordinate-space continuity for contextual paint values remain follow-up work.

## Tests

- `venv/bin/python -m pytest tests/draw/svg/test_markers.py`
- `venv/bin/python -m pytest tests/draw/svg`
- `venv/bin/python -m ruff check weasyprint/svg/__init__.py tests/draw/svg/test_markers.py`

### DIFF
--- a/tests/draw/svg/test_defs.py
+++ b/tests/draw/svg/test_defs.py
@@ -56,3 +56,26 @@ def test_use_symbol_color(assert_pixels):
     svg = SVG.replace('fill="blue"', '')
     svg = svg.replace('href="#square"', 'href="#square" fill="blue"')
     assert_pixels(RESULT, STYLE + svg)
+
+
+@assert_no_logs
+def test_use_context_paint(assert_pixels):
+    assert_pixels('''
+        RRBB__GGMM
+        RRBB__GGMM
+    ''', '''
+      <style>
+        @page { size: 10px 2px }
+        svg { display: block }
+      </style>
+      <svg width="10px" height="2px" xmlns="http://www.w3.org/2000/svg">
+        <defs>
+          <g id="rectangle" stroke="none">
+            <rect width="2" height="2" fill="context-fill" />
+            <rect x="2" width="2" height="2" fill="context-stroke" />
+          </g>
+        </defs>
+        <use href="#rectangle" fill="red" stroke="blue" />
+        <use href="#rectangle" x="6" fill="lime" stroke="magenta" />
+      </svg>
+    ''')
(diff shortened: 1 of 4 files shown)


liZe:
Thanks for the pull request.

I rewrote the context management to avoid the global context attribute and bind the paint data management to the node. It’s more simple, and makes the support of `use` tags more obvious than #2745.

PR #2745, opened 2026-04-22
Support context paint in SVG use elements

moreaki:
Stacked on #2744.

## Context

#2744 adds marker-scoped support for the SVG 2 `context-fill` and `context-stroke` paint values, fixing the concrete marker case from #2667. The SVG 2 paint definition also says that an element inside a subtree instantiated by a `use` element has that `use` element as its context element.

This draft PR adds that second small context-paint case on top of #2744, so maintainers can decide whether they want to include it as a follow-up feature.

## Changes

- Reuse the context-paint tracking introduced by #2744 when drawing children of a `use` element.
- Treat the `use` element as the context element while its referenced subtree is drawn.
- Add a pixel regression test proving that `context-fill` and `context-stroke` are resolved independently for multiple `use` instances with different solid paint values.

## Scope and limitations

This intentionally supports the simple solid-color case. It does not try to solve the harder paint-server geometry case for contextual `url(#gradient)` or `url(#pattern)` values. For those, SVG 2 requires the paint server coordinate space and bounding box to come from the context element, which needs additional gradient/pattern plumbing beyond this small follow-up.

Until #2744 lands, this PR’s diff includes the marker context-paint groundwork from #2744 plus one additional commit for `use` elements.

## Tests

- `venv/bin/python -m pytest tests/draw/svg/test_defs.py::test_use_context_paint tests/draw/svg/test_defs.py::test_use_symbol_color tests/draw/svg/test_markers.py::test_markers_context_paint`
- `venv/bin/python -m pytest tests/draw/svg`
- `venv/bin/python -m ruff check weasyprint/svg/__init__.py tests/draw/svg/test_defs.py tests/draw/svg/test_markers.py`

### DIFF
--- a/tests/draw/svg/test_defs.py
+++ b/tests/draw/svg/test_defs.py
@@ -56,3 +56,26 @@ def test_use_symbol_color(assert_pixels):
     svg = SVG.replace('fill="blue"', '')
     svg = svg.replace('href="#square"', 'href="#square" fill="blue"')
     assert_pixels(RESULT, STYLE + svg)
+
+
+@assert_no_logs
+def test_use_context_paint(assert_pixels):
+    assert_pixels('''
+        RRBB__GGMM
+        RRBB__GGMM
+    ''', '''
+      <style>
+        @page { size: 10px 2px }
+        svg { display: block }
+      </style>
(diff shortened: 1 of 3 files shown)


liZe:
Included in #2744.

Issue #1652, opened 2022-05-30
Inline-block as child of display:flex element overflows since 54


Christoph-Wagner:
When a parent has `display:flex`, a child with `display:inline-block` gets a base-width that far overflows both the parent and the current page. This behavior was introduced in 54 and still happens with 55.  53.4 is working correctly.

I made a minimal example:

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
    <title>empty</title>
    <style>
    .testbox {
        display: flex;
        background-color: #eee;
    }
    .testbox p {
        display: inline-block;
    }
    </style>
</head>
<body>
<div class="testbox">
    <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit,
        sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud
        exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in
        voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident,
        sunt in culpa qui officia deserunt mollit anim id est laborum.
    </p>
</div>
</body>
</html>
```

This results in

![image](https://user-images.githubusercontent.com/1609329/170970497-2dc67961-e720-4718-985e-6dc173a70171.png)

If I add a percentage width, it takes some other, unknown width as basis (`width: 50%`):

![image](https://user-images.githubusercontent.com/1609329/170970737-9f8409dc-2d49-45a8-861b-7e3545a5facb.png)

Changing it to `display:block` fixes everything:

![image](https://user-images.githubusercontent.com/1609329/170970608-b4075cf9-ad90-4fde-bf40-fc4c2c55d31a.png)

n1ngu:
Related issue #1563

liZe:
That was hours of debugging… to change 2 characters!

It may fix other issues, for example for inline-block elements in floats.

Bug is closed and tested in the flex branch, see https://github.com/Kozea/WeasyPrint/issues/2362.

PR #2746, opened 2026-04-22
Support basic display: contents

### LINKED ISSUE #2210: Flexbox CSS 'order' is ignored
I'm using Material for MkDocs with PyMdown Extensions and mkdocs-with-pdf.  PyMdown has a tabbed content feature. This is a simplified example of the structure of the tabbed content with minimal CSS for print. 

``` html
<div style="display: flex; flex-flow: column wrap">
  <div style="display:contents">
    <label style="order: 1;">Tab 1</label>
    <label style="order: 2;">Tab 2</label>
  </div>
  <div style="display: contents">
    <div style="order: 1; display block; ">
    Content Block 1
    </div>
    <div style="order: 2; display block; ">
    Content Block 2
    </div>
  </div>
</div>
```

In the browser this is displayed as:

> Tab 1
> Content Block 1
> Tab 2
> Content Block 2

Weasyprint renders this as:
> Tab 1 Tab 2
> Content Block 1
> Content Block 2

Even placing all the 'order' elements directly under the 'display: flex' element gives the same result.

I proposed a workaround to [PyMdown](https://github.com/facelessuser/pymdown-extensions/pull/2420), but this should be resolved here.

moreaki:
Fixes #2210.

## Context

#2210 reports that tabbed content generated by Material for MkDocs / PyMdown Extensions is rendered in the wrong order when wrappers use `display: contents` inside a flex container. WeasyPrint currently rejects `display: contents` during CSS validation, so these wrappers generate real boxes instead of being flattened into their children.

The issue thread also mentioned that the same order problem could be reproduced without `display: contents` when inline `label` elements are direct flex children. On current `main`, that direct-inline case is already handled correctly by the existing flex item/blockification work related to #1652. The remaining missing part is the actual `display: contents` wrapper behavior.

## Changes

- Accept `display: contents` as a valid `display` value.
- Compute `display: contents` on the root element back to a normal block-flow root box, avoiding a missing root box.
- Flatten normal `display: contents` elements by returning their child boxes instead of a principal box.
- Handle pseudo-element and marker edge cases so the new display value does not fall through to box creation paths that have no `contents` box type.
- Add a regression test for the #2210 flex/tabbed-content structure, checking that the children are flex items in the expected order: Tab 1, Content Block 1, Tab 2, Content Block 2.

## Scope and limitations

This is intended as a first, narrow implementation of `display: contents`. It fixes the concrete wrapper/flex case from #2210 and keeps the change small enough to review.

It does not claim complete CSS Display conformance. In particular, this does not deeply address every possible interaction of `display: contents` with accessibility/tree semantics, unusual pseudo-element behavior, replaced elements, or every layout model. The goal is to provide a useful first fix and let maintainers decide whether this approach is acceptable as a base for broader support.

## Tests

- `venv/bin/python -m pytest tests/layout/test_flex.py tests/css/test_validation.py`
- `venv/bin/python -m ruff check weasyprint/css/validation/properties.py weasyprint/css/computed_values.py weasyprint/formatting_structure/build.py tests/layout/test_flex.py`
- `git diff --check`

### DIFF
--- a/tests/layout/test_flex.py
+++ b/tests/layout/test_flex.py
@@ -1028,6 +1028,31 @@ def test_flex_direction_row_inline_block():
 
 
 @assert_no_logs
+def test_flex_display_contents():
+    # Regression test for issue #2210.
+    page, = render_pages('''
+      <article style="display: flex; flex-flow: column wrap">
+        <div style="display: contents">
+          <label style="order: 1">Tab 1</label>
+          <label style="order: 2">Tab 2</label>
+        </div>
+        <div style="display: contents">
+          <div style="order: 1">Content Block 1</div>
+          <div style="order: 2">Content Block 2</div>
+        </div>
+      </article>
+    ''')
+    html, = page.children
+    body, = html.children
+    article, = body.children
+    tab_1, content_1, tab_2, content_2 = article.children
+    assert tab_1.children[0].children[0].children[0].text == 'Tab 1'
+    assert content_1.children[0].children[0].text == 'Content Block 1'
+    assert tab_2.children[0].children[0].children[0].text == 'Tab 2'
+    assert content_2.children[0].children[0].text == 'Content Block 2'
+
+
+@assert_no_logs
 def test_flex_float():
     page, = render_pages('''
       <article style="display: flex">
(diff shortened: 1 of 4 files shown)


liZe:
Thanks for the pull request.

I feel like it’s possible to get a reliable implementation with a few lines of code, and this implementation is not far from a great solution. But we should do better.

If someone wants to implement this feature, it should be done in `make_box`. As it’s been done in the past for other functions like `element_to_box` or `before_after_to_box`, we should return an iterable of boxes instead of a single box.

PR #2747, opened 2026-04-22
Warn for invalid CSS at-rules

### LINKED ISSUE #2215: @footnote::before and other broken at-rules not discarded
The [spec](https://www.w3.org/TR/css-gcpm-3/#footnotes) mentions the "footnote rule", but doesn't actually specify how to create it.

I tried to create a 20% wide rule with @footnote::before:

```
@page
   @footnote::before {
        display: block;
        width: 20%;
        border-top: solid 0.2mm;
    }
}
```

but the "::before" seems to be stripped and the rule applied to the entire area instead:

ss

Is it reasonable to expect "::before" to apply to the footnote area?

moreaki:
Fixes #2215.

## Context

#2215 reports that invalid at-rules are accepted too loosely. The original example uses `@page { @footnote::before { ... } }`: because `::before` is a pseudo-element and `@footnote` is a page-area at-rule, the rule should be rejected with a warning instead of being treated as a valid `@footnote` rule.

A later issue comment listed more malformed examples that were silently ignored or accepted, including `@import::before`, page-area rules with invalid preludes, and unknown nested `@page` rules. Some cases were already handled by f992fd5, but these examples still needed targeted checks.

## Changes

- Add an explicit list of supported nested `@page` area rules, including margin boxes and `@footnote`.
- Warn and discard syntactically invalid `@import` rules such as `@import::before "test.css"` without adding duplicate warnings for URL-resolution failures that are already reported elsewhere.
- Warn and discard unknown nested `@page` at-rules such as `@unknown`.
- Warn and discard nested `@page` area rules with invalid preludes, including `@top-left-corner super` and `@footnote::before`.
- Keep valid rules such as `@page { @footnote { ... } }` and `@page { @top-left { ... } }` warning-free.

## Scope

This is a targeted follow-up to the previous warning work. It covers the concrete cases reported in #2215 and its comments, but it is not intended as a full audit of every malformed CSS at-rule in every possible context.

## Tests

- `venv/bin/python -m pytest tests/css/test_errors.py::test_warnings tests/css/test_pages.py`
- `venv/bin/python -m pytest tests/css`
- `venv/bin/python -m ruff check weasyprint/css/__init__.py tests/css/test_errors.py`
- `git diff --check`


liZe:
Thanks for the pull request.

Issue #2308, opened 2024-11-22
white-space: pre-wrap wraps spans prematurely


nnsee:
Hi there! Spans seem to be wrapped to a new line in their entirety when `white-space: pre-wrap` is set.

Repro code:

```html
<style>
.highlight {
  background-color: #ffff00;
}
code {
  white-space: pre-wrap;
  display: block;
  max-width: 250px;
}
</style>
<code>
00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
</code>
<hr>
<code>
00 00 00 00 00 00 00 00 00 00 <span class="highlight">00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00</span> 00 00 00 00 00
</code>
```

WeasyPrint output:

![Image](https://github.com/user-attachments/assets/ffb3581a-1292-4727-a0fc-3f87e7e61eed)

Correct output:

![Image](https://github.com/user-attachments/assets/8eaeaf8e-088f-4070-a117-22c2e220268b)

Not quite sure why this is happening or how to stop it from happening.

liZe:
Thanks for the report. That’s probably related to #2025.

moreaki:
I opened #2748 with a small inline line-breaking fix and regression coverage for this example.

On current `main`, the concrete example from this issue already appears to render without moving the whole highlighted span to the next line, but the PR adds a regression test so this behavior is kept covered. The actual code change targets the broader linked cause from #1614: deciding whether `overflow-wrap: break-word` may split a word across inline box boundaries by checking whether the current line has already had a real break opportunity, rather than checking only whether this is the first box on the line.

The PR intentionally does not claim full CSS Text line-breaking compliance. It does not replace the broader work explored in #1840, and it does not solve every interaction with shaping, bidi, CJK rules, hyphenation, `word-break`, `line-break`, and all `overflow-wrap` cases. It is meant as a small targeted fix for the immediate inline-boundary trigger, with coverage for #1614, #2025, and this issue.

Validation run locally: full test suite passed with `4028 passed, 40 xfailed`.

sanderdejongg:
Hey guys, is there any update on this issue?

grewn0uille:
Guess not as the issue is still open :)

Issue #2102, opened 2024-03-19
word-wrapping bug 


survtur:
That long word breaks if i wrap it into additional span. 
Tested on Weasy 61.2

![image](https://github.com/Kozea/WeasyPrint/assets/4893428/47acc853-db9e-43b3-85f9-4ee7fb9ede4f)

```html
<body style="  word-wrap: break-word">
<div style="width: 20em; background-color: #fafaff">
    <em>
        some text
        <span style='color: red'>xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx yyyyyyyyy</span>
        zzzz wwwwwwwwwwwww
        aaaaaa
    </em>
</div>
<br/><br/>
<div style="width: 20em; background-color: #fafaff">
    <em>
        some text
        xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx yyyyyyyyy
        zzzz wwwwwwwwwwwww
        aaaaaa
    </em>
</div>
</body>
```

liZe:
Maybe related to #1614.

moreaki:
I updated #2748 to include this case too.

The added regression test keeps the two blocks from this report equivalent: a long word wrapped in an extra inline span now breaks the same way as the same text without that span. The implementation carries the previous inline fragment state into nested inline boxes, so a break opportunity before the span is still considered when laying out the text inside the span.

This is still a targeted inline-boundary fix rather than full CSS Text line-breaking support. Full suite passed locally after the update: `4029 passed, 40 xfailed`.

Issue #2025, opened 2023-12-22
white-space: pre-wrap; works depending on tag position


galech:
Hi everyone,

depending on space before or after tag (`</b> domicilié`, ` </b>domicilié`) `white-space: pre-wrap` can fail

![image](https://github.com/Kozea/WeasyPrint/assets/10774285/1880d6f1-a20f-49ed-b918-f165478155a7)

```
<style style="text/css">


@page {
	size: A4 ;
	margin-left: 18px;
	margin-right: 18px;
	margin-bottom: 100px;
	margin-top: 50px;

}

pre {
	white-space: pre-wrap;
	border: 2px solid black;
	background: inherit;
	font-family: inherit;
	font-size: inherit;
}

</style>


<body>
	<div style="font-size:13px">
		<div>
			<pre>Cher Confrère,<br><br>Je vous informe de la prise en charge en HAD de <b>hhhhhhh 11111111 (né(e) CLAIRE), né le 2 déc. 2022, </b>domicilié fsdfsdfsdfsdfds </pre>
			<pre>Cher Confrère,<br><br>Je vous informe de la prise en charge en HAD de <b>hhhhhhh 11111111 (né(e) CLAIRE), né le 2 déc. 2022,</b> domicilié fsdfsdfsdfsdfds </pre>
		</div>
	</div>
</body>
```

Thank you for your great job

liZe:
Thanks for the bug report.

That’s strange, we have so many tests to check that this couldn’t happen! We probably need one more…

liZe:
Maybe related to #1614.

moreaki:
I opened #2748 with a small inline line-breaking fix and regression coverage for this report.

The PR includes a test checking that the two `white-space: pre-wrap` variants from this issue, with the preserved space before or after the inline tag boundary, produce the same line text. The code change itself targets the broader linked cause from #1614: `overflow-wrap: break-word` should decide whether it is still in the first otherwise-unbreakable sequence by checking real break opportunities already present on the line, not simply whether the current text is in the first box.

This is deliberately not a full CSS Text implementation. It does not replace the larger line-breaking rewrite explored in #1840 and does not claim to cover all shaping, bidi, CJK, hyphenation, `word-break`, `line-break`, or `overflow-wrap` interactions. It is a small targeted fix with regression coverage for #1614, #2025, and #2308.

Validation run locally: full test suite passed with `4028 passed, 40 xfailed`.

Issue #1614, opened 2022-03-25
Incorrect line breaking with boxes


aschmitz:
At a high level, there are two separate issues that are currently colliding:

1. We [almost] treat inline boxes (like `<span>`s) as if they were separate words for wrapping purposes.
2. We don't currently [always] have a way to wrap them properly even if they were.

A minimal test document may be helpful:

~~~html
<style>
  html { overflow-wrap: break-word; }
  div { background-color: #abcdef; width: 20em; }
</style>

<div>
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa<span>bbbbbbbbbb</span>
</div>
~~~

In Firefox (much like Chrome), this wraps in the middle of the b's:

![The sample document in Firefox](https://user-images.githubusercontent.com/29508/160039189-2d176447-1872-49e8-b172-e2587450059f.png)

In WeasyPrint (as of the current `master`, 2f8d4999d70c9fa2ea26d7accc30a75481a22542), we never wrap:

![Screenshot from 2022-03-24 20-54-06](https://user-images.githubusercontent.com/29508/160039240-e6c02e16-9a49-4554-abdc-17759d83c776.png)

I'm currently hitting this with some syntax-highlighted content that should wrap but instead runs off the page, for a sample real-world use case.

This appears to be happening because split_first_line is using the `is_line_start` parameter to determine whether something is the line's first word. It turns out that we set this a bit naïvely in split_inline_level such that it's only true if the box is the first *box* on the line rather than if it is in the first ["otherwise unbreakable sequence of characters"](https://drafts.csswg.org/css-text-3/#propdef-overflow-wrap).

I suspect the "right" way to fix this is to return a value indicating whether there has been a space in the line so far or not, and setting is_line_start accordingly. That's _probably_ straightforward enough, but I had expected that doing so would mean we wouldn't correctly backtrack if there *were* a space in the previous text box. This turns out to be false: even now, changing one of the "a"s to a space allows the line to break at that position, giving a more appropriate render (which also matches the major browsers' behavior). Clearly I've missed some detail of how line breaking works in that circumstance, and I'm hesitant to just pile on an additional patch until I understand the underlying behavior a bit better.

[...]

PR #2748, opened 2026-04-22
Fix inline break opportunities across boxes

moreaki:
## What changed

This updates inline line breaking so `overflow-wrap: break-word` decides whether a text fragment is at the start of an otherwise unbreakable sequence by checking for real break opportunities already present on the line, instead of only checking whether this is the first box on the line. The state is also carried into nested inline boxes so a previous break opportunity outside a span is still visible when laying out text inside that span.

That fixes the linked case from #1614 where a single long word is split across a text node and an inline box, for example `aaaaaaaa...<span>bbbb...</span>`. The text can now wrap inside the continued word instead of overflowing because the span boundary was treated too much like a word boundary.

The PR also adds regression coverage for related inline-boundary reports from #2102, #2025, and #2308. Current `main` already appears to handle the concrete #2308 example, but the test pins the behavior so it does not regress.

## Historical context

The issue chain is:

- #2308 reports that highlighted spans with `white-space: pre-wrap` can wrap as whole spans.
- #2025 reports different `pre-wrap` behavior depending on whether a preserved space is inside or outside an inline tag.
- #2102 reports that wrapping a long word in an extra span changes `word-wrap: break-word` rendering.
- #1614 identifies the broader line-breaking problem: WeasyPrint can treat inline boxes almost as separate words, and the existing `is_line_start` heuristic is based on box position rather than whether the current line already has an acceptable break opportunity.

There is also the older draft #1840, which explores a much broader line-breaking rewrite using Harfbuzz data. Daniel Fitzpatrick also has a fork commit for the same root problem: https://github.com/danfitz36/WeasyPrint/commit/f9e68e2afdda1e62076008482643cf01bd5f7df3. This PR is intentionally much smaller than #1840, and now also carries the previous fragment state into nested inlines as Daniel’s commit does.

## What this does not claim

This is not full CSS Text line-breaking compliance. It does not implement the complete CSS Text Level 3 model, nor does it solve every interaction between inline boundaries, shaping, bidi, CJK line-breaking rules, hyphenation, `word-break`, `line-break`, and all `overflow-wrap` cases.

The goal here is narrower: fix the immediate inline-boundary triggers reported in #1614 and #2102, and keep the related #2025/#2308 examples covered by tests, while leaving the broader line-breaking universe for a larger implementation.

## Tests

- `venv/bin/python -m ruff check weasyprint/layout/inline.py tests/test_text.py`
- `venv/bin/python -m pytest tests/test_text.py`
- `venv/bin/python -m pytest tests/layout/test_inline.py`
- `venv/bin/python -m pytest` -> 4029 passed, 40 xfailed

Addresses #1614 and #2102.
Related to #2025 and #2308.

### DIFF
--- a/tests/test_text.py
+++ b/tests/test_text.py
@@ -20,6 +20,12 @@ def make_text(text, width=None, **style):
         justification_spacing=0)
 
 
+def box_text(box):
+    if hasattr(box, 'text'):
+        return box.text
+    return ''.join(box_text(child) for child in box.children)
+
+
 @assert_no_logs
 def test_line_content():
     for width, remaining in [(100, 'text for test'),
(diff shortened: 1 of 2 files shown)


moreaki:
Follow-up / prior-art note: I found Daniel Fitzpatrick’s fork commit https://github.com/danfitz36/WeasyPrint/commit/f9e68e2afdda1e62076008482643cf01bd5f7df3, dated 2026-04-04, which addresses the same root `is_line_start` problem for inline span boundaries. I didn’t find an open or closed PR from that branch against this repository, and it has not been merged here.

That commit and this PR use the same underlying idea: `overflow-wrap: break-word` should decide whether a fragment is still in the first otherwise-unbreakable sequence from actual break opportunities, not from “first box on the line”. Daniel’s patch threads the previous `last_letter` through `split_inline_box` / `split_inline_level`; this PR keeps the existing signatures and computes the state from the already-laid-out line children via `line_has_break()`, while also sharing the existing “can break between these fragments” logic.

Daniel’s commit also references #2102, which is another nested-span word-breaking report related to #1614. If the maintainers prefer the explicit `last_letter` threading approach, that fork commit is a useful alternative/reference. If this PR stays as-is, I’m happy to add an extra #2102 regression test too.

moreaki:
I updated the PR with a #2102 regression and extended the implementation to carry previous-fragment state into nested inline boxes. This makes the #2102 span-wrapped long-word case break identically to the same text without the span.

Re-ran validation after the update:

- `venv/bin/python -m ruff check weasyprint/layout/inline.py tests/test_text.py`
- `venv/bin/python -m pytest tests/test_text.py`
- `venv/bin/python -m pytest tests/layout/test_inline.py`
- `venv/bin/python -m pytest` -> 4029 passed, 40 xfailed

liZe:
Thanks for the pull request.

The line breaking algorithm needs deep changes. Until we drop Pango layouts and rewrite the algorithm, we’ll keep the current implementation to avoid possible regressions.

Issue #2647, opened 2026-01-06
grid + float = overlapping text


AstraLuma:
I'm not familiar enough the innards of CSS layout to understand this bug, but I figured I'd file it in case someone else finds it.

Example: https://git.gay/AstraLuma/resume/src/commit/7283815cb816ff44e4eba16c355aba4322e8b2b4 (it's not strictly minimized, but it's not that big)

firefox (browser):

<img width="706" height="385" alt="Image" src="https://github.com/user-attachments/assets/14c3d99b-af4b-4aa3-b155-74c8d9ecbe96" />

firefox (print preview):

<img width="574" height="329" alt="Image" src="https://github.com/user-attachments/assets/11d6025e-58af-42ea-81fd-c46e9323c0bd" />

weasyprint:

<img width="770" height="368" alt="Image" src="https://github.com/user-attachments/assets/ba4d3ed7-1d7c-411f-af4d-eaf377f84ded" />

of note, these are all `<section>` siblings with `display: grid`, but one is also `float: right`

This is Firefox v145.0.1 (Linux) compared to Weasyprint v66.0

AstraLuma:
Weasyprint v67:

<img width="739" height="369" alt="Image" src="https://github.com/user-attachments/assets/624c8c9e-fd34-4934-85db-619d39979556" />

liZe:
Hi,

Thanks for the report and for the useful sample!

Grid and float at the same time, here’s another issue to link in #2145.

Issue #2644, opened 2026-01-05
Handle grid items with intrinsic size


liZe:
Example:

```html
<div style="display: grid; grid-template-columns: 50px 1fr;">
  <img style="width: 20px" src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAQAAAAECAIAAAAmkwkpAAAAE0lEQVR4nGP4zwAE/2EIwcLDAQCQsQ/xhr9I3AAAAABJRU5ErkJggg==">
  <span>Test</span>
</div>
```

Related to #2628.

Issue #2628, opened 2025-12-12
grid and flex rendering issues


lex-zhu:
Hello !

I've noticed some bugs when upgraded from v64.1 to v67:

1) box-sizing: border-box on the flex block makes the inner content of this block shrinked by the amount of the padding. It's just easier to show, the text with a red background is shrinked in the example, plus the block doesn't take all the hight of the parent.
2) bizarre additional margins between grid lines (bottom block, contacts section)
3) the image is stretched to the grid width ignoring it's own width (for example price tag icon)

Again, all the above wasn't an issue in v64.1.
p1 seems to be introduced with v65-66 (hard to say since grid was quite broken in these versions).
p2 starts with the version 67.
p3 starts with the version 66.

Here are examples (an html source file to reproduce the issues above, and renders from different versions)

The original html:
[weasy-test.html](https://github.com/user-attachments/files/24125683/weasy-test.html)

Renders:
[weasy-test-64.pdf](https://github.com/user-attachments/files/24125682/weasy-test-64.pdf)
[weasy-test-65.pdf](https://github.com/user-attachments/files/24125680/weasy-test-65.pdf)
[weasy-test-66.pdf](https://github.com/user-attachments/files/24125684/weasy-test-66.pdf)
[weasy-test-67.pdf](https://github.com/user-attachments/files/24125681/weasy-test-67.pdf)

Thanks in advance !

liZe:
Hi,

Thanks for the report!

- The first point is caused by a poor support of `box-sizing` in grid layouts. Pull request is coming.
- The second point is caused by #2354, listed in #2145.
- The third point is caused by "grid items with intrinsic size (images)" in #2145, I’ll open a dedicated issue with an example.

I’ll close this issue with the pull request, and we’ll track the two other bugs in the dedicated issues.

Issue #2397, opened 2025-03-04
Elements in Grids Overflow Incorrectly


stefocean:
When using `display: grid`, there are situations where the content of a single grid column should be displayed on more than one page. Currently, Weasyprint does not have support for this, and rendered grid columns in PDFs can only exist on one page of output at a time and does not overflow to the next page.

What we would like to see from the PDF output is for the content in a column to overflow to subsequent pages, including decorations such as borders. Furthermore, columns currently try to place themselves on their own page in some situations (see below).

Bug is reproducible through the following text and styling content:

### HTML
```
<!DOCTYPE html>
<html lang="en">
    <head>
        <link rel="stylesheet" href="index.css">
    </head>
<body>
    <div class="grid-container">
        <div>
            THIS SHOULD BE DISPLAYED IN ROW 1<br>
            THIS SHOULD BE DISPLAYED IN ROW 1<br>
            THIS SHOULD BE DISPLAYED IN ROW 1
        </div>
        <div>
[...]

PR #2749, opened 2026-04-22
Fix simple grid content widths

moreaki:
## What changed

This adds a small intrinsic-width helper for grid containers used by preferred width calculations.

Before this change, `min_content_width()` and `max_content_width()` treated `GridContainerBox` like a regular block container. That means shrink-to-fit contexts, such as floats and flex/grid items using intrinsic size contributions, could see a multi-column grid as if its intrinsic width were just the width of one child block. In the minimal example discussed in #2354/#2145, a two-column grid containing `abc def` twice reported a width close to one `abc def`, causing each grid item to wrap to `abc`/`def`.

The new helper handles simple, non-spanning row-flow grid containers by summing per-track intrinsic contributions and gutters. It is deliberately conservative: it improves common grid content width cases without attempting to implement the complete CSS Grid intrinsic sizing algorithm.

## Investigation summary

I followed the #2354 link chain and related work before making this patch:

- #2354 was opened as a follow-up to #2342, replacing the earlier flex/grid mixed example with a grid-only example using nested grids and auto-height rows.
- #2342 links these separate limitations:
  - #1891: `aspect-ratio` is unsupported.
  - #1665: flex `gap` support, later handled by the flex rewrite.
  - #1652 and #1563: flex item sizing/margin issues, later handled by the flex rewrite.
  - #2145: the grid support tracking issue.
  - #2362 / #2387: the flex rewrite, now merged.
- #2145 explicitly tracks #2354 as “auto content size for grid containers”. It also tracks nearby grid gaps including #2644 (“grid items with intrinsic size”, eg images), #2647 (grid + float), #2397/#2564 (grid row fragmentation, merged), #2628/#2643 (`box-sizing: border-box` in grid layout, merged), and #2142/#2162 (`grid-auto-flow: column`, merged).
- I searched for open and closed upstream PRs around #2354/#2342 and the relevant grid terms. I did not find an existing open PR or submitted patch specifically for #2354.

## What this fixes

The PR fixes the maintainer’s minimal intrinsic-size example from #2354/#2145:

```html
<div style="float: left">
  <div style="display: grid; grid: 1fr / 1fr 1fr; border: 1px solid red">
    <div>abc def</div>
    <div>abc def</div>
  </div>
</div>
```

With this patch, the float’s max-content width includes both grid tracks, and the two `abc def` items no longer wrap as if the grid were one child block wide.

The PR now also includes the exact flex-column/grid sample from the #2354 issue comment as `test_grid_container_in_flex_column`. On current main, that standalone sample already lays out as a stretched flex item, so this test is included as related reference coverage; the behavior-changing regression remains the shrink-to-fit grid content width case above.

## What this does not fix

This is not full CSS Grid intrinsic sizing support. In particular, this PR does **not** claim to solve:

- the complete original image/object-fit layout in #2354;
- grid items with intrinsic image sizes (#2644);
- `aspect-ratio` support (#1891);
- absolutely positioned or floating grid items (#2647);
- spanning items in intrinsic sizing;
- complex `min-content` / `max-content` cases;
- `auto-fit` / `auto-fill`;
- `auto` margins for grid items;
- `min-*` / `max-*` constraints on grid items;
- the complete CSS Grid track sizing algorithm for intrinsic contributions.

The intent is to provide a small reviewable base for one narrow part of #2354: simple grid containers contributing a sensible min/max-content width in shrink-to-fit contexts.

## Tests

- `venv/bin/python -m ruff check weasyprint/layout/grid.py weasyprint/layout/preferred.py tests/layout/test_grid.py`
- `venv/bin/python -m pytest tests/layout/test_grid.py`
- `venv/bin/python -m pytest tests/layout/test_preferred.py tests/layout/test_flex.py`
- `venv/bin/python -m pytest` -> 4027 passed, 40 xfailed

Addresses part of #2354.
Related to #2145, #2644, #2342, #1891, #1665, #1652, #2362, #2628, #2647.

### DIFF
--- a/tests/layout/test_grid.py
+++ b/tests/layout/test_grid.py
@@ -104,6 +104,46 @@ def test_grid_container_max_content_width():
 
 
 @assert_no_logs
+def test_grid_container_in_flex_column():
+    # Reference test for the #2354 flex/grid example.
+    page, = render_pages('''
+      <style>
+        body {font-family: weasyprint; font-size: 2px}
+      </style>
+      <div style="padding: 10px; display: flex; flex-direction: column;
+                  background: red">
+        <div style="display: grid; grid-template-columns: 1fr 1fr;
+                    align-items: center; background: pink">
+          <span style="background: lime">a</span>
+          <span style="background: cyan">ab cd ef gh</span>
+        </div>
+      </div>
+    ''')
+    html, = page.children
+    body, = html.children
+    flex, = body.children
+    grid, = flex.children
+    span1, span2 = grid.children
+    line1, = span1.children
+    line2, = span2.children
+    inline1, = line1.children
+    inline2, = line2.children
+    text1, = inline1.children
+    text2, = inline2.children
+
+    assert flex.width == body.width - 20
+    assert grid.position_x == flex.position_x + 10
+    assert grid.width == flex.width
+    assert span1.position_x == grid.position_x
+    assert span2.position_x == span1.position_x + span1.width
+    assert isclose(span1.width, grid.width / 2)
+    assert span1.width == span2.width
+    assert span1.position_y == span2.position_y
+    assert text1.text == 'a'
+    assert text2.text == 'ab cd ef gh'
+
+
+@assert_no_logs
 def test_grid_rows():
     page, = render_pages('''
       <article style="display: grid">
--- a/tests/layout/test_grid.py
+++ b/tests/layout/test_grid.py
@@ -73,6 +73,37 @@ def test_grid_single_percentage_width():
 
 
 @assert_no_logs
+def test_grid_container_max_content_width():
+    # Regression test for #2354.
+    page, = render_pages('''
+      <style>
+        body {font-family: weasyprint; font-size: 2px}
+      </style>
+      <div style="float: left">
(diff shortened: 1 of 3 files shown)


moreaki:
I added the exact flex-column/grid sample from the #2354 issue comment as `test_grid_container_in_flex_column` in commit 26e28bdc.

One note for review: this standalone sample already lays out as a stretched flex item on current `main` in my local run, so the new test is reference coverage for the related issue comment rather than the behavior-changing regression. The regression fixed by this PR is still the shrink-to-fit intrinsic grid width case, where the grid container previously contributed only one child block width and wrapped the two `abc def` items.

Validation after the added test:

- `venv/bin/python -m ruff check weasyprint/layout/grid.py weasyprint/layout/preferred.py tests/layout/test_grid.py`
- `venv/bin/python -m pytest tests/layout/test_grid.py` -> 52 passed, 1 xfailed
- `venv/bin/python -m pytest` -> 4027 passed, 40 xfailed

liZe:
> It is deliberately conservative: it improves common grid content width cases without attempting to implement the complete CSS Grid intrinsic sizing algorithm.

In this case, it may be better to have a reliable intrinsic sizing algorithm than a workaround.

Issue #2486, opened 2025-06-27
Initial Letter Support


anghenfil:
Hey,

it would be great if weasyprint could support the [initial-letter attribute](https://developer.mozilla.org/en-US/docs/Web/CSS/initial-letter), especially since manually setting the first letter bigger doesn't work (see #1859).

Best
anghenfil

liZe:
Yes, that would be very useful!

moreaki:
I opened draft PR #2752 with a first implementation of unprefixed `initial-letter` support for `::first-letter`.

The branch is stacked on #2751, which fixes floated `::first-letter` line positioning. This PR adds parsing for the CSS property, cap-height-based initial-letter sizing, basic drop/sink/raise behavior, and regression tests for validation, punctuation, styled boxes, fractional sizes, and short paragraphs.

The PR description includes visual proof from locally generated PNG/PDF artifacts: https://github.com/Kozea/WeasyPrint/pull/2752

I also opened follow-up draft PR #2753 for the MDN-style larger sink case, for example `initial-letter: 3 4`: https://github.com/Kozea/WeasyPrint/pull/2753

This is not meant to be complete CSS Inline Layout Level 3 support. Known gaps include RTL quality, first inline-level child applicability, `initial-letter-align`, `initial-letter-wrap`, vertical writing modes, fragmentation/page or column interactions, and complex float/exclusion interactions. The goal is to make the common LTR drop-cap cases useful while keeping the patches small enough to review.

PR #2751, opened 2026-04-22
Fix floated first-letter line positioning

### LINKED ISSUE #1859: Possible float render issue with a `:first-letter` pseudo element
Hi,

I'm trying to drop cap the first letter of a paragraph, but I'm running into an odd behavior. The first letter size is increased accordingly, but the first line is not shifted to the right, while the following two lines are. Chrome renders it accordingly to what I'd expect, so this may be a bug on weasyprint.

I've created a relatively small HTML file that can reproduce the issue:

```html
<!DOCTYPE html>
<html>
  <head>
    <title>test</title>
    <style>
p {
  width: 300px;
  margin: auto;
  font-size: 12px;
}

p:first-letter {
  float:left;
  font-size: 35px;
}
    </style>
  </head>
  <body>
    <p>
    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Ut at justo non elit ultrices posuere sit amet at nulla. Aliquam at fermentum nunc. Donec ac placerat eros. Etiam laoreet tempus felis, sit amet iaculis neque luctus eu. Nam eu nibh ullamcorper, aliquam sapien vel, aliquet odio. Curabitur eu viverra felis, quis aliquet mauris. Morbi ante diam, hendrerit ac vestibulum id, ultricies nec velit. Quisque ut pretium orci, ut efficitur elit. Pellentesque cursus nisl in justo vestibulum, ut facilisis diam efficitur. Orci varius natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Etiam gravida et sem sit amet facilisis. Sed pellentesque luctus tempus.
    </p>
  </body>
</html>

```

And this is how Chrome and weasyprint rendered that html:
- [chrome-generated.pdf](https://github.com/Kozea/WeasyPrint/files/11185733/chrome-generated.pdf)
- [weasyprint-generated.pdf](https://github.com/Kozea/WeasyPrint/files/11185734/weasyprint-generated.pdf)

If you add a `text-align: justify;` to the `p` element, it looks like the line size was correctly calculated, but the initial position was not.

Screenshot 2023-04-09 at 17 25 01

weasyprint 58.1 was run with `weasyprint test.html weasyprint-generated.pdf`. Chrome version is 111.0.5563.146. OS is MacOS 13.2.1 (Ventura) with an M1 chip. 

Let me know how I can help further.

moreaki:
Fixes #1859.
Refs #2486.

## What changed

This keeps the updated inline position and maximum line width calculated while laying out out-of-flow children on root line boxes. For floated `::first-letter` boxes, this means the text that follows the generated first-letter box starts after the float instead of being laid out underneath it.

The existing regression test for #1859 already documented the missing assertion; this PR enables that assertion.

## Root cause

`_out_of_flow_layout` updated `position_x` and `max_x` when it laid out a float, but these values were local to the helper. `split_inline_box` kept using the old values for the following text, so a generated floated `::first-letter` could occupy the left side of the line while the remaining text still started at the original line position.

The returned state is only applied for root `LineBox` layout. Nested inline float accounting keeps the previous behavior, which avoids changing existing inline-float wrapping cases.

## Scope

This does not implement the `initial-letter` property requested in #2486. WeasyPrint still treats declarations such as `initial-letter: 3` as unknown. This PR fixes the smaller drop-cap workaround that #2486 points to through #1859: `::first-letter` with `float: left`.

A fuller `initial-letter` implementation would still need parsing for the property values and dedicated size/sink/drop/raise layout behavior.

## Tests

- `venv/bin/python -m pytest tests/layout/test_float.py -q`
- `venv/bin/python -m pytest tests/test_text.py -q`
- `venv/bin/python -m ruff check weasyprint/layout/inline.py tests/layout/test_float.py`

### DIFF
--- a/tests/layout/test_float.py
+++ b/tests/layout/test_float.py
@@ -838,5 +838,4 @@ def test_first_letter_float():
     line1, = p.children
     first_letter, text = line1.children
     assert first_letter.position_x == 0
-    # TODO: fix problem described in #1859.
-    # assert text.position_x == 20
+    assert text.position_x == 20
--- a/weasyprint/layout/inline.py
+++ b/weasyprint/layout/inline.py
@@ -640,6 +640,8 @@ def _out_of_flow_layout(context, box, containing_block, index, child,
         page = context.current_page
         context.running_elements[running_name][page].append(child)
 
+    return position_x, max_x
+
 
 def _break_waiting_children(context, box, max_x, bottom_space, initial_skip_stack,
                             absolute_boxes, fixed_boxes, line_placeholders,
@@ -764,14 +766,16 @@ def split_inline_box(context, box, position_x, max_x, bottom_space, skip_stack,
         child.position_y = box.position_y
 
         if not child.is_in_normal_flow():
-            _out_of_flow_layout(
+            new_position_x, new_max_x = _out_of_flow_layout(
                 context, box, containing_block, index, child, children,
                 line_children, waiting_children, waiting_floats,
                 absolute_boxes, fixed_boxes, line_placeholders, float_widths,
                 max_x, position_x, bottom_space)
+            if isinstance(box, boxes.LineBox):
+                position_x, max_x = new_position_x, new_max_x
             if child.is_floated():
                 float_resume_index = index + 1
-                if child not in waiting_floats:
+                if child not in waiting_floats and not isinstance(box, boxes.LineBox):
                     max_x -= child.margin_width()
             continue
 


liZe:
Thanks for the pull request.

Isn’t the problem in the box creation step? As explained in #1859:

> The problem is caused by first_letter_to_box, that creates a floating box in the line box, but it should be in the line box container instead. inline_in_block handles this during the tree build step, but first_letter_to_box is called.

I’d prefer to fix the root of the problem instead of working around the consequence of a bug.

PR #2752, opened 2026-04-22
Add initial-letter support

moreaki:
Draft PR stacked on #2751.

Refs #2486.

This starts support for the CSS Inline Layout `initial-letter` property for `::first-letter`.

Because this branch depends on #2751, the diff includes the floated `::first-letter` positioning fix until that PR lands. The additional initial-letter implementation is commit 08ec75278bb435e83f2b3d17eb7689f3fe25b938.

## What changed

- Parse unprefixed `initial-letter`, including `normal`, numeric sizes, explicit sink values, `drop`, `raise`, fractional sizes, and integer `calc()` sink values.
- Apply `initial-letter` to `::first-letter` by reusing the floated first-letter layout path.
- Size the initial letter from the requested line span using the parent line height and cap-height ratios.
- Add first-line clearance for raised/sunken initials so following text keeps the expected visual gap.
- Add focused tests for validation, default drop behavior, explicit sink values, raised initials, fractional sizes, punctuation, styled first-letter boxes, and short paragraphs.

## Scope and known gaps

This is intentionally a first, reviewable step rather than a full implementation of CSS Inline Layout Level 3.

Known gaps include:

- RTL rendering is not good enough yet; the branch includes visual proof showing this as an explicit gap.
- `initial-letter-align` and `initial-letter-wrap` are not implemented.
- Vertical writing modes, ruby interactions, and text-box trimming are not covered.
- Fragmentation/page/column interactions are not handled specially.
- Interactions with other floats and exclusion geometry may need more work for complex documents.
- The prefixed `-webkit-initial-letter` alias is intentionally not added here; this PR focuses on the unprefixed CSS property.

## Visual proof

The locally generated PNG/PDF proof artifacts were uploaded to GitHub attachments. The contact sheet is cropped to the visible content for easier review.

compact initial-letter visual proof

Individual proof artifacts:

- initial-letter: 3: [PNG](https://github.com/user-attachments/assets/6715e7e0-138a-40ac-a005-a40d26352b48), [PDF](https://github.com/user-attachments/files/26980671/initial-letter-3.pdf)
- initial-letter: 3 2: [PNG](https://github.com/user-attachments/assets/b8419f69-ef64-40f4-8088-f71db0ed4912), [PDF](https://github.com/user-attachments/files/26980672/initial-letter-3-2.pdf)
[...]

PR #2753, opened 2026-04-22
Handle larger initial-letter sink values

moreaki:
Draft PR stacked on #2752, which is itself stacked on #2751.

Refs #2486.

This extends the initial `initial-letter` implementation with support for sink values larger than the initial-letter size, such as the MDN example `initial-letter: 3 4`.

Because this branch depends on #2752, the diff includes the earlier `initial-letter` implementation until that PR lands. The additional change in this branch is commit 39811e171d30be7eb4e440099bd155940ed7fcd1.

## What changed

- Track an internal initial-letter sink gap when the sink value is larger than the visual size.
- Apply that gap as an internal top margin when laying out the generated first-letter float.
- Copy the float before applying the internal adjustment so the original inline child is not mutated across layout retries.
- Keep the exclusion area starting at the first line while the glyph itself is shifted downward, matching the behavior used by browser implementations for `size < sink`.
- Add a regression test for `initial-letter: 3 4`.

## Prior art / source behavior

The implementation follows the same broad behavior as Blink’s `ComputeInitialLetterBoxBlockOffset` / `CreateExclusionSpaceForInitialLetterBox` logic: when `size < sink`, the initial-letter box is positioned lower, but the exclusion space still starts at the original line block offset and extends down to the sink depth.

Relevant sources:

- Blink initial-letter layout: https://chromium.googlesource.com/chromium/src/+/main/third_party/blink/renderer/core/layout/inline/initial_letter_utils.cc
- CSS Inline Layout block positioning: https://drafts.csswg.org/css-inline/#initial-letter-block-position
- MDN sink-value example range: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/initial-letter

## Visual proof

Generated from this branch and uploaded as local PNG/PDF artifacts:

initial-letter-3-4

- [`initial-letter: 3 4` PDF](https://github.com/user-attachments/files/26981175/initial-letter-3-4.pdf)

## Scope and known gaps

This only handles the `sink > size` case for the existing `::first-letter`-based implementation. It does not add support for first inline-level child applicability, `initial-letter-align`, `initial-letter-wrap`, vertical writing modes, RTL improvements, or fragmentation behavior.

## Checks

- `venv/bin/python -m pytest tests/css/test_validation.py tests/layout/test_float.py tests/test_text.py -q`
- `venv/bin/python -m ruff check weasyprint/layout/inline.py weasyprint/layout/float.py tests/layout/test_float.py`

### DIFF
--- a/tests/css/test_validation.py
+++ b/tests/css/test_validation.py
@@ -321,6 +321,40 @@ def test_line_height_invalid(rule):
 
 
 @assert_no_logs
+@pytest.mark.parametrize(('rule', 'value'), [
+    ('normal', 'normal'),
+    ('1', (1, 1)),
+    ('1.5', (1.5, 2)),
+    ('3', (3, 3)),
+    ('3 1', (3, 1)),
+    ('3 2', (3, 2)),
+    ('3 calc(1 + 1)', (3, 2)),
+    ('2.5 3', (2.5, 3)),
+    ('3 drop', (3, 3)),
+    ('drop 3', (3, 3)),
+    ('3 raise', (3, 1)),
+    ('raise 3', (3, 1)),
+])
+def test_initial_letter(rule, value):
+    assert get_value(f'initial-letter: {rule}') == value
+
+
+@pytest.mark.parametrize('rule', [
+    'auto',
+    '0',
+    '-1',
+    '3 0',
+    '3 2.5',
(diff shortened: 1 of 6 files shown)


liZe:
See #2752.

PR #2754, opened 2026-04-22
Fix SVG text chunk shaping

### LINKED ISSUE #2249: Don’t separate combining characters in SVG text
SVG allows glyphes to be positioned separately. Because of this, we have to draw text glyph by glyph.

Unfortunately, that’s not what we do: we draw [codepoint by codepoint](https://github.com/Kozea/WeasyPrint/blob/c762d50fc66b722eea2107ca0bd02c6a6f7c5686/weasyprint/svg/text.py#L66-L68). It obviously gives broken results for combining characters, ligatures, etc.

A dirty workaround would be to draw the whole text instead of separating codepoints when there’s no specified x, y, dx, dy or rotate values for specific glyphes (which is the most common case by far).

SVG is not really clear about the way combining characters have to be handled, but SVG2 gives a detailed explanation: https://svgwg.org/svg2-draft/text.html#TSpanNotes

(Of course, rtl and bidi are not supported yet in this case. Let’s keep that problem for a future issue.)

moreaki:
## What changed

SVG text nodes now keep simple text chunks shaped as a whole instead of drawing them code point by code point. This lets Pango keep combining marks and ligatures together in the common case where there are no per-character `x`, `y`, `dx`, `dy`, `rotate`, `letter-spacing`, or `textLength` adjustments.

The existing per-character path is still used when SVG positioning or spacing attributes require manual placement.

Fixes https://github.com/Kozea/WeasyPrint/issues/2249.

## Why

The old SVG text drawing path split text before shaping, so sequences such as `e` + U+0301 COMBINING ACUTE ACCENT and font ligatures could be separated before Pango got a chance to shape the text run. SVG 2 also describes special handling for cases where characters and glyphs are not one-to-one, including combining marks and ligatures:

https://svgwg.org/svg2-draft/text.html#TSpanNotes

This patch takes the small, conservative path suggested by the issue: shape and draw the whole text when there are no glyph-specific positioning attributes.

## Visual proof

SVG source proof:

Rendered PNG:

## Scope

This does not implement the full SVG 2 mapping between addressable characters and shaped glyph clusters for positioned text. RTL and bidi support for positioned SVG text also remain separate work.

## Tests

```console
venv/bin/python -m ruff check weasyprint/svg/text.py tests/draw/svg/test_text.py
```

### DIFF
--- a/tests/draw/svg/test_text.py
+++ b/tests/draw/svg/test_text.py
@@ -22,6 +22,49 @@ def test_text_fill(assert_pixels):
 
 
 @assert_no_logs
+def test_text_combining_character(assert_same_renderings):
+    assert_same_renderings('''
+      <style>
+        @page { size: 20px 20px }
+        svg { display: block }
+      </style>
+      <svg width="20px" height="20px" xmlns="http://www.w3.org/2000/svg">
+        <text x="0" y="15" font-family="weasyprint" font-size="16" fill="blue">
+          é
+        </text>
+      </svg>
+    ''', '''
+      <style>
+        @page { size: 20px 20px }
+        svg { display: block }
+      </style>
+      <svg width="20px" height="20px" xmlns="http://www.w3.org/2000/svg">
+        <text x="0" y="15" font-family="weasyprint" font-size="16" fill="blue">
+          é
+        </text>
+      </svg>
+    ''')
+
+
+@assert_no_logs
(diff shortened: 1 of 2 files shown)


liZe:
Same comment as [#2756.](https://github.com/Kozea/WeasyPrint/pull/2756#issuecomment-4903168438).

PR #2755, opened 2026-04-22
Support RTL SVG text anchoring

moreaki:
## What changed

SVG text now passes `direction="ltr"`/`direction="rtl"` to the Pango text layout used for SVG text drawing.

`text-anchor` handling also now follows the SVG inline-base direction:

- LTR `start` keeps the current behavior and extends right from `x`.
- RTL `start` treats `x` as the right edge and extends left.
- RTL `end` treats `x` as the left edge and extends right.

This is a small step toward #106 and is stacked on #2754.

## Why

SVG 2 defines the `direction` property as the inline-base direction for SVG text. It determines the start/end points used by `text-anchor`, and it can affect text positioning when bidi properties are involved:

https://svgwg.org/svg2-draft/text.html#DirectionProperty

Before this change, SVG direction attributes were cascaded but ignored when building the text layout style, so SVG text anchoring stayed effectively LTR.

## Visual proof

Source SVG:

Rendered PNG:

The HTML wrapper used to render the proof is:

```html
<!doctype html>
<meta charset="utf-8">
<style>
  @page {
    size: 1180px 560px;
    margin: 0;
  }
  html,
  body {
    margin: 0;
  }
  img {
    display: block;
    width: 1180px;
    height: 560px;
  }
</style>
<img src="svg-rtl-anchor-proof.svg" alt="">
```

## Scope

This does not implement full RTL/bidi support. In particular, it does not solve the SVG 2 mapping of per-character `x`, `y`, `dx`, `dy`, and `rotate` lists onto shaped visual glyph clusters after bidi reordering.

## Tests

```console
venv/bin/python -m ruff check weasyprint/svg/__init__.py weasyprint/svg/text.py tests/draw/svg/test_text.py
```

### DIFF
--- a/tests/draw/svg/test_text.py
+++ b/tests/draw/svg/test_text.py
@@ -295,6 +295,45 @@ def test_text_anchor_end(assert_pixels):
 
 
 @assert_no_logs
+def test_text_direction_rtl_anchor_start(assert_pixels):
+    assert_pixels('''
+        ____________BBBBBB__
+        ____________BBBBBB__
+    ''', '''
+      <style>
+        @page { size: 20px 2px }
+        svg { display: block }
+      </style>
+      <svg width="20px" height="2px" direction="rtl"
+           xmlns="http://www.w3.org/2000/svg">
+        <text x="18" y="1.5" font-family="weasyprint" font-size="2"
+              fill="blue">
+          ABC
+        </text>
+      </svg>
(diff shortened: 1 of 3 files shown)


liZe:
Thanks for the pull request. I cherry-picked it to avoid unrelated #2754.

PR #2756, opened 2026-04-22
Support simple SVG unicode-bidi text

moreaki:
## What changed

SVG text now applies simple `unicode-bidi` values by translating them to the Unicode bidi controls Pango already understands before shaping whole text chunks.

Supported values in this simple path are `embed`, `isolate`, `bidi-override`, `isolate-override`, and `plaintext`.

This is stacked on #2755, and only the last commit is new here.

## Why

CSS Writing Modes defines `unicode-bidi` values as effectively inserting matching Unicode bidi control codes into the text stream before bidi reordering:

https://drafts.csswg.org/css-writing-modes-3/#unicode-bidi

Before this patch, SVG `unicode-bidi="bidi-override"` was cascaded as an attribute but ignored when WeasyPrint built the SVG text layout.

## Visual proof

Source SVG:

Rendered PNG:

The HTML wrapper used to render the proof is:

```html
<!doctype html>
<meta charset="utf-8">
<style>
  @page {
    size: 1160px 620px;
    margin: 0;
  }
  html,
  body {
    margin: 0;
  }
  img {
    display: block;
    width: 1160px;
    height: 620px;
  }
</style>
<img src="svg-unicode-bidi-proof.svg" alt="">
```

## Scope

This deliberately applies only to simple whole-text chunks. It does not implement full SVG 2 mapping of per-character `x`, `y`, `dx`, `dy`, and `rotate` lists onto shaped visual glyph clusters after bidi reordering.

## Tests

```console
venv/bin/python -m ruff check weasyprint/svg/text.py tests/draw/svg/test_text.py
```

### DIFF
--- a/tests/draw/svg/test_text.py
+++ b/tests/draw/svg/test_text.py
@@ -334,6 +334,27 @@ def test_text_direction_rtl_anchor_end(assert_pixels):
 
 
 @assert_no_logs
+def test_text_unicode_bidi_override(assert_pixels):
+    assert_pixels('''
+        BBBBBBBBBB__________
+        BBBBBBBBBB__________
+        BBBBBBBBBB__________
+        BBBBBBBBBB__________
+    ''', '''
+      <style>
+        @page { size: 20px 4px }
+        svg { display: block }
+      </style>
+      <svg width="20px" height="4px" xmlns="http://www.w3.org/2000/svg">
+        <text x="10" y="3" font-family="weasyprint" font-size="4"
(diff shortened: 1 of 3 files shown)


liZe:
Thanks for the pull request.

We want to rewrite the whole text management in WeasyPrint, possibly removing the use of Pango layouts. Until then, let’s keep the code as it is and voluntarily ignore Unicode-related features, that will be one less thing to port. We’ll have solid bases after that to add features like this one.

PR #2759, opened 2026-04-22
Document bundled sRGB ICC profile attribution

### LINKED ISSUE #2758: Document bundled sRGB ICC profile attribution
WeasyPrint bundles `weasyprint/pdf/sRGB2014.icc` for sRGB/PDF-A output intents. The profile is a small redistributable ICC v2 sRGB profile from the International Color Consortium, but the repository license text does not currently point readers to its source and reuse terms.

Proposed improvement:

- document the bundled ICC profile source and copyright/license terms;
- keep the current `sRGB2014.icc` profile for the existing `--srgb` and PDF/A behavior;
- strengthen the API regression test so it checks the generated PDF output intent structure, not only the presence of the string `sRGB`.

References:

- ICC sRGB profiles: https://www.color.org/srgbprofiles.xalter
- ICC profile licensing FAQ: https://color.org/faqs.xalter

moreaki:
Closes #2758.

## What changed

- Add source, copyright, and reuse terms for the bundled `sRGB2014.icc` profile to `LICENSE`.
- Strengthen `test_pdf_srgb` to check the generated PDF output intent dictionary fields instead of only looking for the `sRGB` string.

## Why

The bundled ICC profile is redistributable and appropriate for the existing sRGB/PDF-A output intent behavior, but its attribution and reuse terms should be visible in the repository license text.

## Checks

- `python -m pytest tests/test_api.py::test_pdf_srgb tests/test_api.py::test_pdf_no_srgb -q`
- `python -m ruff check tests/test_api.py`

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -27,3 +27,19 @@ SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
 CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
 OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+Bundled ICC profile
+-------------------
+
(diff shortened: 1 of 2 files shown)


liZe:
From [the ICC](https://registry.color.org/profile-library/#license):

> This profile is made available by the International Color Consortium, and may be copied, distributed, embedded, made, used, and sold without restriction.

PR #2767, opened 2026-05-05
Preserve replaced box height on relayout

taxilian:
## What changed

`resolve_percentages` now preserves an already computed numeric height for block replaced boxes when the containing block height is `auto`.

This prevents a relayout pass from changing a floated replaced box, such as an image, back to `height = 'auto'` after replaced-box layout has computed its used height.

## Why

In documents with floated images and page remakes, a block replaced box can be laid out again after its height has already been resolved from its intrinsic ratio. Resetting that height to the string `auto` can leave the box in float exclusion data with a non-numeric height. Later calls to `margin_height()` then try to add that string to numeric padding or border values and raise a `TypeError`.

## Tests

- `python -m pytest tests/layout/test_image.py::test_images_auto_height_block_replaced_relayout -q`
- `python -m pytest tests/layout/test_image.py tests/layout/test_float.py -q`
- `python -m ruff check weasyprint/layout/percent.py tests/layout/test_image.py`

## Background

I hit this when trying to format a book I use Weasyprint with (hambook.org) for a kindle sized screen. Due to time constraints, the fix, repro case, and PR were all figured out by various AI tools, but as best I can tell it seems reasonable. Could be you know a better fix, but at least the test I think shows the failure; I can upload another repro case if needed.

### DIFF
--- a/tests/layout/test_image.py
+++ b/tests/layout/test_image.py
@@ -3,6 +3,7 @@
 import pytest
 
 from weasyprint.formatting_structure import boxes
+from weasyprint.layout.percent import resolve_percentages
 
 from ..testing_utils import assert_no_logs, capture_logs, render_pages
 
@@ -289,6 +290,25 @@ def test_images_16():
     assert img.element_tag == 'img'
     assert img.width == 300
     assert img.height == 200
+
+
+@assert_no_logs
+def test_images_auto_height_block_replaced_relayout():
+    page, = render_pages('''
+        <style>
+            @page { size: 100px }
+            img { display: block; width: 40px }
+        </style>
+        <body>
+            <img src="pattern.png">
+    ''')
+    html, = page.children
+    body, = html.children
+    img, = body.children
+    assert isinstance(img, boxes.BlockReplacedBox)
+    assert img.height == 40
+    resolve_percentages(img, (100, 'auto'))
+    assert img.margin_height() == 40
 
 
 @assert_no_logs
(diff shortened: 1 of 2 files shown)


grewn0uille:
Hi,

Thanks for your PR. Could you please share a sample you’re sure that it shows the issue?

You think the test shows the failure, so we can’t be sure that the fix is actually fixing the "right" thing. And it’d be better for the project quality to know why we’re adding lines of code :)

taxilian:
I had one, may take me a bit to track down where it was. I was formating a book for a small device screen - 3.75 x 4.25 inches, IIRC. I should be able to locate it.

taxilian:
[kindle.css](https://github.com/user-attachments/files/28156291/kindle.css)
[README.md](https://github.com/user-attachments/files/28156292/README.md)
[repro.html](https://github.com/user-attachments/files/28156293/repro.html)

liZe:
@taxilian Thanks for your example, but it’s huge and I can’t reproduce the bug with it. The test included explicitly calls `resolve_percentages` after the layout, it’s not a good example that shows the bug with a normal rendering.

Could you please provide a smaller sample? I tried to find a way to reproduce with the description in your first comment, but I didn’t find a way so far.

liZe:
Feel free to add a comment if needed.

taxilian:
Sorry for the delay, I've just had a ton going on. I can repro this pretty easily, but it requires files that I can't make public; the example I gave you was one contrived by an ai tool to try to provide a reproduction case. it is of course conceivable that it's been fixed by other changes in the project since when I opened this; next time I'm working on that I'll attempt it again.

liZe:
> but it requires files that I can't make public

You can share them by mail if needed.

> next time I'm working on that I'll attempt it again.

Thanks!

Issue #2526, opened 2025-08-15
Borders add an order of magnitude in file size since 66


Christoph-Wagner:
Tested under windows 11 with Weasyprint 65.1 and 66

66: 33.6 KB
65.1: 3.74 KB

```html
<!DOCTYPE html>
<html>
  <head>
    <style>
      body {
        border: 1px dotted black;
      }
    </style>
  </head>
  <body>
  </body>
</html>
```

liZe:
Hi!

You’re right, these dots takes a lot of place. The change happened because of #2265: dotted lines in CSS must be circles, not squares. There’s an operator to draw a rectangle in PDF, but not a circle, so we have to use Bézier curves instead, and they require a lot of coordinates (24 numbers instead of 4 for a rectangle).

We could try different strategies to avoid that. One is to use stroked lines instead of filled circles, as playing with line cap styles gives the possibility to draw circles. But with rounded corners, different styles for different sides, and border images, it quickly becomes a nightmare.

Prince also creates very large documents (actually sometimes larger than WP) with dotted lines and rounded corners, with Bézier curves just like us, so I suspect there’s no other good solution. But they use stroked lines as a smart shortcut when there’s no rounded corner.

If someone finds an efficient and maintainable solution, I’d be happy to talk about it!

Christoph-Wagner:
Ah, that is is fair. I guess the issue should stay open then? 

I don’t persoonally mind, just accidentally encountered it while investigating the problem in the other ticket, and thought I should report it ;)

liZe:
> I guess the issue should stay open then?

Yes, we can do better. That’s actually a good way to dive into WeasyPrint’s code for anyone who likes geometry. 😄 

> I don’t persoonally mind, just accidentally encountered it while investigating the problem in the other ticket, and thought I should report it ;)

Thanks!

Dino0204:
Hello, I’d like to contribute to this issue. I’ve opened a PR at #2776 !

liZe:
Fixed by #2776.

PR #2776, opened 2026-05-24
Use stroked dashes for uniform dotted and dashed borders

Dino0204:
**Fix** #2526.

For uniform dotted borders with no rounded corners, stroke one line per
side with a round line cap and a zero-length dash pattern instead of
emitting a filled Bézier circle per dot. Each dot becomes a round-cap of
 the stroke, so we replace ~24 numbers per dot with one `moveto` +
`lineto` + dash setup per side.

Any other dotted case (mixed colors, different widths, rounded corners)
falls through to the existing slow path.

## Results

Repro case from the issue (`<body style="border: 1px dotted black">`):

| Metric | Before | After |
|---|---|---|
| PDF size | 34,445 B | 836 B |
| `curve_to` operators in content stream | 2,544 | 0 |

Equivalent dashed repro (`<body style="border: 1px dashed black">`):

| Metric | Before | After |
|---|---|---|
| PDF size | 1,185 B | 829 B |

5-case visual sample (1px / 3px / 5px uniform dotted, `border-radius`, 4
 different colors). BEFORE on the left, AFTER on the right. Cases 4 and
5 take the slow path unchanged.

| Metric | Before | After |
|---|---|---|
| PDF size | 102 K | 53 K |
| Pixel diff vs before (150 DPI) | – | 0.02% (anti-alias noise only) |

Same layout for dashed:

| Metric | Before | After |
|---|---|---|
| PDF size | 6,313 B | 5,590 B |
| Pixel diff vs before (150 DPI) | – | 0.14% (anti-alias noise only) |

## Iteration

The first attempt (27896e49) stroked a single closed rectangle with one
dash period applied across the entire perimeter. That collapsed the
whole border into one `re` + dash, but because the perimeter is rarely
an integer multiple of `2 × width`, dots near the corners landed
slightly off — visible especially in the 5px case below:

visual-compare-v1

The second commit (2e3ef796) strokes each side individually and picks
the dash period per side so dots land exactly on both corners, matching
the per-side spacing of the generic path. Corners snap to dots, and the
pixel diff against the generic path drops from 0.6% to 0.02%:

visual-compare

The third commit (`7544edfc`) extends the same function to handle
dashed borders, per @liZe's suggestion. The dotted/dashed split is
contained to three small knobs (cap, inset, dash array); net delta vs
the dotted-only version is +12 effective lines.

dashed-visual-compare

## Tests

- `pytest -k 'dotted or border'`: 404 passed, 2 xfailed (same as main)
- `pytest -k 'dashed'`: passes, no new failures
- `pytest`: no new failures
- `ruff check`: clean

### DIFF
--- a/weasyprint/draw/border.py
+++ b/weasyprint/draw/border.py
@@ -98,16 +98,16 @@ def draw_border(stream, box):
             draw_rounded_border(stream, box, styles[0], colors[0])
         return
 
-    # Dotted fast path: uniform dotted border with no rounded corners. We
-    # stroke a single rectangle with round caps and a zero-length dash
-    # pattern, which is orders of magnitude smaller in the PDF than the
-    # generic per-dot Bézier clip path.
+    # Dotted/dashed fast path: uniform border with no rounded corners. We
+    # stroke one line per side using PDF dash arrays, which is orders of
+    # magnitude smaller in the PDF than the generic per-segment clip path.
     radii = box.rounded_border_box()[4:]
     no_radii = all(rx == 0 and ry == 0 for rx, ry in radii)
-    if (set(styles) == {'dotted'} and single_color and four_sides
-            and len(set(widths)) == 1 and no_radii):
+    if (set(styles) in ({'dotted'}, {'dashed'}) and single_color
+            and four_sides and len(set(widths)) == 1 and no_radii):
         with stream.artifact():
-            draw_dotted_rect_border(stream, box, widths[0], colors[0])
+            draw_dotted_dashed_rect_border(
+                stream, box, styles[0], widths[0], colors[0])
         return
 
     # We're not smart enough to find a good way to draw the borders, we must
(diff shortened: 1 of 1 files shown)


liZe:
Thanks a lot!

liZe:
@Dino0204 The idea is cool, the pull request is OK, but we’ll wait after 69.0 to merge because the release is already huge and late.

Do you think there’s a way to adapt the function for dashes? If the result is short and readable, it may be interesting to handle them both.

Dino0204:
> @Dino0204 The idea is cool, the pull request is OK, but we’ll wait after 69.0 to merge because the release is already huge and late.
> 
> Do you think there’s a way to adapt the function for dashes? If the result is short and readable, it may be interesting to handle them both.

Thank you for the suggestion. I will look into extending the function to cover dashes as well!

Dino0204:
@liZe Took a shot at this in e626fe9 would appreciate your review on the approach. The PR description has been updated accordingly.

liZe:
Thanks a lot for your code. I just cleaned a couple of little things, let’s merge!

PR #2781, opened 2026-05-26
Log an error on unknown render and write_pdf options

### LINKED ISSUE #2731: write_pdf() silently ignores unknown kwargs (e.g. options={"pdf_variant": ...})
## Summary

When calling `write_pdf()` with an unknown keyword argument — or with valid options wrapped in a dict named `options={}` — WeasyPrint silently accepts it and produces a PDF without the requested features. No warning, no exception.

This matters specifically for `pdf_variant`: a caller intending to produce a PDF/A-2b file can end up with a plain PDF and not notice until a downstream validator (veraPDF etc.) flags it. In my case this silently masked a broken PDF/A pipeline until I ran the output through veraPDF.

## Minimal reproducer

```python
from weasyprint import HTML

# This looks right but is NOT activating PDF/A mode.
HTML(string="<p>test</p>").write_pdf(
    "output.pdf",
    options={"pdf_variant": "pdf/a-2b"},
)
```

Running the output through veraPDF shows a plain PDF, not PDF/A-2b. The `options={...}` wrapper is absorbed by `**options` in `Document.write_pdf` under the literal key `options`, so `options['pdf_variant']` is never read.

Correct usage is of course:

```python
HTML(string="<p>test</p>").write_pdf(
    "output.pdf",
    pdf_variant="pdf/a-2b",
)
```

## Expected behavior

Either:
- `TypeError: write_pdf() got an unexpected keyword argument 'options'` (Python's default behavior for **kwargs functions that validate), or
- a logged warning identifying the unknown key.

## Actual behavior

Silently ignored. No indication anything is wrong — the returned PDF looks fine but doesn't carry the PDF/A markers, OutputIntent, /ID, or /Metadata the caller asked for.

## Suggested fix

In `weasyprint/document.py` inside `Document.write_pdf`, after merging user options into `DEFAULT_OPTIONS`:

```python
unknown = set(options) - set(DEFAULT_OPTIONS)
if unknown:
    raise TypeError(
        f"write_pdf() got unexpected keyword argument(s): {sorted(unknown)}"
    )
```

Three lines, strictly additive. No impact on code that passes only valid option keys.

## Environment

- weasyprint 68.1
- Python 3.12
- Linux (WSL2)

Happy to open a PR if the direction is welcome.

danfitz36:
WeasyPrint uses Python logging that is typically silenced by default, so a typo like `write_pdf(options={"pdf_variant": "pdf/a-2b"})` silently produces a plain PDF when the caller expects PDF/A — only discovered later by a downstream validator like veraPDF. `HTML.render` only logged a `warning` (rarely seen), and `HTML.write_pdf` / `Document.write_pdf` did not validate options at all.

Log an `error` listing each unknown keyword instead. Errors are far more visible than the previous warning, so the silent-failure problem is addressed while keeping the behavior backward compatible for callers that pass extra keywords (per review feedback).

`HTML.write_pdf` forwards only known options to `Document.write_pdf`, so an unknown option already reported by `render()` is not reported a second time.

Fix #2731.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -371,9 +371,11 @@ def test_python_render(assert_pixels_equal, tmp_path):
 
 @assert_no_logs
 def test_unknown_options():
+    # Regression test for #2731.
     with capture_logs() as logs:
         pdf_bytes = FakeHTML(string='test').write_pdf(zoom=2, unknown=True)
     assert len(logs) == 1
+    assert logs[0].startswith('ERROR')
     assert 'unknown' in logs[0]
     assert pdf_bytes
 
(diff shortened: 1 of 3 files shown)


liZe:
Thanks for this pull request !

I agree with the idea, but I think that it would be better to log an error instead of raising an error, mainly for backward compatibility reasons.

danfitz36:
Backward compatibility makes sense, and an error-level log still surfaces the problem.

I've pushed an update:

- `render()` and `Document.write_pdf()` now log an error listing each unknown option instead of raising.
- `HTML.write_pdf()` forwards only known options to `Document.write_pdf()`, so an unknown option already reported by `render()` isn't logged a second time.
- Tests updated to assert the error log rather than a raised exception.

Let me know if you'd prefer different wording for the message.

liZe:
It’s perfect, thank you!

Issue #2782, opened 2026-05-27
Fontconfig 2.18.0 breaks font matching


fabioclawd:
# @font-face url() silently ignored on macOS Apple Silicon — fontconfig 2.18.0 regression

## TL;DR

WeasyPrint 68.1 on macOS Apple Silicon silently ignores every `@font-face` rule with `src: url()` and falls back to system fonts (`PT-Serif`, `Hiragino-Sans`). No warning, no log. Root cause is a scoring regression in **fontconfig 2.18.0** (released ~2026-05-26, shipped immediately by Homebrew) that ranks dynamically-added fonts via `FcConfigAppFontAddFile` below system fonts when resolving generic families (`serif`, `sans-serif`). WeasyPrint correctly decodes and registers the fonts; fontconfig then refuses to return them.

Confirmed by pinning fontconfig 2.17.1: same WeasyPrint binary, same HTML, fonts now embed correctly.

Related but distinct: #2510 (fixed in 67.0), #2335 (fixed), #2677 (closed, similar symptom).

## Environment

```
$ weasyprint --info
System: Darwin
Machine: arm64
Version: Darwin Kernel Version 25.5.0 (macOS 26 Tahoe)
WeasyPrint version: 68.1
Python version: 3.14.5
Pydyf version: 0.12.1
Pango version: 15701   (1.57.1)
```

- Apple Silicon, macOS Tahoe 26
- Python 3.14.5 (Homebrew)
- WeasyPrint installed via `pip3`
- **fontconfig 2.18.0** (Homebrew, the trigger)
- Pango 1.57.1, HarfBuzz 14.2.0, Cairo 1.18.4 (Homebrew)

## Minimal Reproduction

```html
<!DOCTYPE html>
<html><head><meta charset="UTF-8"><style>
  @font-face {
    font-family: "Newsreader";
    src: url("fonts/Newsreader.ttf") format("truetype");
  }
  body { font-family: "Newsreader", serif; font-size: 18pt; }
</style></head>
<body>
<h1>Newsreader test via @font-face</h1>
<p>The quick brown fox jumps over the lazy dog. 0123456789.</p>
</body></html>
```

With `fonts/Newsreader.ttf` next to the HTML (Google Fonts Newsreader, converted from woff2 via fontTools).

```bash
weasyprint test.html out.pdf
pdffonts out.pdf
```

**Expected**: `Newsreader-Regular` embedded as TrueType.

**Actual**:
```
[...]

PR #2787, opened 2026-05-28
Fix /Namepace typo in PDF 2 structure-tree namespace

### LINKED ISSUE #2786: Structure-tree namespace dictionary has /Type /Namepace typo
Introduced in commit c7190293 ("Add namespace to Document tag in PDF 2", #2770) on 2026-05-11; affects 69.0.

[`weasyprint/pdf/tags.py:36`](https://github.com/Kozea/WeasyPrint/blob/main/weasyprint/pdf/tags.py#L36):

```python
if str(pdf_version) >= '2.0':
    namespace = pydyf.Dictionary({
        'Type': '/Namepace',   # should be '/Namespace'
        'NS': pydyf.String('http://iso.org/pdf2/ssn'),
    })
```

This is emitted for every tagged PDF 2.0 document (`pdf_variant='pdf/ua-2'`, PDF/A-4 with tags). Conforming readers and PDF/UA validators (e.g. veraPDF) will either ignore the malformed dictionary or flag it; the `Namespaces` array in the structure root then references an object with an invalid `/Type`.

### Reproducer

```python
import weasyprint
pdf = weasyprint.HTML(string='<body>x').write_pdf(
    pdf_variant='pdf/ua-2', uncompressed_pdf=True)
assert b'/Type /Namespace' in pdf  # fails on 69.0
```

PR follows.

danfitz36:
The structure-tree namespace dictionary for tagged PDF 2.0 documents (PDF/UA-2, PDF/A-4 + tags) was emitting `/Type /Namepace` instead of `/Type /Namespace`. Introduced in c7190293 (#2770).

Fix #2786.

## Test plan
- [x] Added regression test `test_pdf_ua_2_namespace_type` that renders with `pdf_variant='pdf/ua-2'` and asserts `/Type /Namespace` is present and `/Namepace` is not
- [ ] CI green

### DIFF
--- a/tests/test_pdf.py
+++ b/tests/test_pdf.py
@@ -766,7 +766,7 @@ def test_pdf_tags_inline_table():
 def test_pdf_ua_2_namespace_type():
     # The structure-tree namespace dictionary must use /Type /Namespace,
     # not the misspelled /Namepace.
-    pdf = FakeHTML(string='<body>abc').write_pdf(
+    pdf = FakeHTML(string='<html lang="en"><body>abc').write_pdf(
         pdf_variant='pdf/ua-2', uncompressed_pdf=True)
     assert b'/Type /Namespace' in pdf
     assert b'/Namepace' not in pdf
--- a/tests/test_pdf.py
+++ b/tests/test_pdf.py
@@ -760,3 +760,13 @@ def test_pdf_tags_inline_table():
     FakeHTML(string='''
       <html lang="en"><table style="display: inline"><td>abc
     ''').write_pdf(pdf_tags=True)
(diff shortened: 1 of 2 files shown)


liZe:
Thank you! Could you please add a `lang` attribute so that the test can pass?

danfitz36:
Done — added `lang="en"` to the test's root element in 15eea9f7. Confirmed locally that without it the missing-lang error trips `@assert_no_logs`; with it the test passes.

liZe:
The tests fail on Windows, but it’s unrelated. I guess it’s because of #2782.

PR #2788, opened 2026-05-29
Fix minor errors in presentational hints

### LINKED ISSUE #2785: Regression: <ol start> and <li value> presentational hints raise NameError
Two `NameError` regressions in [`weasyprint/css/__init__.py`](https://github.com/Kozea/WeasyPrint/blob/main/weasyprint/css/__init__.py) introduced by commit e158264e ("Use HTML parsers for presentational hints", #2773) on 2026-05-21.

Both the `<ol>` and `<li>` branches reference a local variable `size_attr` that only exists inside the earlier `<font size=…>` branch:

```python
elif element.tag == 'ol':
    # From https://www.w3.org/TR/css-lists-3/#ua-stylesheet.
    if size := element.get('start'):
        value = html.parse_integer(size_attr)   # NameError: should be `size`
        ...
elif element.tag == 'li':
    # From https://www.w3.org/TR/css-lists-3/#ua-stylesheet.
    if value := element.get('value'):
        value = html.parse_integer(size_attr)   # NameError: should be `value`
        ...
```

In a fresh generator iteration `size_attr` is unbound; in a later iteration it still holds whatever `<font size=…>` saw last, producing wrong counter values.

### Reproducer

```python
import weasyprint
weasyprint.HTML(string='<ol start="5"><li>a<li>b</ol>').write_pdf(
    'out.pdf', presentational_hints=True)
```

Expected: list numbered 5, 6.
Actual: `NameError: name 'size_attr' is not defined` (or stale value if a `<font>` preceded the list in the same call).

### Bonus: dead `elif` branch in the same PR

[`weasyprint/css/__init__.py:462`](https://github.com/Kozea/WeasyPrint/blob/main/weasyprint/css/__init__.py#L462):

```python
if element.tag in ('tr', 'td', 'th'):
    # …height, width handled here…
elif element.tag == 'tr':   # unreachable
    if height := element.get('height'):
        value = html.map_to_dimension_property(height)
```

The `elif` is unreachable because `'tr'` is in the outer `if`. The pre-refactor code treated `<tr height>` with `map_to_dimension_property` (allowing zero) but the new merged branch uses `map_to_dimension_property_ignoring_zero` for `<tr>` too, so `<tr height="0">` is now silently ignored — a behavioural change worth restoring if intentional.

WeasyPrint 69.0.

liZe:
Fix #2785.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -35,7 +35,9 @@ Features:
   `#2773 <https://github.com/Kozea/WeasyPrint/pull/2773>`_:
   Use HTML parsers for presentational hints
 * `#2631 <https://github.com/Kozea/WeasyPrint/issues/2631>`_,
-  `#2778 <https://github.com/Kozea/WeasyPrint/pull/2778>`_:
+  `#2778 <https://github.com/Kozea/WeasyPrint/pull/2778>`_,
+  `#2785 <https://github.com/Kozea/WeasyPrint/issues/2785>`_,
+  `#2788 <https://github.com/Kozea/WeasyPrint/pull/2788>`_:
   Allow users to set PDF output intent
 
 Bug fixes:
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -166,6 +166,25 @@ def test_ph_lists_types():
 
 
 @assert_no_logs
+def test_ph_lists_start():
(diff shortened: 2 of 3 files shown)


liZe:
(Tests fail on macOS and Windows because of #2782.)

PR #2793, opened 2026-06-05
Accept pathlib.Path base_url in CSS()

### LINKED ISSUE #2764: CSS() fails if base_url is given as a Path
`HTML.__init__()` converts `base_url` to a string if it was passed as a `Path` object.
`CSS.__init__()` is missing this bit and dies:

```
  File "/path1/bug.py", line 6, in <module>
    CSS(string=style.read_text(), base_url=layout)
  File "/path2/weasyprint/__init__.py", line 298, in __init__
    with result as (file_obj, base_url, protocol_encoding, mime_type):
  File "/usr/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/path2/weasyprint/urls.py", line 226, in select_source
    base_url = ensure_url(base_url)
  File "/path2/weasyprint/urls.py", line 195, in ensure_url
    return string if url_is_absolute(string) else path2url(string)
  File "/path2/weasyprint/urls.py", line 111, in url_is_absolute
    return bool(scheme.match(url))
TypeError: expected string or bytes-like object
```

The API documentation gives the type for `base_path` as "(str or pathlib.Path)".

Fixed by applying this patch to `__init__.py`:

```
--- __init__.py.orig    2026-05-01 21:15:28.576230144 +0200
+++ __init__.py 2026-05-01 21:15:34.764262183 +0200
@@ -290,6 +290,8 @@
         PROGRESS_LOGGER.info(
             'Step 2 - Fetching and parsing CSS - %s',
             filename or url or getattr(file_obj, 'name', 'CSS string'))
+        if isinstance(base_url, Path):
+            base_url = str(base_url)
         if url_fetcher is None:
             url_fetcher = URLFetcher()
         result = select_source(
```

assinscreedFC:
## Summary

Fixes #2764.

`HTML.__init__()` converts a `pathlib.Path` `base_url` to a string before passing it down, but `CSS.__init__()` was missing this step. As a result:

```python
CSS(string=..., base_url=Path(...))
```

crashed with `TypeError: expected string or bytes-like object` in `url_is_absolute()`, even though the documented type for `base_url` is "str or pathlib.Path".

This applies the same `Path` → `str` conversion to `CSS.__init__()` (the fix suggested by the reporter, and confirmed by @grewn0uille in the issue).

## Changes

- `weasyprint/__init__.py`: convert a `Path` `base_url` to `str` in `CSS.__init__()`, mirroring `HTML.__init__()`.
- `tests/test_api.py`: add `test_css_base_url_as_path` regression test.

## Note

`Attachment.__init__()` (same module) forwards `base_url` to `select_source()` without the conversion too, so it likely has the same latent issue. I left it out to keep this PR scoped to the reported crash — happy to include it here or in a follow-up if you'd prefer.

## Verification

- `ruff check weasyprint/__init__.py tests/test_api.py` — clean.
- Both files compile.
- I could not run the full `pytest` suite locally (the native Pango/Cairo stack isn't installed on my Windows machine), so I kept the change minimal and identical to the existing, working `HTML` code path. The added regression test reproduces the issue exactly and should be validated by CI.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -248,6 +248,19 @@ def test_css_parsing():
     _test_resource(CSS, 'latin1-test.css', check_css, encoding='latin1')
 
 
+@assert_no_logs
+def test_css_base_url_as_path():
+    """Regression test for #2764.
+
+    ``CSS()`` must accept a ``pathlib.Path`` ``base_url`` like ``HTML()``
+    does, instead of raising ``TypeError`` in ``url_is_absolute()``.
+    """
+    with chdir(Path(__file__).parent):
+        path = Path('resources') / 'utf8-test.css'
+        # Must not raise.
+        CSS(string=path.read_text('utf-8'), base_url=path)
+
+
 def check_png_pattern(assert_pixels_equal, png_bytes, x2=False, blank=False,
                       rotated=False):
     if blank:
--- a/weasyprint/__init__.py
+++ b/weasyprint/__init__.py
@@ -291,6 +291,8 @@ class CSS:
             filename or url or getattr(file_obj, 'name', 'CSS string'))
         if url_fetcher is None:
             url_fetcher = URLFetcher()
+        if isinstance(base_url, Path):
+            base_url = str(base_url)
         result = select_source(
             guess, filename, url, file_obj, string, base_url=base_url,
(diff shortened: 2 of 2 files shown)


liZe:
Thank you!

PR #2799, opened 2026-06-10
Keep HarfBuzz font faces alive during PDF subsetting

jurriaan:
This fixes a production crash we hit when rendering a document with multiple fonts + svg objects to PDF. The crash was a segfault during font subsetting, after layout had already finished.

`Font` stores an `hb_face_t*` during layout and uses it later when the PDF is finalized. That face was borrowed from Pango, but WeasyPrint kept the borrowed pointer. If the Pango/fontconfig owner was released or reused before subsetting, HarfBuzz could receive an invalid face pointer.

The first commit fixes the ownership issue by keeping the existing PangoFc lookup path, but taking an explicit HarfBuzz reference with `hb_face_reference()` and releasing it with `hb_face_destroy()` through `ffi.gc()`.

The second commit is optional cleanup: it gets the face through Pango’s HarfBuzz font API instead of the PangoFc font-map helper.

The regression test creates a `Font` from an `@font-face` font, releases the original layout/font configuration objects, forces collection, and then subsets the font. Without the first commit, this segfaults for me on CPython 3.13.

### DIFF
--- a/tests/test_fonts.py
+++ b/tests/test_fonts.py
@@ -3,10 +3,9 @@
 import gc
 
 from weasyprint import CSS
-from weasyprint.css import InitialStyle
 from weasyprint.pdf.fonts import Font
-from weasyprint.text.fonts import FontConfiguration, get_pango_font_key
-from weasyprint.text.line_break import create_layout
+from weasyprint.text.ffi import ffi, gobject, pango
+from weasyprint.text.fonts import FontConfiguration
 
 from .testing_utils import BASE_URL, assert_no_logs, render_pages
 
@@ -28,22 +27,23 @@ def test_font_face_harfbuzz_face_lifetime():
     font_config = FontConfiguration()
     CSS(string='''
       @font-face {
-        font-family: repro;
+        font-family: bug;
         src: url(weasyprint.otf);
       }
-    ''', base_url=BASE_URL, font_config=font_config)
-    style = InitialStyle(font_config)
-    style['font_family'] = ('repro',)
-    layout = create_layout(
-        'ab', style, context=None, max_width=None, justification_spacing=0)
-    line, _ = layout.get_first_line()
-    pango_font = line.runs.data.item.analysis.font
-    _, description, font_size = get_pango_font_key(pango_font)
-    font = Font(pango_font, description, font_size)
-
-    del font_config, layout, line, pango_font, style
+    ''',
+    base_url=BASE_URL, font_config=font_config)
+    context = ffi.gc(
+        pango.pango_font_map_create_context(font_config.font_map),
+        gobject.g_object_unref)
+    font_description = ffi.gc(
+        pango.pango_font_description_new(), pango.pango_font_description_free)
+    pango.pango_font_description_set_family(font_description, ffi.new('char[]', b'bug'))
+    pango_font = ffi.gc(
+        pango.pango_font_map_load_font(font_config.font_map, context, font_description),
+        gobject.g_object_unref)
+    font = Font(pango_font, font_description, 10)
+    del context, font_description, pango_font, font_config
     gc.collect()
-
     font.clean({1: 'a', 2: 'b'}, hinting=False)
 
 
--- a/tests/test_fonts.py
+++ b/tests/test_fonts.py
@@ -1,6 +1,14 @@
 """Test the fonts features."""
 
-from .testing_utils import assert_no_logs, render_pages
+import gc
+
+from weasyprint import CSS
+from weasyprint.css import InitialStyle
+from weasyprint.pdf.fonts import Font
+from weasyprint.text.fonts import FontConfiguration, get_pango_font_key
+from weasyprint.text.line_break import create_layout
+
+from .testing_utils import BASE_URL, assert_no_logs, render_pages
 
 
 @assert_no_logs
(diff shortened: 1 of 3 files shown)


liZe:
Thanks for the pull request, and thanks a lot for the debug work!

The fix is really clean, and the test should definitely not crash. I’d like to improve the test, so that we only use public API, but it’s hard to reproduce this way. Any idea? You can share the code you use by mail, if needed.

jurriaan:
@liZe I found it very hard to reproduce as well, I could not get a clear reproducer without writing the test like this, I cannot really share the code, since it seems to be super dependent on the actual contents of the document, it was a document with mixed fonts, being used in nested/embedded svgs as well. 

Just a really specific configuration triggered something as it is dependent on how the values are garbage collected it seemed. I initially tried to write a more high-level reproduction, but it was really hard to do. I couldn't get it to trigger in a test setup, but pretty reliable on production for the specific contents (it just SIGSEGVed).

jurriaan:
@liZe I've found a better reproducer, will try to minimize the HTML contents first. Will let you know when I have something that works

jurriaan:
@liZe I got a reproducer that instead of SEGSEGVing, now returns a different error which we also saw in production, where it tries to use the wrong font, basically swapping two fonts around (a variable one, and a non-variable font) causing a `KeyError: 'fvar'` in weasyprint.

Somehow I only got it working in this weird nested case, but it is easy to reproduce this way, using public weasyprint APIs

```python
#!/usr/bin/env -S uv run --script
# /// script
# requires-python = ">=3.12"
# dependencies = ["weasyprint==69.0"]
# ///
import faulthandler
import functools
import tempfile
import threading
from http.server import SimpleHTTPRequestHandler, ThreadingHTTPServer
from pathlib import Path

from fontTools.ttLib import TTFont, newTable
from fontTools.ttLib.tables._f_v_a_r import Axis
from weasyprint import HTML

ROOT = Path(__file__).resolve().parent
STATIC_FONT = ROOT / "weasyprint.otf"
OUTPUT = ROOT / "repro_public_render_font_swap.pdf"


class QuietHandler(SimpleHTTPRequestHandler):
    def log_message(self, format, *args):
        return


def make_variable_font(source, target):
    font = TTFont(source)
    axis_name_id = font["name"].addName("Weight")

    axis = Axis()
    axis.axisTag = "wght"
    axis.minValue = 100.0
    axis.defaultValue = 400.0
    axis.maxValue = 900.0
    axis.flags = 0
    axis.axisNameID = axis_name_id

    fvar = newTable("fvar")
    fvar.axes = [axis]
    fvar.instances = []
    font["fvar"] = fvar
    font.save(target)


def main() -> None:
    if not STATIC_FONT.exists():
        raise SystemExit(f"missing fixture font next to script: {STATIC_FONT}")

    faulthandler.enable()

    with tempfile.TemporaryDirectory(prefix="weasyprint-font-swap-") as temp_dir:
        temp_dir = Path(temp_dir)
        static_font = temp_dir / "static.otf"
        variable_font = temp_dir / "variable.otf"
        outer_svg = temp_dir / "outer.svg"
        nested_svg = temp_dir / "nested.svg"

        static_font.write_bytes(STATIC_FONT.read_bytes())
        make_variable_font(static_font, variable_font)

        nested_svg.write_text("""\
<svg xmlns="http://www.w3.org/2000/svg">
  <style>
    @font-face {
      font-family: StaticRepro;
      src: url("static.otf") format("opentype");
    }
  </style>
  <text
    x="40"
    y="80"
    style="font-family: StaticRepro; font-size: 44px"
  >foobar</text>
</svg>
""")
        outer_svg.write_text("""\
<svg xmlns="http://www.w3.org/2000/svg">
  <image href="nested.svg" width="700" height="393"/>
</svg>
""")

        handler = functools.partial(QuietHandler, directory=temp_dir)
        server = ThreadingHTTPServer(("127.0.0.1", 0), handler)
        host, port = server.server_address
        server_url = f"http://{host}:{port}"

        html = f"""\
<!doctype html>
<style>
  @font-face {{
    font-family: VariableRepro;
    src: url("{server_url}/variable.otf") format("opentype");
    font-weight: 100 900;
  }}

  body {{
    font-family: VariableRepro, sans-serif;
    font-size: 14px;
    font-weight: 400;
  }}
</style>
<p>foobar</p>
<object data="{server_url}/outer.svg" type="image/svg+xml"></object>
"""

        thread = threading.Thread(target=server.serve_forever, daemon=True)
        thread.start()
        try:
            document = HTML(string=html).render()
            document.write_pdf(target=OUTPUT)
            print(f"OK: wrote {OUTPUT}")
        finally:
            server.shutdown()
            server.server_close()


if __name__ == "__main__":
    main()

```

Running it results in:

```text
  File "./repro_public_render_font_swap.py", line 114, in <module>
    main()
    ~~~~^^
  File "./repro_public_render_font_swap.py", line 106, in main
    document.write_pdf(target=OUTPUT)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File ".../site-packages/weasyprint/document.py", line 341, in write_pdf
    pdf = generate_pdf(self, target, zoom, **options)
  File ".../site-packages/weasyprint/pdf/__init__.py", line 313, in generate_pdf
    pdf_fonts = build_fonts_dictionary(
        pdf, document.fonts, compress, subset, options)
  File ".../site-packages/weasyprint/pdf/fonts.py", line 310, in build_fonts_dictionary
    font.clean(to_unicode, options['hinting'])
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../site-packages/weasyprint/pdf/fonts.py", line 137, in clean
    axes = {axis.axisTag: axis for axis in ttfont['fvar'].axes}
                                           ~~~~~~^^^^^^^^
  File ".../site-packages/fontTools/ttLib/ttFont.py", line 805, in __getitem__
    table = self._readTable(tag)
  File ".../site-packages/fontTools/ttLib/ttFont.py", line 813, in _readTable
    data = self.reader[tag]
           ~~~~~~~~~~~^^^^^
  File ".../site-packages/fontTools/ttLib/sfnt.py", line 112, in __getitem__
    entry = self.tables[Tag(tag)]
            ~~~~~~~~~~~^^^^^^^^^^
KeyError: 'fvar'
```

liZe:
I pushed a version of the test that relies a bit less on internal APIs.

> @liZe I got a reproducer that instead of SEGSEGVing, now returns a different error which we also saw in production, where it tries to use the wrong font, basically swapping two fonts around (a variable one, and a non-variable font) causing a `KeyError: 'fvar'` in weasyprint.

@jurriaan Is that also fixed by the fix included in this PR?

jurriaan:
@liZe yes it should be fixed! I'm running this branch in production currently and it fixes it.

liZe:
> yes it should be fixed! I'm running this branch in production currently and it fixes it.

OK! Then let’s keep the current test that shows the problem quite clearly, and it’s time to merge!

Thanks a lot for your work on this issue.

PR #2803, opened 2026-06-18
Fix inline width after backtracked line breaks

Greunlis:
Recompute split inline box width from retained children instead of using a stale cursor position. This avoids phantom trailing width when later inline content is moved to the next line.

Before:
[test.pdf](https://github.com/user-attachments/files/29092395/test.pdf)
After:
[test.pdf](https://github.com/user-attachments/files/29092364/test.pdf)

Co-authored-by: ChatGPT <noreply@openai.com>

### DIFF
--- a/tests/test_text.py
+++ b/tests/test_text.py
@@ -86,6 +86,54 @@ def test_line_breaking_nbsp():
 
 
 @assert_no_logs
+def test_line_breaking_trailing_inline_width():
+    page, = render_pages('''
+      <style>
+        @page { size: 520px 180px; margin: 20px }
+        body {
+          margin: 0;
+          font-family: weasyprint;
+          font-size: 16px;
+          line-height: 1.25;
+        }
+        div {
+          width: 27ch;
+          text-align: right;
+          border-right: 1px solid black;
+        }
+      </style>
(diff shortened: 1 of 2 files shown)


liZe:
Hi!

Thanks for the pull request.

I don’t like to change the line breaking algorithm, but here we remove a hack and use the same, correct way to set the line height, so… Let’s merge, and hope that we won’t introduce a regression!

> Co-authored-by: ChatGPT

Please 🙏 carefully read AI pull requests in the future, the code duplication is obvious and masks the fact that this code is actually a possible simplification… when we use a brain to think about that. The time contributors don’t spend is time spent by maintainers, and that doesn’t scale well. In case you don’t want to spend this time, an issue is often better than a pull request: maintainers know how to use AI to write code if they want to, and if they prefer to write code they definitely don’t want to spend time fixing ChatGPT’s code.

(And thanks for the short comment, not the useless walls of text AI usually generates!)

Greunlis:
Thanks for the fix!

I saw the duplication, but din't have enough code understanding to responsibly refactor it. Good point, sometimes it might be better to make an issue, I'll think about it the next time.

PR #2807, opened 2026-06-23
Use POSIX paths with fontconfig

rfricz:
Fixes issue #2782 by ensuring that only forward slashes are used as directory separators in fontconfig config files, even on Windows. This is necessary because of fontconfig commit e4218828 – [do not mix up a slash and a backslash in file object on Win32](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/commit/e42188283f0ee1fb23089f16dbf95e0f3bcbacaf).


liZe:
Thank you!

Issue #2264, opened 2024-10-03
clarification: line breaking algorithm


ondras:
Hi,
I am trying to understand WeasyPrint in more detail. As far as I know, there are two families of line-breaking algorithms: fast-and-ugly, typically implemented by web browsers, and expensive-but-nice (e.g. Knuth-Plass) that are used in more serious typography. What kind of line breaking algorithm is implemented in WeasyPrint, please?

liZe:
Hi!

> What kind of line breaking algorithm is implemented in WeasyPrint, please?

We’re a browser, we implement the ugly one. :smile:

But happily for you, things will change, even in browsers: the new [`text-wrap-style`](https://www.w3.org/TR/2024/WD-css-text-4-20240529/#propdef-text-wrap-style) provides new ways to break lines, including a `pretty` value that is not specified but is a great opportunity to implement something close to the Knuth-Plass algorithm.

Interested in turning this issue into a feature request like "Support text-wrap-style"? (Note that it may take *a lot* of work to implement.)

ondras:
> Interested in turning this issue into a feature request like "Support text-wrap-style"? (Note that it may take a lot of work to implement.)

Yeah! Interested a lot -- because I am typesetting a book with WeasyPrint and the line-breaking algorithm is an important part. Unfortunately, I am afraid that this feature, due to its complexity, will not make it in time (the book needs to be published in 1-2 months).

Related to line breaking -- I just submitted https://github.com/Kozea/WeasyPrint/issues/2267

liZe:
> Unfortunately, I am afraid that this feature, due to its complexity, will not make it in time (the book needs to be published in 1-2 months).

Probably not. It may be part of a bigger change, like #1840.

PR #2808, opened 2026-06-24
perf: faster rendering via style sharing + measurement caching (byte-identical output)

jellologic:
## What this does

Four focused changes that cut rendering time on layout-heavy documents while
producing **byte-identical PDF output**:

1. **Share computed styles between identical elements** — when two elements
   resolve to the same cascaded declarations, parent style, root style, pseudo
   type and attributes, they share one `ComputedStyle` instance instead of
   recomputing. This is the bulk of the win.
2. **Share `AnonymousStyle` objects among siblings** — anonymous boxes generated
   under the same parent reuse one style object.
3. **Cache preferred-width text measurement** — the intrinsic-width pass shapes
   text that is shaped again at line layout; this memoizes only the
   measurement-time call (returns no layout, used solely by `preferred.py`).
4. **Fast-path `check_math` for non-function tokens** — skips an allocation on
   the common path.

## Numbers (warm, in-process, this machine)

| Document | Before | After | Speedup |
|---|---|---|---|
| 1200-row table | 1935 ms | 940 ms | **2.06×** |
| 3-column text (40 paras) | 112 ms | 104 ms | 1.08× |
| trivial doc | 17 ms | 17 ms | ~1× |

The gain scales with how much style/measurement work a document does; trivial
documents are unaffected.

## Correctness

- Output is **byte-identical** to `main` (verified by SHA-256 over uncompressed
  PDF for the docs above and the test corpus).
- Full test suite: same pass/fail set as `main` on this machine (the only
  failures are pre-existing, font/environment-related).

## The one thing to review carefully

Sharing `ComputedStyle` introduces an invariant: **no box may mutate a shared
style in place.** Two existing sites did (footnote `float` reset in
`element_to_box`, and the table-wrapper property split in `wrap_table`); both
now `.style.copy()` before mutating (see `formatting_structure/build.py`). This
is the real cost of the optimization — a future in-place style mutation would
silently corrupt shared instances. I'm happy to drop the sharing commits and
keep only the contained text-measurement cache (commit "perf(text): …") if you'd
prefer the lower-risk subset.

## Notes

This is a deliberately slimmed PR: an earlier version also carried ~11 rounds of
micro-optimizations (precomputed constants, hoisted lookups, single-pass loops).
Those added ~6% for a lot of added noise, so they've been dropped in favor of
the four changes that actually earn their footprint. Speed isn't WeasyPrint's
primary goal — these are the changes I'd defend on a value/complexity basis, and
each can be dropped independently.

### DIFF
--- a/weasyprint/css/__init__.py
+++ b/weasyprint/css/__init__.py
@@ -73,6 +73,11 @@ class StyleFor:
         #     values: a PropertyValue-like object
         self._computed_styles = {}
 
+        # Style-sharing cache: elements with the same cascaded declarations,
+        # the same parent computed style and the same attributes get identical
+        # computed styles, so they can share one immutable style object.
+        self._computed_cache = {}
+
         # Set when the first page is created, used for viewport-based units.
         self.initial_page_sizes = {'box': None, 'area': None}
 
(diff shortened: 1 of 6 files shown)


liZe:
Hi!

Before the review, I’d like to say that [speed is not WeasyPrint’s main goal](https://doc.courtbouillon.org/weasyprint/stable/going_further.html#why-python), and that we often prefer simple code to fast code. So, no offense if we don’t keep everything: it doesn’t mean that your code is not good, it just means that we prefer simple code. 😁

I’ll review the different changes soon!

jellologic:
No worries, we're using on our own fork and figured I'd do a PR.

liZe:
Thanks a lot for splitting. I’ll cherry pick some commits and add comments for the other ones.

liZe:
That was a cool pull request to look at, thanks!

I cherry picked 703e84f and fe41fdd, with extra changes in 90406887 to keep the cached anonymous style logic in a separate style class (we should have created this class before).

The two other commits need a few more words. 😁

fe41fdd is a long story. An old goal of WeasyPrint is to remove the dependency to Pango, as the other browsers do. We won’t do this now, but a first step is to break lines in WeasyPrint instead of relying on Pango layouts, which currently are an easy but slow solution. A secret, ultimate goal is to #2264, with fast code for the ugly default line-breaking strategy and slower but very beautiful results with a LaTeX-like algorithm.

Until the moment we rewrite the whole line-breaking algorithm without Pango layouts, the rule is to keep the current code as it is if there’s no big bug. It’s probably one of the dirtiest parts of WeasyPrint, with a lot of hacks to handle corner cases and a lot of bug reports that were a nightmare to understand, so we really don’t want to take the risk to break something for performance improvements. I’m not proud of that, but that’s a way to keep some time for the endless list of bugs and features we have to take care of. 😅

72d643f is also a long story. First of all, it’s a good idea, and I’d really like to get it in WeasyPrint. There’s [a 2017 article](https://hacks.mozilla.org/2017/08/inside-a-super-fast-css-engine-quantum-css-aka-stylo/) that talks about this (in the last chapter) and other performance improvements, such as sharing parts of the style that usually go together.

I like the idea, but I feel that the implementation should be more simple. I tried many strategies in the past, and I was satisfied with none of them, so I may be very selective, sorry. 😄 I have some time during the week-end, I’ll try to play again with the idea, and I’ll ping you if I have something interesting to share!

liZe:
@jellologic I opened #2813 with an idea equivalent as yours, we can continue the discussion over there!

jellologic:
Appreciate you taking the time on this. Glad two landed. Anonymous-style class makes sense.

Fair on the line-breaking — not worth poking that for perf. And #2813 is cleaner than mine. Pulling `element` out of `ComputedStyle` and doing `attr()` at tree-build hits the root instead of working around it. Memory win's a nice bonus. Happy to help test over there.

liZe:
Let’s continue the discussion on #2813.

Issue #2809, opened 2026-06-24
Switch to UCRT64 MSYS2 environment on Windows


rfricz:
MSYS2 [has deprecated](https://www.msys2.org/news/#2026-03-15-deprecating-the-mingw64-environment) the MINGW64 [environment](https://www.msys2.org/docs/environments/) recently. Both Weasyprint's tests and Windows build should eventually switch to the UCRT64 environment.

Users can already use the UCRT64 env by installing `mingw-w64-ucrt-x86_64-pango` and running with `C:\msys64\ucrt64\bin` on `PATH` and `WEASYPRINT_DLL_DIRECTORIES` env var set to `"C:\msys64\ucrt64\bin;C:\Program Files\GTK3-Runtime Win64\bin"`. All tests pass and bundling with PyInstaller works with this setup, so the transition should be smooth. I think I could create a PR for this if you're interested.

liZe:
Definitely! Related to #1241.

liZe:
> All tests pass and bundling with PyInstaller works with this setup, so the transition should be smooth. I think I could create a PR for this if you're interested.

That would be cool, thanks. Please keep 'C:\\msys64\\mingw64\\bin;' in the default value of `'WEASYPRINT_DLL_DIRECTORIES'` (and add the new one) so that it doesn’t break existing installations.

liZe:
Fixed by #2810.

PR #2810, opened 2026-06-25
Switch to MSYS2 UCRT64 environment

rfricz:
Relates to issue #2809. Backward compatibility with the deprecated MINGW64 environment is retained, so it will not break existing installations.


liZe:
Thank you!

PR #2813, opened 2026-06-27
Share computed styles between elements

liZe:
Removing lines of code to get the same documents with better performances is always a cool achievement. 😄

Most of the work has been done to remove the element from the computed style, so that we can share it between elements easily.

I get pretty good results, going from 52s, 1.68GB to 37s, 1.00GB with the large document I use for this kind of tests.

It probably requires more work and more tests, but that’s a start!


liZe:
It’s time to merge before we get conflicts!

PR #2823, opened 2026-07-03
Don't forget children overflowing fixed-height boxes at page breaks

### LINKED ISSUE #2798: WeasyPrint truncates PDF files when HTML carries style height:100% and encounters unlucky layout
I encountered a very odd bug where an unlucky layout resulted in my mkdocs-to-pdf output getting arbitrarily cut off after page 161 of 726. The rest of the pages were entirely missing.

I created a simple reproduction case:

```
import weasyprint
weasyprint.HTML(string='<html style="height:100%"><body><h3 style="padding-top: 914.8px;">Header</h3><hr/>Overflow</body></html>').render().write_pdf("minimal.pdf")
```

Example output: [minimal.pdf](https://github.com/user-attachments/files/28736251/minimal.pdf). Note that the string "Overflow" is mysteriously absent from the output.

Here is my proposed patch:

```
diff --git a/weasyprint/layout/block.py b/weasyprint/layout/block.py
index 0d76f798..baca686e 100644
--- a/weasyprint/layout/block.py
+++ b/weasyprint/layout/block.py
@@ -806,7 +806,8 @@ def block_container_layout(context, box, bottom_space, skip_stack, page_is_empty
                 max_lines)
         elif stop:
             # Stop after the rendering of child.
-            if box.height != 'auto':
+            style_height = box.style.get('height', 'auto')
+            if box.height != 'auto' and style_height != 'auto' and getattr(style_height, 'unit') != '%':
                 box_bottom = box.position_y + box.border_height()
                 bottom_margin = collapse_margin(adjoining_margins)
                 if context.overflows(box_bottom, position_y - bottom_margin):
```

I believe the problem is that `box.height != 'auto'` does not necessarily mean that the user defined a fixed height with the intention of cutting off any content beyond that height. In particular, this is untrue when the height is expressed as a percentage.

Example output with fix applied: [minimal-fixed.pdf](https://github.com/user-attachments/files/28736339/minimal-fixed.pdf)

federicobond:
When a fixed-height box stops its layout with content past its bottom, the remaining children were always forgotten. But when the last child's border or padding cannot be fragmented and pushes the content past the page bottom itself, the stop is a page break: the remaining children must be rendered on the next page instead of being silently discarded.

Only forget overflowing children when the content bottom stays within the current page.

Fixes #2798.

### DIFF
--- a/tests/draw/test_overflow.py
+++ b/tests/draw/test_overflow.py
@@ -116,6 +116,46 @@ def test_overflow_5(assert_pixels):
 
 
 @assert_no_logs
+@pytest.mark.parametrize('edge', (
+    'border: 2px solid blue',
+    'padding: 2px; background: blue',
+))
+def test_overflow_6(edge, assert_pixels):
+    # Regression test for #2798. The second div’s unbreakable border or
+    # padding overflows the page bottom, the following text must not be
+    # forgotten but rendered on the next page.
+    assert_pixels('''
+        RRRRRRRR
+        RRRRRRRR
+        RRRRRRRR
+        RRRRRRRR
+        RRRRRRRR
+        BBBBBBBB
+        BBBBBBBB
+        BBBBBBBB
+
+        BBBB____
+        BBBB____
+        ________
+        ________
+        ________
+        ________
+        ________
+        ________
+    ''', f'''
+      <style>
+        @page {{ size: 8px }}
+        html {{ height: 100% }}
+        body {{ color: blue; font-family: weasyprint; line-height: 1;
+                font-size: 2px }}
+      </style>
+      <div style="height: 5px; background: red"></div>
+      <div style="{edge}"></div>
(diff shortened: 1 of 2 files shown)


liZe:
Hi!

Thanks for the pull request.

> But when the last child's border or padding cannot be fragmented and pushes the content past the page bottom itself, the stop is a page break: the remaining children must be rendered on the next page instead of being silently discarded.

Unfortunately, this is not what we want. If a block has a fixed height, we don’t want to continue the rendering on the next page. We only continue for root blocks, because there’s content left in the main flow. I added a failing test to give an example.

As explained in #2798, the problem is in the way we detect overflows: bottom margins are ignored to detect overflow, and that’s OK, but it should not be the case for padding and borders, that should trigger an overflow. As a consequence, the 3 examples should put the blue div on the second page.

PR #2827, opened 2026-07-07
Support <title> element inside an <svg> element to specify the alt text for a figure.

### LINKED ISSUE #2816: SVG: Unable to specify an alt text for screen readers
Hello,

I have an SVG image embedded in my HTML file. I want to provide an alt text so that screen readers will provide a description of the image when reading the PDF out loud.

I have provided the <title> tag in my SVG structure:
```xml
<svg xmlns:xlink="http://www.w3.org/1999/xlink" width="288pt" height="288pt" viewBox="0 0 288 288" xmlns="http://www.w3.org/2000/svg" version="1.1"><title>Tracé du parcours de la ligne 21.</title>...
```
and as you can see, it's being picked up by the browser:

Image

However, in the PDF file created by Weasyprint, I don't see any title associated with the SVG file. Nothing is being read by the screen reader.

Image

**What is the best way to provide an alt text for svg images?**

Here are the HTML and PDF files:
[SO21-20260106_21001231 (12).html](https://github.com/user-attachments/files/29473972/SO21-20260106_21001231.12.html)
[SO21-20260106_21001231 (26).pdf](https://github.com/user-attachments/files/29473973/SO21-20260106_21001231.26.pdf)

Thanks

davidmurray:
Fixes #2816.

FYI. I feel like the code could benefit from some better inline documentation. For example, a docstring in this function that explains what each of the parameters are would be helpful. It would help new contributors to know exactly what each variable does instead of having to guess.
https://github.com/Kozea/WeasyPrint/blob/1a32b8e447e6156f38985fa7456adacd3bd5c7b2/weasyprint/pdf/tags.py#L124

### DIFF
--- a/weasyprint/pdf/tags.py
+++ b/weasyprint/pdf/tags.py
@@ -113,7 +113,7 @@ def _get_pdf_tag(tag):
         return tag.upper()
     elif tag in ('thead', 'tbody', 'tfoot'):
         return tag[:2].upper() + tag[2:]
-    elif tag == 'img':
+    elif tag == 'img' or tag == '{http://www.w3.org/2000/svg}svg':
         return 'Figure'
     elif tag in ('caption', 'figcaption'):
         return 'Caption'
@@ -196,7 +196,14 @@ def _build_box_tree(box, parent, pdf, page_number, nums, links, tags):
             'O': '/Layout',
             'BBox': pydyf.Array((x1, y1, x2, y2)),
         })
-        if alt := box.element.attrib.get('alt'):
+        # Add alternate text for images/figures.
+        # If it's an SVG element, we look for the <title> element inside.
+        # Otherwise, we revert to the "alt" attribute if its not an SVG.
+        if box.element.tag == '{http://www.w3.org/2000/svg}svg':
+            title_elem = box.element.find('{http://www.w3.org/2000/svg}title')
+            if title_elem is not None and title_elem.text:
+                element['Alt'] = pydyf.String(title_elem.text)
+        elif alt := box.element.attrib.get('alt'):
             element['Alt'] = pydyf.String(alt)
         else:
             source = box.element.attrib.get('src', 'unknown')


weasyprint/pdf/tags.py
```diff
@@ -115,6 +115,8 @@ def _get_pdf_tag(tag):
         return tag[:2].upper() + tag[2:]
     elif tag == 'img':
         return 'Figure'
+    elif tag == "figure":
```
liZe: The tag is `'{http://www.w3.org/2000/svg}svg'`, not `'figure'`. You can merge this block with the `elif` above.

weasyprint/pdf/tags.py
```diff
@@ -115,6 +115,8 @@ def _get_pdf_tag(tag):
...
+        return "Figure"
```
liZe: Please use single quotes for strings. You can [check the code with `ruff`](https://doc.courtbouillon.org/weasyprint/stable/contribute.html#tests).

weasyprint/pdf/tags.py
```diff
@@ -196,7 +198,12 @@ def _build_box_tree(box, parent, pdf, page_number, nums, links, tags):
             'O': '/Layout',
             'BBox': pydyf.Array((x1, y1, x2, y2)),
         })
-        if alt := box.element.attrib.get('alt'):
+        # Add alternate text for images/figures.
+        # If it's an SVG file, We look for the <title> element inside, or revert to the "alt" attribute otherwise.
+        title_elem = box.element.find(".//{http://www.w3.org/2000/svg}title")
+        if element.tag == "figure" and title_elem is not None and title_elem.text:
```
liZe: You should fix the tag here too.

weasyprint/pdf/tags.py
```diff
@@ -196,7 +198,12 @@ def _build_box_tree(box, parent, pdf, page_number, nums, links, tags):
             'O': '/Layout',
             'BBox': pydyf.Array((x1, y1, x2, y2)),
         })
-        if alt := box.element.attrib.get('alt'):
+        # Add alternate text for images/figures.
+        # If it's an SVG file, We look for the <title> element inside, or revert to the "alt" attribute otherwise.
+        title_elem = box.element.find(".//{http://www.w3.org/2000/svg}title")
```
liZe: The title should be searched only for `svg` tags. The condition of the next line can be split into two nested `if` blocks for that.

liZe (changes requested):
Thanks a lot for the pull request! The overall logic is good, here are a few comments.

liZe:
> I feel like the code could benefit from some better inline documentation. For example, a docstring in this function that explains what each of the parameters are would be helpful.

I’ll add a docstring.

davidmurray:
Thanks for the helpful comments @liZe. I pushed a fixed version of the PR.

I just tested with the following HTML file and the titles ("A red circle", "A blue circle") were correctly read by my PDF reader!
```html
<!doctype html>
<html lang="en">
<body>
  <figure>
    <svg width="100" height="100">
      <title>A red circle</title>
      <circle cx="50" cy="50" r="40" fill="red" />
    </svg>
  </figure>
  <svg width="100" height="100"">
    <title>A blue circle</title>
    <circle cx="50" cy="50" r="40" fill="blue" />
  </svg>
</body>
</html>
```

liZe:
Thanks a lot!

takesson:
I would like to ask, does this make the SVG title take precedence over the html 'alt' attribute on the img element? Or does this only apply when the svg element is embedded in the html?

liZe:
> I would like to ask, does this make the SVG title take precedence over the html 'alt' attribute on the img element? Or does this only apply when the svg element is embedded in the html?

Only when it’s embedded.

PR #2828, opened 2026-07-07
Don't let a deferred float inflate its block formatting context

mvanherwijnen:
## What's wrong

When a block that establishes a block formatting context (`display: flow-root`, `overflow` other than `visible`, etc.) contains floats, a float that doesn't fit on the current page can push the **entire** block to the next page

This is very visible in print documents full of figures that use `float` + wrapping text: blocks jump to a fresh page instead of partially filling the current one.

## Minimal example

```html
<style>
  @page { size: 100px; margin: 0 }
  body { margin: 0 }
  div.bfc { display: flow-root }
  p { height: 20px; margin: 0 }
  div.float { float: right; width: 30px; height: 40px }
</style>
<div style="height: 70px"></div>
<div class="bfc">
  <p>a</p>
  <div class="float"></div>
  <p>b</p>
</div>
```
- **Expected:** page 1 = spacer + `a`; page 2 = float + `b`.
- **Actual:** page 1 = spacer only; the whole `.bfc` (both paragraphs *and* the float) is pushed to page 2, wasting ~30px of page 1.

## Cause

`float_layout` appends every laid-out float to `context.excluded_shapes`. When the float overflows and `_out_of_flow_layout` breaks *before* it (deferring it to the next page via `resume_at`), the float is left in `excluded_shapes`.

`finish_block_formatting_context` then grows the BFC box's height to `max(shape bottoms)` including the deferred float, which floated up beside the content that stays on the page. The fragment's height therefore exceeds the page, the caller rejects it, and the whole block is moved to the next page instead of keeping the part that fits.

## Fix

When breaking before an overflowing float, drop it from the current formatting context's excluded shapes so it no longer inflates the fragment that stays on the page:

```python
context.excluded_shapes = [
    shape for shape in context.excluded_shapes
    if shape is not new_child]
```

## Test

Adds `test_float_overflow_splits_formatting_context` in `tests/layout/test_float.py`, reproducing the example above and asserting the block is split (paragraph `a` stays on page 1). It fails before the change and passes after. The full `tests/layout/` suite passes with the change.

### DIFF
--- a/tests/layout/test_float.py
+++ b/tests/layout/test_float.py
@@ -308,17 +308,11 @@ def test_float_overflow_splits_formatting_context():
     # context on the current page, or the whole block is pushed to the next page
     # instead of being split across the page break.
     page_1, page_2 = render_pages('''
-      <style>
-        @page { size: 100px; margin: 0 }
-        body { margin: 0 }
-        div.bfc { display: flow-root }
-        p { height: 20px; margin: 0 }
-        div.float { float: right; width: 30px; height: 40px }
-      </style>
+      <style>@page { size: 100px }</style>
       <div style="height: 70px"></div>
-      <div class="bfc">
-        <p>a</p>
-        <div class="float"></div>
+      <div style="display: flow-root">
+        <p style="height: 20px">a</p>
+        <div style="float: right; width: 30px; height: 40px"></div>
         <p>b</p>
       </div>
     ''')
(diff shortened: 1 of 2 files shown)


liZe:
Hi!

Thanks for the pull request.

The problem is actually larger than what your patch fix. I’ll add a commit on top of yours to fix that!

Issue #2673, opened 2026-01-29
Error when specifying table width as calc expression


rhodon-jargon:
Since version 67, the following minimal HTML returns an error:
```html
<!DOCTYPE html>
<head></head>
<body>
  <table style="width: calc(90% - 6em)"> <!-- < THIS IS THE PROBLEM -->
    <tbody>
      <tr>
        <td>
          (i)
      <tr>
        <td>
          (ii)
    </tbody>
  </table>
</body>
```

On version 68.0 on Windows 11, the following traceback is reported:

```Python traceback
Traceback (most recent call last):
  File "%USERPROFILE%\.pyenv\pyenv-win\versions\3.10.11\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "%USERPROFILE%\.pyenv\pyenv-win\versions\3.10.11\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "%HOME%\venv\Scripts\weasyprint.exe\__main__.py", line 6, in <module>
  File "%HOME%\venv\lib\site-packages\weasyprint\__main__.py", line 232, in main
    html.write_pdf(output, **options)
  File "%HOME%\venv\lib\site-packages\weasyprint\__init__.py", line 266, in write_pdf
    self.render(font_config, counter_style, color_profiles, **options)
  File "%HOME%\venv\lib\site-packages\weasyprint\__init__.py", line 222, in render
    return Document._render(
  File "%HOME%\venv\lib\site-packages\weasyprint\document.py", line 201, in _render
    [Page(page_box) for page_box in page_boxes],
  File "%HOME%\venv\lib\site-packages\weasyprint\document.py", line 201, in <listcomp>
    [Page(page_box) for page_box in page_boxes],
  File "%HOME%\venv\lib\site-packages\weasyprint\layout\__init__.py", line 131, in layout_document
    pages = list(make_all_pages(context, root_box, html, pages))
  File "%HOME%\venv\lib\site-packages\weasyprint\layout\page.py", line 1005, in make_all_pages
    page, resume_at = remake_page(i, context, root_box, html)
  File "%HOME%\venv\lib\site-packages\weasyprint\layout\page.py", line 946, in remake_page
    page, resume_at, next_page = make_page(
  File "%HOME%\venv\lib\site-packages\weasyprint\layout\page.py", line 664, in make_page
    root_box, resume_at, next_page, _, _, _ = block_level_layout(
[...]

PR #2831, opened 2026-07-09
Fix line_height() crash on calc() values

### LINKED ISSUE #2812: line_height() crashes on calc() — missing check_math() guard
Same class of bug as #2673, but in line_height() which was missed by the fix in #2675.

```html
<p style="line-height: calc(1.2em + 2px)">Hello</p>
```
File "weasyprint/css/computed_values.py", line 729, in line_height

```python
elif not value.unit:
         ^^^^^^^^^^
AttributeError: 'FunctionBlock' object has no attribute 'unit'
```

length() has the check_math(value) guard at line 357, line_height() at line 724 doesn't.

Version 69.0 on macOS.

apoorvdarshan:
## Summary

`line-height: calc(...)` crashes when laying out text. This is the same class of bug as #2673 in `length()`, but in `line_height()`, which was missed by the fix in #2675.

```html
<p style="line-height: calc(100% + 2px)">Hello</p>
```

```
File "weasyprint/css/computed_values.py", line 673, in line_height
    elif not value.unit:
             ^^^^^^^^^^
AttributeError: 'FunctionBlock' object has no attribute 'unit'
```

A `calc()` value is a `FunctionBlock`, which has no `.unit` attribute, so `elif not value.unit:` raises `AttributeError`. The sibling functions `length()` and `vertical_align()` already guard with `check_math(value)`; `line_height()` did not.

## Fix

`weasyprint/css/computed_values.py` — add a `check_math(value)` branch to `line_height()`, mirroring `length()` and `vertical_align()`. Because `line_height()` must return a resolved `('NUMBER', …)` / `('PIXELS', …)` value (its result is unpacked in `strut()`), the branch resolves the `calc()` via `resolve_math()` with percentages referring to the font size, matching the existing percentage branch. The resolved value then flows through the normal number / percentage / length handling, so `calc()` line-heights compute identically to their literal equivalents (e.g. `calc(1.2em + 2px)` → 14px at `font-size: 10px`, and `calc(1.5 * 1)` → the number 1.5).

## Tests

Added `test_math_functions_line_height` in `tests/css/test_math.py`, parametrized over calc line-heights containing percentages, font units, lengths, and bare numbers. Each renders a paragraph with text (so `line_height()` is exercised during layout via `strut()`). The test fails on `main` with the `AttributeError` above and passes with the fix.

- `pytest tests/css/` — all pass
- `ruff check` — clean

Disclosure: prepared with AI assistance; reviewed and verified locally.

### DIFF
--- a/tests/css/test_math.py
+++ b/tests/css/test_math.py
@@ -275,6 +275,23 @@ def test_math_functions_percentage_and_font_unit(css_property):
     assert len(math_logs) == len(logs)
 
 
+@assert_no_logs
+@pytest.mark.parametrize('line_height', [
+    'calc(50% + 1em)',
+    'calc(100% + 2px)',
+    'calc(1.2em + 2px)',
+    'calc(1.5 * 1)',
+])
+def test_math_functions_line_height(line_height):
+    # Regression test for #2812.
+    # A calc() line-height is a FunctionBlock with no ``unit`` attribute, so
+    # laying out text (which computes line-height through strut()) must not
+    # raise ``AttributeError: 'FunctionBlock' object has no attribute 'unit'``.
+    render_pages(f'''
+      <p style="line-height: {line_height}">Hello</p>
+    ''')
+
+
 @pytest.mark.parametrize('display', [
     'block', 'inline', 'flex', 'grid',
     'list', 'list-item',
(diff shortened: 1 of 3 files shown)


liZe:
Thanks for the pull request. I did the resolution later, not in the computed value, for consistency purpose.

PR #2835, opened 2026-07-13
Feature/svg font variant numeric

ledvinap:
(AI-generated, human-verified)
 
  ## What changed
 
  - Parse cascaded SVG `font-variant-numeric` values using the existing CSS validation helpers.
  - Pass the validated internal value to the text style used by Pango.
  - Add regression coverage for:
    - `proportional-nums` directly on `<text>`
    - inheritance from the root `<svg>`
    - a child `normal` override
    - standalone invalid values
  - Add a small test font containing deterministic `pnum` advances.
 
  ## Why
 
  The SVG cascade already propagates `font-variant-numeric` to text nodes, but the SVG text renderer creates a new style from `INITIAL_VALUES` and previously copied only selected font properties. As a result, numeric
  variants were dropped before `split_first_line` and never reached Pango.
 
  ## Known limitation
 
  The simplified SVG cascade cannot restore an inherited value after an invalid child declaration overrides it. This pre-existing limitation also affects other inherited SVG properties and requires a generic cascade-validation fix outside this change.


grewn0uille:
Thanks for the PR. But we'll support this feature when we'll have a real CSS support in SVG.

ledvinap:
Ok, no problem. I can (AI) port CSS from HTML if such PR is acceptable, but it will take some time.

Is there some preferred workaround?

grewn0uille:
Thanks for asking before submitting the PR but we prefer to avoid having a new AI PR to review.

We love free and open source software, sharing knowledge with people, that’s why we do open source since many years. But it’s been quite a time since we mainly review AI PR. And it’s exhausting, really.
It’s always "human reviewed" or "human supervised", but in the end, no human had a real look at the problem. It ends up with low quality code looking more like a work-around rather than fixing the source of the issue.

And we end up to spend more time to review the PR than the person to submit it. Even to just close it. Because we have to understand what the PR does, if it does really fix/add what it says it fixes/addes, if it does indeed follow the specification… We almost don’t event choose on what we’re working on now, just what AI comes up with.

We thought about closing or limiting PRs, requesting more quality on issues, and stuff like this. Is it a good idea? Probably not. It’d add a barrier to newcomers wanting to participate to the project, and probably be useless regarding AI.

That’s not in particular for you, maybe you even have the intention to improve the tool for everyone. But we’re tired.

PR #2838, opened 2026-07-14
Keep side-by-side floats on the same page

### LINKED ISSUE #2790: Content "widowed" onto next page using CSS columns
Weasyprint 68.1.  (I actually tried upgrading to current today to see if it helped, no joy.  v65.1 was where I started the day.)

We generate PDF letters in a letterhead format, with a logo and text in a left column, a spacer column, and the body in a right column.  These  generally work fine.  I have a case today where the right column gets "widowed" onto a second page.  I suspect this is because the body of the letter is slightly longer than usual.

I've attached the HTML input, the CSS file, and the logo.  I can privately provide the resulting PDF, but would prefer not to make it public here.

Hoping someone can explain to me how my input HTML isn't viable and how to fix it.  Or, you know, if I've found an actual bug, that'll work too. :)

Thanks!

[publetter_20260601-145346.html](https://github.com/user-attachments/files/28478082/publetter_20260601-145346.html)
[publet.css](https://github.com/user-attachments/files/28477765/publet.css)
Image

Sanjays2402:
Fixes #2790.

An overflowing float was deferred whenever an earlier sibling float existed, even when both remained side by side at the top of an otherwise empty page. Keep that top-aligned float on the current page while preserving page breaks for floats pushed down by earlier content.

The regression test fails by creating two pages before the fix and passes with both floats on one page. The float layout tests pass (36 passed, 1 xfailed); the broader layout suite passes with the pre-existing Ghostscript/font-dependent cases excluded (900 passed, 12 xfailed).

### DIFF
--- a/tests/layout/test_float.py
+++ b/tests/layout/test_float.py
@@ -272,6 +272,27 @@ def test_floats_page_breaks_2():
 
 
 @assert_no_logs
+def test_floats_page_breaks_side_by_side():
+    # Don’t move an overflowing float when it’s next to another float.
+    page, = render_pages('''
+      <style>
+        @page { size: 100px; margin: 10px }
+        div { height: 81px; float: left }
+        .left { width: 20% }
+        .right { width: 80% }
+      </style>
+      <div class="left">left</div>
+      <div class="right">right</div>
+    ''')
+
+    divs = [
+        box for box in page.descendants()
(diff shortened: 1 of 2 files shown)


liZe:
We don’t want to force all floats at the top of their parents to be displayed.

PR #2839, opened 2026-07-14
Fix fixed-height table row distribution

### LINKED ISSUE #1092: Wrong line height of tables with fixed height
See #1026.

Sanjays2402:
Fixed-height tables currently leave surplus height as blank space below their row groups, so rows and cells don’t fill the declared table height. This distributes the surplus to auto-height rows (or all rows when none are automatic), repositions following rows and groups, and stretches cells without changing their vertical alignment.

Regression coverage includes the reported single-row case and mixed auto/fixed rows.

Fixes #1092.

### DIFF
--- a/tests/layout/test_table.py
+++ b/tests/layout/test_table.py
@@ -1946,6 +1946,52 @@ def test_table_row_height_4():
 
 
 @assert_no_logs
+def test_table_row_height_5():
+    # Regression test for #1092.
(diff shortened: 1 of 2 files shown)


liZe:
`extra_by_row` is a dictionary that links rows ids to a single value shared by all the rows. I won’t read further.

PR #2840, opened 2026-07-14
Keep manual hyphens after line breaks

### LINKED ISSUE #2614: Soft hyphenation doesn't render properly, but only at certain font sizes
Hello,

Not sure exactly what's going on, but I've found that sometimes when WeasyPrint breaks a word where a \&shy; character is, the hyphen is not rendered. But this seems to be dependent on font size, for example this example works correctly for me at 12pt but not at 11pt. Not sure if it's actually about the font size or some other quirk about the line length that's indirectly related to font size, or what.

See attached my example:

[test.html](https://github.com/user-attachments/files/23693498/test.html)

which renders for me as:

Image

FWIW, I encounter the same bug when using `hyphenate-character` to change the hyphen that should be rendered to something else.

Sanjays2402:
At 11pt, Pango can return a fitting line ending in the invisible soft hyphen. The no-next-word fast path returned before manual hyphenation appended `hyphenate-character`; letting that case continue fixes the reported rendering.

The regression test reproduces the missing visible hyphen without the source change and passes with it. `tests/test_text.py`: 153 passed, 1 xfailed.

Fixes #2614.

### DIFF
--- a/tests/test_text.py
+++ b/tests/test_text.py
@@ -818,6 +818,22 @@ def test_hyphenate_manual_4():
 
 
 @assert_no_logs
+def test_hyphenate_manual_after_previous_line():
+    # Regression test for #2614.
+    page, = render_pages('''
+        <style>
+          p { width: 55px; font-size: 11pt; hyphens: manual }
+        </style>
+        <p>An extreme&shy;ly long English word</p>
+    ''')
+    html, = page.children
+    body, = html.children
+    paragraph, = body.children
+    lines = paragraph.children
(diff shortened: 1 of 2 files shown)


liZe:
Hi.

> At 11pt, Pango can return a fitting line ending in the invisible soft hyphen.

The problem is not the exact 11pt size, that was just a random value in the bug report to show the problem. The test doesn’t even set a font family, the fix is only there to make the test pass, I’m pretty sure you don’t really know what you’re doing.

Capture d’écran du 2026-07-20 11-35-52

LOL.

PR #2844, opened 2026-07-17
Improve filename detection for attachments

### LINKED ISSUE #1862: Download attribute does not work with pdf attachments
When using the download attribute with an <a> tag the value is not used in the 'name' field for pdf attachments, the linked filename is used instead. If using a data url no name is passed and attachments cannot be opened.

liZe:
Fix #1862.

### DIFF
--- a/tests/test_pdf.py
+++ b/tests/test_pdf.py
@@ -701,6 +701,24 @@ def test_annotations():
     assert b'test.txt' in pdf
 
 
+@assert_no_logs
+def test_annotations_base64():
+    pdf = FakeHTML(string='''
+      <title>Test document</title>
+      <meta charset="utf-8">
+      <a
+        rel="attachment"
+        href="data:text/plain;base64,c3VwZXI="
+        download>A link that lets you download an attachment</a>
+    ''').write_pdf()
+
+    assert hashlib.md5(b'super').hexdigest().encode() in pdf
+    assert b'/FileAttachment' in pdf
+    assert b'/EmbeddedFiles' not in pdf
+    assert b'c3VwZXI=' not in pdf
+    assert b'attachment.bin' in pdf
+
+
 @pytest.mark.parametrize(('style', 'media', 'bleed', 'trim'), [
     ('bleed: 30pt; size: 10pt',
      [-30, -30, 40, 40],
--- a/tests/test_pdf.py
+++ b/tests/test_pdf.py
@@ -689,14 +689,16 @@ def test_annotations():
       <title>Test document</title>
       <meta charset="utf-8">
       <a
+        download="test.txt"
         rel="attachment"
         href="data:,some data"
-        download>A link that lets you download an attachment</a>
+      >A link that lets you download an attachment</a>
     ''').write_pdf()
 
     assert hashlib.md5(b'some data').hexdigest().encode() in pdf
     assert b'/FileAttachment' in pdf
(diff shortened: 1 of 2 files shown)


wtschueller:
I modified the new test into

```python
import weasyprint
import base64

data = base64.encodebytes(b'Weasyprint!\n').decode("utf-8")

html = f'''
  <title>Test document</title>
  <meta charset="utf-8">
  <a
    rel="attachment"
    href="data:text/plain;base64,{data}"
    download>A link that lets you download an attachment</a>
'''

weasyprint.HTML(string=html).write_pdf('test_data_url_with_line_break.pdf') 
```

Then there is no attachment. I think it is the line break at the end of `data` resulting in the following `html`

```html
  <title>Test document</title>
  <meta charset="utf-8">
  <a
    rel="attachment"
    href="data:text/plain;base64,V2Vhc3lwcmludCEK
"
    download>A link that lets you download an attachment</a>

```

This works with 69.0.

liZe:
> I think it is the line break at the end of `data` resulting in the following `html`

Oh, you’re right, thanks for the report. That’s because of other, unrelated changes in current main branch.

liZe:
@wtschueller I updated the main branch and this pull request to handle these spaces correctly and test this case to avoid regressions. Everything should work now, thanks a lot for the report!

wtschueller:
I think there is still some odd behaviour:

```python
import weasyprint
import base64

data = base64.encodebytes(b'Weasyprint!\n'*5).decode("utf-8")

html = f'''
  <title>Test document</title>
  <meta charset="utf-8">
  <a
    rel="attachment"
    href="data:text/plain;base64,{data}"
    download>A link that lets you download an attachment</a>
'''

weasyprint.HTML(string=html).write_pdf('test_data_url_with_inner_line_break.pdf')

```
produces an inner line break

```html
  <title>Test document</title>
  <meta charset="utf-8">
  <a
    rel="attachment"
    href="data:text/plain;base64,V2Vhc3lwcmludCEKV2Vhc3lwcmludCEKV2Vhc3lwcmludCEKV2Vhc3lwcmludCEKV2Vhc3lwcmlu
dCEK
"
    download>A link that lets you download an attachment</a>
```

and then there is still no attachment for download. If I use the same procedure for `` then everytthing works as expected.

liZe:
I learned so many details thanks to this example, thanks a lot…

In `href` attributes we have [valid URL potentially surrounded by spaces](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#valid-url-potentially-surrounded-by-spaces). Then, let’s remove the leading and trailing spaces and it will work, right?

No.

From [the URL specification](https://url.spec.whatwg.org/#url-parsing):

> 2. If input contains any ASCII tab or newline, invalid-URL-unit validation error.
> 3. Remove all ASCII tab or newline from input.

So, if a URL contains a newline, it should be an error, but we should remove the newline to give this poor URL a chance to be valid.

It works in WeasyPrint for non-href attributes (href puts the link in CSS, that’s a hack that fails for all newlines), but only for base64 URLs, because Python’s function allows newlines in the base64 strings.

I’ll open a new issue to allow URLs to be surrounded by spaces and to ignore ASCII tab and newline characters in them.

liZe:
Merging, as the remaining problems are in #2851, to be solved before next version.

Thanks again for testing and reporting!

PR #2846, opened 2026-07-18
Support leading partial column spans

### LINKED ISSUE #2845: Support leading partial column spans with integer values
### Problem

WeasyPrint currently accepts `none` and `all` for `column-span`, but not the positive integer syntax introduced by the [CSS Multi-column Layout Level 2 editor's draft](https://drafts.csswg.org/css-multicol-2/#column-span).

A useful initial case is a leading element that spans the first *N* logical columns while later columns keep their full height. For example:

```html
<main>
  <header>Heading across two columns</header>
  <p>Content continues in a single automatic column flow...</p>
</main>
```

```css
main {
  columns: 3;
}

header {
  column-span: 2;
}
```

The first two columns should start below the header, while the third column should retain the full block size. Content should continue sequentially through these irregular column areas without assigning content to physical columns in application code.

This is related to, but distinct from, #980, which concerns `column-span: all` on descendants of a multicolumn container.

### Proposed initial scope

Because the Level 2 draft explicitly says that the integer definition is not yet sufficient and leaves several layout questions unresolved, I propose a deliberately narrow first implementation:

* Accept strictly positive integer `column-span` values.
* Treat `1` like `none`.
* Treat a value greater than or equal to the column count like `all`.
* Apply a partial span only to the leading direct child of the multicolumn container.
* Cover the first *N* logical columns, respecting writing direction.
* Start covered columns below the spanner and leave uncovered columns at the original block-start position.
* Continue normal sequential flow through the resulting irregular column areas.
* Leave non-leading partial spans in normal flow for now.
* Do not attempt balancing of partial spans until the intended semantics are clearer.

This scope can be expanded later once the CSS Working Group resolves the open questions or WeasyPrint establishes compatible semantics.

### Implementation and testing

I have a working implementation ready as a pull request. Its generic tests cover parsing, normal flow, right-to-left layout, nonzero gaps, more uncovered columns, full-span equivalence, automatic height, non-leading fallback, pagination, overflow columns, painting, and column rules.

Local verification completed with Python 3.14 and native Windows Ghostscript:

* 964 parser and column-layout tests passed.
* 631 drawing tests passed, with 25 expected failures.
* Ruff passed.
* The broader suite reported 4,269 passes and 41 expected failures. The remaining local anomaly is an existing command-line test helper splitting this checkout's absolute path at spaces, unrelated to the layout change.

The behavior was motivated by a real production-tooling layout requirement and was also exercised against that application during development, but the proposed upstream tests and examples are intentionally generic and contain no project data.

I started this work because the feature is useful to my own tooling, but I would also like to contribute the implementation upstream so other users can benefit and the behavior does not need to remain in a long-lived private fork.

### Pull request

* #2846 implements the proposed leading integer partial-span behavior and closes this issue when merged.

claell:
Closes #2845.

### Summary

* Accept strictly positive integer values for `column-span`.
* Lay out a leading partial spanner across the first requested logical columns.
* Start covered columns below the spanner while uncovered columns retain their full block size.
* Preserve sequential content flow through irregular column areas without assigning content to columns in application code.
* Keep column-rule painting correct when column rows have different block-start positions.
* Document the deliberately limited behavior and the unresolved specification questions.

### Scope

The [CSS Multi-column Layout Level 2 editor's draft](https://drafts.csswg.org/css-multicol-2/#column-span) accepts integer values but explicitly says that their definition is not yet sufficient. This implementation therefore provides a narrow initial model:

* `1` behaves like `none`.
* A value greater than or equal to the column count behaves like `all`.
* Only a leading direct child creates a partial span.
* The span covers the first *N* logical columns, including right-to-left layout.
* Non-leading partial spans remain in normal flow.
* Partial-span balancing is not attempted.

These constraints are documented so the behavior can be extended deliberately as the specification and project direction become clearer.

### Tests

Generic regression tests cover:

* property parsing and invalid values;
* normal and right-to-left layout;
* nonzero column gaps;
* multiple uncovered columns;
* equivalence with full spans;
* automatic container height;
* non-leading fallback;
* pagination and fixed-height overflow columns;
* painting and column rules.

Local results with Python 3.14:

* 964 parser and column-layout tests passed.
* 631 drawing tests passed, with 25 expected failures.
* Ruff passed.
* The broader suite reported 4,269 passes and 41 expected failures. One unrelated local command-line test remains affected by spaces in the checkout's absolute path.

The implementation originated from a real production-tooling layout requirement and was exercised there during development. All code, documentation, and regression examples included in this pull request are generic and contain no project data.

### Authorship and review

The initial implementation and tests were authored with ChatGPT Codex using the GPT-5.6-Sol model at high reasoning effort, under human direction and review. Maintainer review is required, particularly for the intentionally narrow interpretation of currently unresolved partial-span semantics. I am happy to revise the design and implementation based on maintainer feedback.

### DIFF
--- a/docs/api_reference.rst
+++ b/docs/api_reference.rst
@@ -693,8 +693,8 @@ specification, content can be flowed into multiple columns with a gap and a
 rule between them."
 
 Simple multi-column layouts are supported in WeasyPrint. Features such as
-constrained height, spanning columns or column breaks are **not**
-supported. Pagination and overflow are not seriously tested.
+constrained height or column breaks are **not** supported. Pagination and
+overflow are not seriously tested.
 
 The ``column-width`` and ``column-count`` properties, and the ``columns``
 shorthand property are supported.
@@ -707,11 +707,20 @@ The ``break-before``, ``break-after`` and ``break-inside`` properties are
 supported.
 
 The ``column-span`` property is supported for direct children of columns.
+Alongside ``none`` and ``all``, a limited form of the positive integer syntax
+from `CSS Multi-column Layout Module Level 2`_ is supported. A leading direct
+child can span the first N columns in the logical direction. Following content
+sequentially fills the shorter covered columns before using the full-height
+uncovered columns. A value of ``1`` behaves as ``none`` and a value equal to or
+greater than the column count behaves as ``all``. Other placements and column
+balancing for partial spans are not supported because their behavior is still
+undefined in the draft specification.
 
 The ``column-fill`` property is supported, with a column balancing algorithm
 that should be efficient with simple cases.
 
 .. _CSS Multi-column Layout Module: https://www.w3.org/TR/css-multicol-1/
+.. _CSS Multi-column Layout Module Level 2: https://drafts.csswg.org/css-multicol-2/
 
 CSS Fragmentation Module Level 3 / 4
 ++++++++++++++++++++++++++++++++++++
--- a/tests/css/test_validation.py
+++ b/tests/css/test_validation.py
@@ -54,6 +54,25 @@ def test_unstable_prefix():
 
 
 @assert_no_logs
+@pytest.mark.parametrize(('rule', 'value'), [
+    ('none', 'none'),
+    ('all', 'all'),
+    ('1', 1),
+    ('2', 2),
+])
+def test_column_span(rule, value):
+    """Integer spans are preserved for the multicol layout algorithm."""
+    assert get_value(f'column-span: {rule}') == value
+
+
+@assert_no_logs
+@pytest.mark.parametrize('rule', ['0', '-1', '1.5'])
+def test_column_span_invalid(rule):
+    """The Level 2 grammar only accepts strictly positive integers."""
+    assert_invalid(f'column-span: {rule}')
+
+
+@assert_no_logs
 def test_normal_prefix():
     assert_invalid(
         '-weasy-display: block', 'prefix on this attribute is not supported')
--- a/tests/draw/test_column.py
+++ b/tests/draw/test_column.py
@@ -83,6 +83,39 @@ def test_column_rule_span(assert_pixels):
 
 
 @assert_no_logs
+def test_leading_partial_column_span(assert_pixels):
+    """A partial spanner and irregular column rules paint in their own areas."""
+    assert_pixels('''
+        RRRra
+        arara
+        arara
+    ''', '''
+      <style>
+        @page { margin: 0; size: 5px 3px }
+        body { margin: 0; font-size: 0 }
+        div {
+          columns: 3; column-gap: 1px; column-fill: auto;
+          column-rule: 1px solid red;
+          width: 5px; height: 3px;
+        }
+        section {
+          column-span: 2; height: 1px; background: red;
+        }
+        img { display: block; width: 1px; height: 1px }
+      </style>
+      <div>
+        <section></section>
+        <img src=blue.jpg>
+        <img src=blue.jpg>
+        <img src=blue.jpg>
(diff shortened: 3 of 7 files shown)


liZe:
Hi @claell.

Thanks for your contributions.

You opened 2 tickets and 3 pull requests in 2 hours, without asking if they were interesting for the project maintainers, this is too much. Please wait for feedback before opening new issues / pull requests.

claell:
Hi @liZe,

Thanks for your quick response. And luckily, you seem to have locked PR request creation for my account, otherwise, you'd probably have gotten around 5 more today (because only that locking made me notice your comment) for some other features that I had opened issues for in the past (issues that are still open and seemed to welcome community contributions due to priorities of implementation of the project).

One thing I'd like to point out, though: It is only partly true that I did not ask whether things were interesting to maintainers. For two issues with the associated PRs, that is true. For the PR addressing a part of #13, I was under the impression that you explicitly stated general interest; obviously after a review.

With the current speed of agentic coding, sometimes, it is even nicer to directly have a PR ready for immediate discussion of a (at least mostly) working solution (a bit like in SCRUM with its generated artifacts per sprint).

For now, I'll wait then for your future directions and continue implementing features that I need in my fork for now. Maybe, they can land in upstream some day, we'll see.

liZe:
> And luckily, you seem to have locked PR request creation for my account, otherwise, you'd probably have gotten around 5 more today

I limited them to 1, for everybody.

> I was under the impression that you explicitly stated general interest; obviously after a review.

Not for a partial support without blur.

> With the current speed of agentic coding, sometimes, it is even nicer to directly have a PR ready for immediate discussion of a (at least mostly) working solution (a bit like in SCRUM with its generated artifacts per sprint).

So far, for WeasyPrint, the current speed of agentic coding has main increased the number of rejected pull requests and the days we spent reading useless code and talking with robots.

> For now, I'll wait then for your future directions and continue implementing features that I need in my fork for now. Maybe, they can land in upstream some day, we'll see.

Thank you. 🙏

claell:
> Not for a partial support without blur.

I just checked, it was not in #13, but here, it sounded a bit like it (although of course mentioning need for discussion, ...): https://github.com/Kozea/WeasyPrint/pull/149#issuecomment-1254718532

I'll comment on #13 now with my current thoughts and plan, then we have everything in one place for that stream.

> So far, for WeasyPrint, the current speed of agentic coding has main increased the number of rejected pull requests and the days we spent reading useless code and talking with robots.

I can understand that (that is why I explicitly mentioned GPT 5.6 Sol, high reasoning effort to show I at least did not use a random model, but kind of the best what frontier has to offer right now). I will remain at your disposal for human-to-human interaction ;), just note that oftentimes I need a bit AI enhancement to efficiently do stuff, especially for topics that I am by far not an expert in. We can see whether we can find a sensible workflow for such things, or wether it is more effort than benefit for maintainers. I will at least try as best as I can to make this a good experience for everybody.

liZe:

> The [CSS Multi-column Layout Level 2 editor's draft](https://drafts.csswg.org/css-multicol-2/#column-span) accepts integer values but explicitly says that their definition is not yet sufficient.

This reason, plus the fact that this is not supported by any major browser, is usually a red flag.

> These constraints are documented so the behavior can be extended deliberately as the specification and project direction become clearer.

Supporting a feature that’s:
- not well specified, and
- not implemented in major browsers, and
- not asked by many users
is usually a bad idea. It’s better not to support it than to change the behavior in the future, because we want to avoid breaking rendering for supported features. CSS has been built to be backward compatible, so handling unsupported features is much easier for users than handling changing behaviors.

Let’s wait for a solid specification, or multiple and compatible implementations.

PR #2847, opened 2026-07-18
Use native Ghostscript commands on Windows

### LINKED ISSUE #2848: Find native Ghostscript commands in Windows drawing tests
### Problem

WeasyPrint's drawing-test helper invokes Ghostscript as `gs` on every platform. Native Windows Ghostscript installations expose the command-line executables as `gswin64c.exe` or `gswin32c.exe`, so the tests cannot use an otherwise correctly installed Ghostscript available through `PATH`.

For example, with the official 64-bit Windows Ghostscript installation present:

```powershell
Get-Command gswin64c
```

succeeds, while invoking `gs` fails because no command with that name is installed.

### Proposed behavior

On Windows, the drawing-test helper should select the first available command in this order:

* `gswin64c`
* `gswin32c`
* `gs` as a compatibility fallback

Other platforms should continue to use `gs` unchanged.

This keeps the change limited to the test harness and supports both native Windows packages and environments that already provide a generic `gs` executable.

### Verification

With native Ghostscript 10.07.1 on Windows, this change allowed the complete drawing suite to run:

* 631 tests passed.
* 25 tests retained their existing expected-failure status.
* Ruff passed.

### Pull request

* #2847 implements the proposed Windows command discovery and closes this issue when merged.

claell:
Closes #2848.

### Summary

* Discover the native 64-bit or 32-bit Ghostscript command on Windows.
* Keep using `gs` on other platforms.
* Preserve the existing test helper behavior after command discovery.

Official native Windows Ghostscript installations expose `gswin64c.exe` or `gswin32c.exe`, rather than `gs`. As a result, WeasyPrint's drawing tests could not use an otherwise correctly installed native Ghostscript from `PATH`.

The helper now selects `gswin64c`, then `gswin32c`, and falls back to `gs` so existing environments remain compatible.

### Tests

With Ghostscript 10.07.1 installed natively on Windows and available through `PATH`, the complete drawing suite ran successfully:

* 631 tests passed.
* 25 tests had their existing expected-failure status.

Ruff also passed.

### Authorship and review

This change was authored with ChatGPT Codex using the GPT-5.6-Sol model at high reasoning effort, under human direction and review. Maintainer review is required, and I am happy to adjust the discovery order or platform handling to match the project's preferences.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -22,6 +22,13 @@ from . import draw
 
 MAGIC_NUMBER = b'\x89\x50\x4e\x47\x0d\x0a\x1a\x0a'
 
+if os.name == 'nt':
+    # Official Windows installers expose platform-specific command names.
+    GHOSTSCRIPT = (
+        shutil.which('gswin64c') or shutil.which('gswin32c') or 'gs')
+else:
+    GHOSTSCRIPT = 'gs'
+
 
 def document_write_png(document, target=None, resolution=96, antialiasing=1,
                        zoom=4/30, split_images=False):
@@ -29,7 +36,8 @@ def document_write_png(document, target=None, resolution=96, antialiasing=1,
     with NamedTemporaryFile(delete=False) as pdf:
         document.write_pdf(pdf, zoom=zoom)
     command = (
-        'gs', '-q', '-sDEVICE=png16m', f'-dTextAlphaBits={antialiasing}',
+        GHOSTSCRIPT, '-q', '-sDEVICE=png16m',
+        f'-dTextAlphaBits={antialiasing}',
         f'-dGraphicsAlphaBits={antialiasing}', '-dBATCH', '-dNOPAUSE',
         '-dPDFSTOPONERROR', f'-r{resolution / zoom}', '-dUsePDFX3Profile',
         '-sOutputFile=-', pdf.name)


grewn0uille:
Hi,

We prefer to keep using the Ghostscript provided by MSYS as we know it and that it’s up-to-date rather than using the one from Windows. We had several issues in the past with Windows environment that have been fixed thanks to MSYS and by creating a WeasyPrint exe, so let’s avoid adding new issues enterpoints :smile: 

Also, to use WeasyPrint on Windows, we recommand using MSYS, so people has already all the stuff provided by MSYS, no need to add complexity either.

claell:
Alright, got it. I already have MSYS installed (I was actually suggesting Weasyprint to use MSYS for some of the required dependencies on Windows in the first place IIRC), but did not use it to install `gs` on my machine. Will look into that then as a viable solution! Hopefully, it can work in a sensible way, so it is then on my PATH and can just be used.

BTW: It seems to me that the normal Ghostscript from Windows also is maintained and up to date, no? -> https://ghostscript.com/releases/gsdnld.html

liZe:
> It seems to me that the normal Ghostscript from Windows also is maintained and up to date, no?

Oh, we thought that Ghostscript was already provided by Windows, not that you installed it with Ghostscript’s installer. Then it’s even more obvious that we prefer the one packaged by MSYS.

claell:
Ah, got it. At least for me, it does not seem to be provided out of the box (Windows 11), and even winget does not have it in its repository when I checked.

Right now, I am using the MSYS version (just installed it). Potentially does not *fully* run out of the box, yet and needs slight manual addition to PATH, altough that might be solved when actually restarting my shell environment. I'll try to report back and potentially also suggest some slight modification to make out of the box experience even easier, if sensible (to avaoid needing local PATH hints, ...).

claell:
Following up with concrete results:

I installed `mingw-w64-ucrt-x86_64-ghostscript`, which provides `C:\msys64\ucrt64\bin\gs.exe`. Restarting the application did not make `gs` discoverable from native PowerShell because the MSYS package does not add that directory to the Windows `PATH`.

Adding it for the current PowerShell process works:

```powershell
$env:Path += ';C:\msys64\ucrt64\bin'
.\.venv\Scripts\python.exe -m pytest
```

A representative drawing test then passes using the MSYS Ghostscript.

Running the existing native Windows virtual environment from a full UCRT64 login shell made `gs` available automatically, but this mixed environment crashed inside Pango/GLib before Ghostscript was invoked. This appears to be a local native-library interaction rather than a Ghostscript or WeasyPrint defect, so native PowerShell with the session-only `PATH` addition is the working setup here.

Would a documentation-only PR clarifying this Windows/MSYS setup be useful? Alternatively, would you accept a test-only fallback that invokes the default MSYS Ghostscript executable by absolute path, avoiding exposure of the complete UCRT64 directory through `PATH`? I would not open that code PR unless it is wanted.

liZe:
> Running the existing native Windows virtual environment from a full UCRT64 login shell made `gs` available automatically, but this mixed environment crashed inside Pango/GLib before Ghostscript was invoked.

Tests [pass on CI](https://github.com/Kozea/WeasyPrint/actions/runs/29616135104), you can check what’s done in [tests.yml](https://github.com/Kozea/WeasyPrint/blob/main/.github/workflows/tests.yml).

claell:
Thanks, I checked the CI setup and the contributor documentation. The Windows CI configuration makes sense: it uses native Python and makes the MSYS UCRT64 tools available to the test process.

The remaining concern is the wording in `docs/contribute.rst`. It says that Ghostscript must be installed and available on the local path, and links to the official Ghostscript website. On Windows, however, the official installer exposes `gswin64c.exe`, while the tests invoke the literal command `gs`. The recommended MSYS package provides `gs.exe`, but the preference for that package is not currently documented there.

Would you accept a documentation-only PR clarifying that the tests require the `gs` command and that Windows contributors are advised to install `mingw-w64-ucrt-x86_64-ghostscript` through MSYS2? It could also include an optional PowerShell example for adding `C:\msys64\ucrt64\bin` to the current test process's `PATH`. No test or runtime behavior would change.

PR #2849, opened 2026-07-18
Support lossless zero-blur box shadows

claell:
Partially addresses #13.

## Summary

* Parse and compute the CSS `box-shadow` grammar when every blur radius is zero.
* Paint multiple outer and inset shadows in the specified front-to-back order.
* Support offsets, positive and negative spread, explicit colors, `currentColor`, rounded corners, transparent-box knockout, and collapsed-border table rules.
* Keep PDF output lossless and scalable by representing shadows entirely as vector paths.
* Reject nonzero blur radii with an actionable warning instead of silently accepting an effect that cannot be painted faithfully by this implementation.

## Rationale

The previous implementations in #149 and #859 rasterized shadows. That adds image data to PDF output, scales poorly, and was the central reason those approaches were not merged.

Zero-blur shadows do not have that constraint: their sharp perimeters can be represented directly with PDF paths. A maintainer explicitly noted that this subset would be useful and invited further work in the #13 discussion: https://github.com/Kozea/WeasyPrint/pull/149#issuecomment-1254718532.

This pull request deliberately implements only that lossless subset. It does not introduce Pillow, a blur library, image surfaces, or any new dependency. Blurred shadows, `text-shadow`, and `filter: drop-shadow()` remain separate concerns with different painting and output-policy tradeoffs.

## Implementation notes

* Outer shadows use the border box and are knocked out inside that box even when its background is transparent.
* Inset shadows use the inverse of the shifted padding-box perimeter and are clipped to the rounded padding edge.
* Spread-adjusted radii follow the current CSS Backgrounds radius-expansion algorithm, including its coverage correction for rounded shapes.
* Outer shadows are skipped on internal table boxes in the collapsing-border model; inset shadows remain paintable.
* Shadows are marked as PDF artifacts and participate in the existing transform, opacity, clipping, and stacking machinery.

## Verification

* 913 CSS validation tests pass.
* 9 generic shadow drawing tests pass, covering outer and inset painting, transparent knockout, layer order, spread and `currentColor`, rounded geometry, negative inset spread, collapsed tables, and vector-only PDF output.
* The complete suite reports 4,282 passed and 41 expected failures. The sole failure is the pre-existing Windows path-with-spaces behavior in `test_command_line_render`, where the test helper applies `args.split()` to this checkout's absolute path.
* Ruff reports no issues in the changed Python and test files.
* A generic six-case HTML specimen was rendered to PDF with Ghostscript 10.07.1 and visually inspected. No project-specific data or fixtures are included.

The full suite used the Windows Ghostscript discovery from #2847 as a temporary local test-harness change; that unrelated change is not included in this branch.

## AI assistance disclosure

This change was authored with ChatGPT Codex under human direction and review. Maintainer review remains necessary, particularly for CSS conformance, painting integration, and long-term architecture.

### DIFF
--- a/docs/api_reference.rst
+++ b/docs/api_reference.rst
@@ -580,10 +580,11 @@ WeasyPrint also supports the `border images part`_ of this module, including the
 ``border-image-width``, ``border-image-outset`` and ``border-image-repeat``
 properties.
 
-WeasyPrint does **not** support the `box shadow part`_ of this module,
-including the ``box-shadow`` property. This feature has been implemented in a
-`git branch`_ that is not released, as it relies on raster implementation of
-shadows.
+The ``box-shadow`` property is supported when every shadow has a zero blur
+radius. Multiple outer and inner shadows, spread distances, colors, and rounded
+corners are supported and remain vector graphics in PDF output. Declarations
+with a nonzero blur radius are ignored with a warning because PDF has no native
+Gaussian blur operation and raster shadows do not scale losslessly.
 
 .. _CSS Backgrounds and Borders Level 3: https://www.w3.org/TR/css-backgrounds-3/
 .. _border part: https://www.w3.org/TR/css-backgrounds-3/#borders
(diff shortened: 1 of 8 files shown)


liZe:
Thanks for this pull request.

We opened a shorter and more simple pull request, that also handles blur: #2863.

PR #2852, opened 2026-07-22
Handle spaces and newlines in URLs

### LINKED ISSUE #2851: Handle spaces and newlines in URLs
As detailed in #2844, handling URLs in HTML is a bit complex:

- Surrounding HTML whitespaces must be removed.
- HTML newlines and tabs in the URL must be removed.

This may be a good occasion to remove the infamous `-weasy-link` CSS attribute.

Sanjays2402:
Fix #2851.

Normalize HTML URL attributes before resolution, stripping surrounding HTML whitespace and removing embedded ASCII tabs and newlines. The regression covers links and document base URLs.

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -33,6 +33,30 @@ def test_malformed_url_link(url, base_url):
 
 
 @assert_no_logs
+@pytest.mark.parametrize('whitespace', ['\t', '\n', '\r\n'])
+def test_html_url_whitespace(whitespace):
+    """Test HTML whitespace in URLs."""
+    pdf = FakeHTML(string=(
+        f'<a href=" \t\nhttps://weasyprint.org/foo{whitespace}bar\r\n ">'
+        'My Link</a>')).write_pdf()
+
+    assert re.findall(b'/URI \\((.*)\\)', pdf) == [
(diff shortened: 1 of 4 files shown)


wtschueller:
Works for me as before (66.0). No `WARNING (weasyprint): Ignored `-weasy-link:"[bad string]` anymore.

Please excuse my nit-picking.
In the future, I should consider using `base64.b64encode()` instead of `base64.encodebytes()`. I can’t even remember why I originally used the latter. Perhaps because MIME RFC 2045 seemed somehow appropriate and it just worked.
Seen in that light, it’s just a coincidence that I insisted on tolerating line wraps.

In any case, it’s definitely a good idea to make the parser more tolerant here.

Sanjays2402:
No nit-picking at all, glad it's behaving on 66.0.

For what it's worth I don't think `encodebytes()` was the wrong call, tolerating line wraps is the safer reading of RFC 2045 and there's real data out there with wrapped payloads. That's mostly why I made the parser lenient rather than tightening the encoder side.

grewn0uille:
Please don’t open new pull requests on our projects.

See https://github.com/Kozea/WeasyPrint/pull/2840#issuecomment-5020795453.

PR #2853, opened 2026-07-23
Fix cleared float layout after page breaks

jurriaan:
I ran into an `AssertionError` while rendering a document containing a `float: right; clear: right` block split across pages. The error reached `draw_line()` with:

    x1 = 0
    x2 = 4.0
    y1 = nan
    y2 = nan

The added test reduces this to a cleared float following another float split across pages. On unmodified main, the underline reaches the same assertion. Without the underline, the output is 4 × 4 instead of the expected 4 × 12.

Since the second float has `clear: left`, it should appear below the first one. I think WeasyPrint tries to place it before the first float has finished across the pages. This change makes it wait until the first float is complete.

### DIFF
--- a/tests/draw/test_float.py
+++ b/tests/draw/test_float.py
@@ -872,6 +872,44 @@ def test_float_split_clear(assert_pixels):
 
 
 @assert_no_logs
+def test_float_split_clear_float(assert_pixels):
+    assert_pixels('''
+        BBBB
+        BBBB
+        BBBB
+        BBBB
+        BBBB
+        BBBB
+        BBBB
+        BBBB
+        BBBB
+        BBBB
+        rrrr
+        rrrr
+    ''', '''
+        <style>
+            @page {
+                size: 4px;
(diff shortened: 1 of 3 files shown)


liZe:
Hi,

Thanks for the pull request, there’s definitely a problem here.

Instead of treating `inf` as a special value everywhere we call `get_clearance`, we should avoid `inf` and return a very large value instead. That’s mathematically dubious, but practically acceptable.

My physics teacher had a joke about that.

>A group of mathematicians and a group of physicists are asked about the best way to measure the volume of a cow. The mathematicians explain that it’s easy: we should put the cow in water and measure the new level of water. The physicists answer that mathematicians don’t live in the reality: it would kill the cow, they can’t survive in water. The physicists then tell "In fact, we have a much better solution. Given the fact that the cow can be approximated as a parallellepiped…"

So. Given the fact that infinity can be approximated as `2**63-1`…

PR #2854, opened 2026-07-23
Add test for duplicated inline flex item borders

jurriaan:
I ran into an issue where the border of an `inline-flex` element is drawn twice when the element is itself a flex item.

The added test compares otherwise identical `flex` and `inline-flex` items. It currently fails because a pixel that is blue in the flex version is red in the inline-flex version because of the extra border.

This PR only adds a small reproduction because I’m not sure what the correct fix should be.

### DIFF
--- a/tests/draw/test_box.py
+++ b/tests/draw/test_box.py
@@ -154,6 +154,25 @@ def test_borders_box_sizing(assert_pixels):
 
 
 @assert_no_logs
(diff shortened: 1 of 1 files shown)


grewn0uille:
Hi,

Thanks for the report. Could you please rather open an issue with the reproducing sample?

PR #2856, opened 2026-07-23
Paginate floats using their margin box

mvanherwijnen:
Floats whose content fits at the bottom of a page but whose margins overflow can push their entire formatting context to the next page. They are now deferred using their complete margin box, preserving valid breaks before them.

### DIFF
--- a/tests/layout/test_float.py
+++ b/tests/layout/test_float.py
@@ -339,6 +339,49 @@ def test_float_overflow_splits_formatting_context():
 
 
 @assert_no_logs
+def test_float_margin_overflow_splits_formatting_context():
+    page_1, page_2 = render_pages('''
+      <style>
+        @page { size: 100px; margin: 0 }
+        body { margin: 0 }
+        div.bfc { display: flow-root }
+        p { height: 20px; margin: 0 }
+        div.float {
+          float: right;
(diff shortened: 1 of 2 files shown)


liZe:
Hi,

Thanks for the pull request.

Unfortunately, this is not a correct fix. According to [the specification](https://www.w3.org/TR/CSS22/page.html#allowed-page-breaks):

> When an unforced page break occurs here, the used values of the relevant 'margin-top' and 'margin-bottom' properties are set to '0'.

Here’s an example that would break with this pull request:
```html
<style>
  @page { size: 100px; margin: 0 }
  body { margin: 0 }
  div {
    background: red;
    width: 50px;
    height: 30px;
    margin: 11px;
    float: left;
    clear: left;
  }
  article {
    clear: left;
  }
</style>
<div></div>
<div></div>
<div></div>
<article>abc</article>
```

To be honest, there are many cases like this one that are badly handled by WeasyPrint, with or without floats. Fixing them would require a real support of the [fragmentation module](https://drafts.csswg.org/css-break-4/), with many tests, and most small fixes will probably break as many cases as they fix. And before this huge fix, we probably want to fix the text layout (something like #1840, we already work on that).

PR #2861, opened 2026-07-27
Resolve math functions in color values

gaoflow:
`parse_color()` is the only value type in `css/tokens.py` that never runs the `check_math`/`resolve_math` prologue that `get_number`, `get_percentage`, `get_length` and `get_angle` all open with, so every `<color>`-valued property silently rejects `calc()` and friends:

```css
color: rgb(calc(10 + 10), 20, 30);          /* ignored */
background: rgba(10, 20, calc(30), calc(80%));
background-image: linear-gradient(hsl(calc(10 + 10), 20%, 20%), blue);
```

This adds the missing sibling wrapper, `get_color()`, and routes the validation-time call sites (`color`/`*-color` validators, the `border`/`outline`/`column-rule`/`text-decoration` expanders, gradient color stops, and the color computer) through it. It covers `color`, `background-color`, every `border-*`/`outline`/`column-rule`/`text-decoration` color, gradient stops and `light-dark()`. The 2 xfail tests you added in "Add more tests for math functions" now pass and are un-xfailed.

Colors carry their channels as numbers, percentages and angles, and auditing that surface turned up a second, pre-existing bug: `min()`/`max()`/`clamp()` assumed length arguments, so `clamp(1, 2, 3)`, `clamp(1deg, 2deg, 3deg)` and mixed-unit arguments raised an uncaught `AttributeError` and aborted the whole render (independent of colors, e.g. `width: clamp(1, 2, 3)`). The three branches are factored into one `_resolve_math_argument` helper that compares by unit family, so those cases resolve or are rejected cleanly.

`rgb(calc(1em), 0, 0)` and other length-in-color forms stay invalid, since colors take no lengths.

Tested: `tests/css/test_math.py` (un-xfailed + new parametrized coverage for the color functions, every math function in a color, the property family, gradient stops, and the negative cases); full non-`draw` suite green with no new failures.

### DIFF
--- a/tests/css/test_math.py
+++ b/tests/css/test_math.py
@@ -30,6 +30,14 @@ from ..testing_utils import assert_no_logs, capture_logs, render_pages
     'min(100px)',
     'min(100%, 20em, 100px)',
     'calc(min(4, 2) * 50px)',
+    'calc(tan(min(45deg, 60deg)) * 100px)',
+    'calc(tan(max(45deg, 30deg)) * 100px)',
+    'clamp(50px, 100px, 200px)',
+    'clamp(100px, 250px, 50px)',
+    'calc(clamp(1, 2, 3) * 50px)',
+    'calc(clamp(1, 5, 2) * 50px)',
+    'calc(clamp(2, 5, 1) * 50px)',
+    'calc(tan(clamp(30deg, 45deg, 60deg)) * 100px)',
     'calc(sqrt(4) * 50px)',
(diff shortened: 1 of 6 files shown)


liZe:
Hi,

Thanks for the pull request.

Could you please split your commit into 2 different commits, one for the fix in min / max / clamp, and the other for math in colors?

liZe:
See https://github.com/Kozea/tinycss2/pull/74.

PR #2863, opened 2026-07-28
Add support of the box-shadow property

liZe:
Related to #13.

The current code should work correctly for sharp shadows.

Blur is approximated using gradients, which give a pretty good results when the shadow is light, but is not a Gaussian blur as required by the specification. Gaussian blur requires rasterization, and we don’t want that in PDF.

Known limitations with blur are:

- poor PDF rendering, including artifacts because of rounding errors,
- large or dark shadows, where the difference with Gaussian blur is very visible,
- different x/y values of border radii are not correctly approximated by radial gradients,
- with very large border radii, radial gradients are not cropped correctly.

We don’t want to fix these limitations now, and will only fix them in the future if the changes are small and easy to maintain. Blurred shadows are pretty rare in printed documents, and usually very subtle, so they should be rendered quite correctly with the current code.

### DIFF
--- a/docs/api_reference.rst
+++ b/docs/api_reference.rst
@@ -572,26 +572,27 @@ background layers per box), including the ``background``, ``background-color``,
 ``background-position``, ``background-clip``, ``background-origin`` and
 ``background-size`` properties.
 
-WeasyPrint also supports the `rounded corners part`_ of this module, including
+WeasyPrint supports the `rounded corners part`_ of this module, including
 the ``border-radius`` property.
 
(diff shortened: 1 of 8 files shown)


liZe:
There’s also [extra discussion](https://github.com/w3c/csswg-drafts/issues/10171) about the way shadows should be drawn on split elements. The current implementation works as if `box-decoration-break` is set to `clone`, which shouldn’t be the default. We can wait for a clean specification before changing that.

PR #2865, opened 2026-07-28
Document automatic PDF regeneration on source changes

### LINKED ISSUE #1360: Automatically genereate pdf when html/css changes
I was looking for a way to auto generate a pdf file upon saving the html or css files and came up with the following shell script. Maybe it helps some other linux users trying to solve the same problem.

Dependencies
* inotify-tools

**Disclaimer:** I'm not very good at shell scripting, use it at your own risk.

This looks for changes to files within the directory this script is placed in and then re-generates the pdf. It takes the first html file it finds as input.

I'm sure this can be improved upon a lot.

```
#!/bin/sh


WATCH_DIR=$(dirname $(realpath $0))
INPUT=$(fd -e html --max-results 1 . $WATCH_DIR)
if [ "$INPUT" = "" ]
then
    printf "\n    NO SUITABLE (HTML) INPUT DOCUMENT FOUND\n"
    exit -1
fi
OUTPUT=$WATCH_DIR/$(basename -s html $INPUT)pdf

while true
do
    inotifywait -r $WATCH_DIR -e modify
    printf "\n"
    weasyprint $INPUT $OUTPUT
    if [ "$?" -ne 0 ]
    then
        printf "\n    FAILED TO COMPILE DOCUMENT %\n\n\n" $INPUT
    else
        printf "\n    COMPILED DOCUMENT TO %s\n\n\n" $OUTPUT
    fi
done
```

Thanks for the input @grewn0uille @liZe

JJordan0C:
## Summary

- document a cross-platform workflow for regenerating PDFs with `watchexec`
- include an example for watching specific HTML and CSS files
- include an extension-filter example for watching HTML and CSS files recursively

## Motivation

This gives developers a concise workflow for automatically rebuilding a PDF while editing its HTML and CSS sources on Linux, macOS and Windows.

Closes #1360.

## Validation

- built the complete Sphinx documentation with warnings treated as errors
- tested the specific-file and extension-filter modes with Watchexec 2.5.1 on Windows
- verified that HTML and CSS changes rerun the configured command

### DIFF
--- a/docs/common_use_cases.rst
+++ b/docs/common_use_cases.rst
@@ -2,6 +2,49 @@ Common Use Cases
 ================
 
 
+Regenerate PDFs Automatically
+-----------------------------
+
+During development, you can automatically regenerate a PDF whenever an HTML or
+CSS file changes. On Linux, install ``inotify-tools`` and save the following
+script as ``watch.sh``:
+
+.. code-block:: sh
+
+  #!/bin/sh
+
+  input=${1:-document.html}
+  output=${2:-${input%.*}.pdf}
+
+  render() {
+      if ! weasyprint "$input" "$output"; then
+          printf 'PDF generation failed\n' >&2
+      fi
+  }
+
+  render
+  while inotifywait \
+      --quiet \
+      --recursive \
+      --event close_write,create,delete,move \
+      --include '\.(html|css)$' \
+      .; do
+      render
+  done
+
+Make the script executable and give it the paths of the input and output
+files:
+
+.. code-block:: sh
+
+  $ chmod +x watch.sh
+  $ ./watch.sh document.html document.pdf
+
+The script watches the current directory and its subdirectories. Change the
+final ``.`` passed to ``inotifywait`` if the files to watch are stored
+elsewhere.
+
+
 Include in Web Applications
 ---------------------------
 
--- a/docs/common_use_cases.rst
+++ b/docs/common_use_cases.rst
@@ -6,19 +6,20 @@ Regenerate PDFs Automatically
 -----------------------------
 
 During development, you can automatically regenerate a PDF whenever an HTML or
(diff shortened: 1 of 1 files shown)


grewn0uille:
Hi,

Thanks for the pull request.

I think that we can have something simpler with just the main while loop, like:
```sh
while inotifywait -e modify file.html file.css; do                                                                                     
    python -m weasyprint file.html file.pdf
done
```
So people can copy-paste that into their terminal and customizes it according to their needs.

And what can be really useful to fully handle the issue too, it’s to have a solution for macOS and Windows systems!

JJordan0C:
Thanks for the feedback! I agree that a shorter, copy-pasteable example is more useful here. I’ve simplified the Linux example to the main loop and added a single `watchexec` example that works on macOS and Windows. The complete Sphinx build passes with warnings treated as errors. I’ll push the update shortly.

grewn0uille:
If `watchexec` works on all three OS, it could be worth it to just have this one in the documentation then.

JJordan0C:
That makes sense. To keep it even simpler, what do you think about using a single extension filter?

```sh
watchexec -e html,css -- python -m weasyprint document.html document.pdf
```

This would work on Linux, macOS and Windows, watch all HTML and CSS files in the current directory and its subdirectories, and avoid listing each source file separately. If this looks good to you, I’ll update the documentation accordingly.

grewn0uille:
We could have both examples:
* one for watching specific files,
* one with the extension filter for people who’d like to watch recursively.

grewn0uille:
Looks good to me. Let’s merge it!

JJordan0C:
Thanks! I’ve updated the documentation to use Watchexec on Linux, macOS and Windows, with both requested examples: one for watching specific files and one using the extension filter for recursive watching. The complete Sphinx build still passes with warnings treated as errors, and both modes were tested with Watchexec 2.5.1 on Windows.

PR #2869, opened 2026-07-31
Give every (glyph, text) pair its own ToUnicode identity (#2841)

### LINKED ISSUE #2841: Corrupted ToUnicode CMap (empty bfchar entries) for real-GSUB Arabic text; also emits an empty [] TJ
**Environment**

- WeasyPrint: 69.0
- Pango: 1.58.0
- HarfBuzz: 14.2.1
- Python: 3.13.7
- OS: macOS 26.5.2 (Build 25F84), Darwin 25.5.0, arm64

**Summary**

Rendering a single Arabic word with a real font that has a GSUB table (Noto Sans Arabic)
produces two defects that don't appear when rendering the same font/CSS with unshaped
(Latin) text:

1. Two of the seven `ToUnicode` `bfchar` entries for the shaped run map to an *empty*
   destination string instead of a Unicode code point. Any tool that extracts text from the
   PDF then falls back to the raw CID value at those positions, injecting literal control
   characters into the extracted text — even though the glyphs render correctly on the page.
   This affects everyone who copies/extracts text from such a PDF, independent of any
   particular viewer.
2. Separately, an empty `[] TJ` operator is emitted right after the run's `Tm`/`Tf`. Per the
   PDF spec this is a harmless no-op (shows nothing, advances by zero) — not a spec
   violation — so this isn't a "must fix", just worth knowing: it seems to serve no purpose
   and could simply not be emitted. One consumer we know of, Adobe Illustrator, does crash on
   documents containing it (see "Notes" below), which is the only reason it caught our
   attention.

**Steps to reproduce**

1. `python3 -m venv .venv && .venv/bin/pip install weasyprint==69.0`
2. Get the font: https://github.com/notofonts/arabic/releases/tag/NotoSansArabic-v2.012
   (static instance, `NotoSansArabic-Regular.ttf`, version 2.012 — has a real GSUB table).
3. Save as `arabic.html`, with the font file in the same directory:

   <!DOCTYPE html>
   <html>
   <head>
   <meta charset="utf-8">
   <style>
   @font-face {
     font-family: "NotoArabicRepro";
     src: url("NotoSansArabic-Regular.ttf");
   }
   body { font-family: "NotoArabicRepro"; font-size: 40px; }
   </style>
   </head>
   <body>
   كتاب
   </body>
   </html>

4. Render: `.venv/bin/python -m weasyprint arabic.html arabic.pdf`
   (macOS + Homebrew: prefix with `DYLD_LIBRARY_PATH=/opt/homebrew/lib`, per WeasyPrint's own
   install docs — unrelated to this bug.)
5. Inspect the font's `/ToUnicode` CMap (e.g. via `pikepdf`): two of seven `bfchar` entries
   have an empty destination:

     <000e> <>
     <0012> <>

6. Inspect the page content stream: a `[] TJ` appears right after the `Tm`/`Tf` for the text
   object.

**Control (proves this is shaping-specific, not general)**

Identical HTML/CSS, only the body text changed to the Latin transliteration "ktab" (same
font-face, same font file):

- All five `bfchar` entries in ToUnicode resolve to a real code point, none empty.
- No `[] TJ` anywhere in the content stream.
- Extracted text (PyMuPDF `page.get_text()`) is clean: `'ktab \n'`.

Whereas the Arabic version extracts as:

    'ك\x12تا\x0eب\n'

— literal `\x12`/`\x0e` control bytes injected exactly where the empty bfchar entries are.

**Observation**

Both glyphs with an empty `bfchar` destination are the same base glyph in different forms:
`<000e>` is glyph `uni066E`, and `<0012>` is glyph `uni066E.medi.wide` — a contextual
(medial) form of that same glyph.

**Expected**

ToUnicode `bfchar` entries for shaped Arabic glyphs resolve to real Unicode code points, so
extracted text matches what's visually rendered.

**Notes on the `[] TJ`**

Isolated via single-variable testing in an internal investigation: two otherwise-identical
documents (same corrupted-CMap Arabic run in both), one with the empty `[] TJ` present and
one with it stripped — Illustrator opens the file without it, and fails with "Out of memory"
on the one with it. That test only varied the `[] TJ`'s presence; whether `[] TJ` alone, in a
document with an otherwise-correct ToUnicode CMap, also triggers this has not been tested.
Stripping empty `[] TJ` operators from the content stream is a pure no-op removal (0.000%
pixel diff before/after in that investigation).

**Related**

- The ToUnicode corruption looks like the same class of bug covered by the existing
  regression test `test_to_unicode_rtl` (tests/test_api.py; originally named `test_cmap_rtl`
  when added by #2404, which fixed #1686), but that test uses `font-family: weasyprint` — the
  built-in test font — which doesn't have a real GSUB table, so it doesn't catch this case
  even though it uses the same "اب" input.
- Confirmed independent of variable- vs static-font-instance — this repro already uses the
  static instance.
- Not affected by the `full_fonts`/`hinting` options (checked in the original investigation,
  not re-verified in this minimal repro).

avinash-mall:
Fixes #2841 — the mapping half; c9f2626 covered the standalone empty strings. *(Force-pushed once: the first version only redistributed existing glyph ids and could not recover letters whose every glyph is shared; this version makes every mapping exact.)*

## The defect

`draw_first_line` maps each glyph from a per-glyph utf-8 slice, first come first served. When GSUB draws one character as several glyphs (Noto Sans Arabic draws a dotted letter as a skeleton plus a *shared* dots glyph) two things go wrong:

- inside a multi-glyph cluster the slice `utf8_positions[i:i+2]` gives the whole character to one glyph and the **empty string** to the others — the empty bfchar entries in the issue title;
- a **shared** glyph keeps whichever character it was first drawn with and carries it into every other cluster that reuses it: `الطاقة` extracts as `الطاةة`, a different real word, silently.

## The change

1. Cluster codepoints are found before drawing; the first shown glyph of each cluster carries them, the other glyphs of the cluster carry none.
2. `Font.resolve_glyph` keeps the map exact: the first text a glyph is drawn with keeps the real glyph id, and any *other* text gets a **substitute glyph id**. `Font._embed_aliases` adds each substitute to the font file before subsetting (or whole embedding) as a **composite glyph pointing at the real glyph** — a component, not a copy, so renderers rasterize it through the glyph it stands for. Instances of one file share the id registry, so a file used at several sizes or in forms stays consistent.
3. A single-glyph cluster never steals a glyph that already reads back as other *text* — U+00A0 keeps reading back as the space it shares a glyph with, synthesized small caps keep reading back as their capital, and none of their rendering changes (that constraint came from the draw tests: renderers can rasterize the same outline differently under another id at tiny sizes). But a glyph that reads back as *nothing* does get a substitute even alone, or bare س would vanish from every document where ش came first.
4. The silent extra glyphs of a cluster keep an **explicitly empty** CMap entry (`<gid> <>`), deliberately. Measured:

| non-carrier glyph | pypdf | pdfminer.six |
|---|---|---|
| absent from CMap | falls back to `chr(glyph_id)` mojibake (`ğ`), and that stray LTR letter then derails its bidi reordering, losing *other* letters | `(cid:N)` inside words |
| mapped to `<>` | reads back as nothing | reads back as nothing |

5. Fonts where a substitute cannot be drawn (CFF, bitmap, color fonts whose tables are looked up by id, a full 16-bit glyph space) keep the first text and degrade exactly as before — never as an invented longer string.

## Measured effect (Noto Sans Arabic 2.013, static)

| text | main | this branch |
|---|---|---|
| `الطاقة`, pypdf | `الطاةة` (a different word) | `الطاقة` exact |
| 48-word wrapped paragraph, words findable, pypdf | 13/48 | **48/48** |
| a second 26-word paragraph, words findable | 8/26 | **26/26** |

Rendering is unchanged: Ghostscript renders of the same paragraph on main and on this branch differ by 10 antialiasing pixels on a 1240×1754 page, letterforms identical; the full draw suite passes.

## On c9f2626

Since it was asked in the issue: with c9f2626 in, no standalone empty string is shown anymore; empty `<>` elements still occur *inside* TJ arrays (`[<>103<0055>0] TJ`) — measured harmless to both extractors above.

## Tests

`test_to_unicode_cluster_glyphs` builds a nine-glyph TTF with fontTools (already a dependency) whose `ccmp` decomposes four characters over a shared dots glyph and a shared skeleton glyph — the Noto shape without shipping a font. On main, three of the four characters never reach the CMap; with this change each has exactly one entry, one of them through a substitute id. Full suite: 4264 passed, 40 xfailed (draw tests included), ruff clean.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -674,6 +674,59 @@ def test_to_unicode_rtl():
 
 
 @assert_no_logs
+def test_to_unicode_cluster_glyphs(tmp_path):
+    # Regression test for #2841.
+    # A font whose GSUB draws one character as two glyphs, shared between
+    # characters — the shape of Noto Sans Arabic, where a dotted letter is a
+    # skeleton glyph plus a dots glyph. 'b' and 'c' share their dots glyph;
+    # 'q' and 'r' share their skeleton glyph. Each character must keep exactly
+    # one ToUnicode entry of its own, whether its private glyph is real
+    # ('b', 'c', 'q') or a substitute id had to be made ('r').
+    from fontTools.feaLib.builder import addOpenTypeFeaturesFromString
+    from fontTools.fontBuilder import FontBuilder
+    from fontTools.pens.ttGlyphPen import TTGlyphPen
+
+    from weasyprint.text.fonts import FontConfiguration
+
+    pen = TTGlyphPen(None)
+    pen.moveTo((50, 0))
+    pen.lineTo((50, 700))
+    pen.lineTo((550, 700))
+    pen.lineTo((550, 0))
+    pen.closePath()
+    box = pen.glyph()
+    builder = FontBuilder(1000, isTTF=True)
+    names = [
+        '.notdef', 'b', 'c', 'q', 'r', 'b.skel', 'c.skel', 'qr.skel', 'dots']
+    builder.setupGlyphOrder(names)
+    builder.setupCharacterMap({ord(letter): letter for letter in 'bcqr'})
+    builder.setupGlyf({name: box for name in names})
+    builder.setupHorizontalMetrics({name: (600, 50) for name in names})
+    builder.setupHorizontalHeader(ascent=800, descent=-200)
+    builder.setupNameTable({'familyName': 'cluster', 'styleName': 'Regular'})
+    builder.setupOS2()
+    builder.setupPost()
+    addOpenTypeFeaturesFromString(builder.font, '''
+        feature ccmp {
+            sub b by b.skel dots;
+            sub c by c.skel dots;
+            sub q by qr.skel dots;
+            sub r by qr.skel dots;
+        } ccmp;''')
+    builder.save(str(tmp_path / 'cluster.ttf'))
+
+    font_config = FontConfiguration()
+    pdf = FakeHTML(string='''
+      <style>
+        @font-face { font-family: cluster; src: url(cluster.ttf) }
+        body { font-family: cluster }
+      </style>bcqr''', base_url=str(tmp_path / 'index.html')).write_pdf(
+        uncompressed_pdf=True, font_config=font_config)
+    for letter in 'bcqr':
+        assert pdf.count(b' <%04x>' % ord(letter)) == 1, letter
+
+
+@assert_no_logs
 def test_no_redirect():
     with http_server() as root_url:
         _run(f'--no-http-redirect {root_url}/gzip -')
--- a/weasyprint/draw/text.py
+++ b/weasyprint/draw/text.py
@@ -175,10 +175,26 @@ def draw_first_line(stream, textbox, text_overflow, block_ellipsis, matrix):
             string = ''
             stream.set_font_size(font.hash, 1 if font.bitmap else font_size)
         string += '<'
+        # The codepoints of each cluster, found before drawing: the first shown
+        # glyph of a cluster carries them all in ToUnicode, the other glyphs of
+        # the cluster carry none — glyphs shared between clusters, like the
+        # dots of Arabic letters, must not carry any one letter’s codepoints.
+        cluster_texts = {}
+        start = 0
+        for i in range(1, number_of_glyphs + 1):
+            if i == number_of_glyphs or clusters[i] != clusters[start]:
+                utf8_slice = slice(*sorted(
+                    (utf8_positions[start], utf8_positions[i])))
+                cluster_texts[start] = (
+                    utf8_text[utf8_slice].decode(), i - start > 1)
+                start = i
+        text, in_cluster = '', False
         for i in range(number_of_glyphs):
             glyph_info = glyphs_info[i]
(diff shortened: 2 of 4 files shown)


avinash-mall:
Tested c9f2626 against its parent 4f370d0, only that commit differing — Noto Sans Arabic 2.013, uncompressed output:

| fixture | `[] TJ` operators at 4f370d0 | with c9f2626 |
|---|---|---|
| one word (`الطاقة`) | 1 | **0** |
| 40-word wrapped paragraph | 1 | **0** |

So c9f2626 does fix the empty TJ elements. The empty `<>` elements *inside* non-empty TJ arrays (`[<>103<0055>0] TJ`) are a separate artifact it leaves unchanged — 7 and 84 occurrences, byte-identical before and after — but those measure as cosmetic: patching them out of a finished PDF changes neither pypdf nor pdfminer.six extraction by a character.

liZe:
The idea is good, but we won’t change the font for that. We have to find a better solution.

kirso:
Is there any better solution proposed?

PR #2878, opened 2026-08-05
Support display: contents;

mmulqueen:
This is an attempt to implement `display: contents;` robustly. See issue #2210. I looked at #2746 and your review comments, but this PR is entirely from scratch.

Thanks for all your work on Weasyprint. Long time user and sponsor, first attempt at contributing to the codebase. I've organised this PR into multiple commits in case you want to cherrypick, but most of it is interdependent. This PR was human-written and human-reviewed, AI was used only in a supporting role.

I've tried to keep the implementation as simple as possible (though sadly, nowhere near as simple as I'd hoped it would be based on your comment). I considered whether we could partially support `display:contents` and document the limitations, but in the end it was hard to actually carve much out. The main issue that I've run into is that Weasyprint makes various assumptions about boxes and about text that would have been completely fine before `display: contents;` existed (I'm pretty sure there were no layout transparent elements before that) - I think it's right to keep those assumptions and so have worked around them.

I've taken a test driven approach, starting with the requirements of the spec and I've also tested some common contents cases for sanity.

Most of the implementation is completely to spec, but there are some places where the spec was ambiguous or where strict compliance would have made the implementation messy:
- If an element has `display: contents;` IDs as anchors for internal hyperlinks are pushed down to a rendered child if possible, because there is no representation of the contents box to target. I did not do this for similar print-centric features like target-counter because of complexity - and I think it's reasonable that the user should change their markup/CSS to avoid the issue altogether. The children of a `display: contents;` element work normally for all features - the quirk is only around elements that have both `display: contents;` and an ID that something else targets.
- `float: footnote;`, `::footnote-call`, `::footnote-marker` - ignore any display: contents; setting because it makes no sense and breaks
- I've made no attempt to support `string-set` and `bookmark-*` on elements with `display: contents;` Again, those will work fine on children of a `display: contents;` element.
- Some elements are evaluated to `display: none;` - most of this is exactly as specified by Appendix B, and some of this is for convenience because of how Weasyprint handles the element (e.g. `svg`) - I can't see this causing a problem for users in the real world.

I believe everything else handled is to spec:
- Children that are unboxed due to display: contents; still inherit from it (e.g. `display: contents; color: blue;` should result in blue text)
- The above is achieved by a box that is used as the parent, but never put into the render tree (we assert this with `_sanity_check`).
- Multiple text fragments that end up with different styles, despite ending up in the same box are supported. Where those text fragments find themselves in a flex/grid container, we shove them into a shared box when computing the flex/grid layout (previously unnecessary because all would have been same style and merged earlier into single text node). This also handles tail text.
- `a[href]` with `display: contents` becomes a text node and I've handled linking that up. 
- `::first-letter` and `::first-line` directly on a `display: contents` element have no effect
- Contents is properly handled and tested on table/thead/tbody/tfoot/tr/td/th/colgroup/col.

Everything I could think of is tested and I have included xfails for the things that I've deliberately left unsupported. In addition to the automated tests committed, I also visually compared a "kitchen sink" example page before and after adding support and against Firefox/Chromium (for non-print features).

Apologies for submitting such a big PR as my first - I thought this was going to be a lot smaller when I started ha ha.

### DIFF
--- a/tests/css/test_counters.py
+++ b/tests/css/test_counters.py
@@ -641,3 +641,43 @@ def test_counter_multiple_extends():
     assert li_6.children[0].children[0].children[0].text == '6. '
     assert li_7.children[0].children[0].children[0].text == '7. '
     assert li_8.children[0].children[0].children[0].text == '8. '
+
+
+@assert_no_logs
+def test_counter_display_contents():
+    page, = render_pages('''
+      <style>
+        section { counter-reset: c }
+        p { counter-increment: c }
+        p::before { content: counter(c) }
+      </style>
+      <section><p>a</p><div style="display: contents"><p>b</p><p>c</p></div>
+      <p>d</p></section>
+    ''')
+    html, = page.children
+    body, = html.children
+    section, = body.children
+    values = [
+        p.children[0].children[0].children[0].text for p in section.children]
+    assert values == ['1', '2', '3', '4']
+
+
+@assert_no_logs
+@pytest.mark.parametrize(('display', 'value'), [
+    ('none', '0'),
+    ('contents', '1'),
+    ('block', '1'),
+])
+def test_counter_increment_display(display, value):
+    page, = render_pages(f'''
+      <style>
+        body {{ counter-reset: c }}
+        x {{ counter-increment: c; display: {display} }}
+        p::before {{ content: counter(c) }}
+      </style>
+      <x></x><p>y</p>
+    ''')
+    html, = page.children
+    body, = html.children
+    p = body.children[-1]
+    assert p.children[0].children[0].children[0].text == value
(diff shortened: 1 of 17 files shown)


liZe:
Hi!

Thanks for this pull request, and for your support! ❤️

> Apologies for submitting such a big PR as my first - I thought this was going to be a lot smaller when I started ha ha.

Yeah, that wasn’t the easiest feature to start!

We’ll take a look after we release 70, which is already big.

mmulqueen:
I'm happy to rebase/refactor to bring this up-to-date with HEAD by the way - as long as you think it's on roughly the right track when you get a moment to glance over it (no rush!).

PR #2879, opened 2026-08-05
Always add nested lists tags after list items tags

### LINKED ISSUE #2874: PDF/UA: nested list is tagged before its parent list item, breaking reading order
A nested list is tagged as a **preceding sibling** of the list item it belongs to, instead of inside that item's `LBody`. The resulting structure tree puts the sub-items *before* their parent item, so assistive technology reads them in the wrong order.

This is not caught by PDF/UA validation — veraPDF 1.30.2 reports no violation for the file below, because reading order is not machine-determinable. It shows up in PAC 2026's "screenreader preview", and it is audible in NVDA.

## Minimal example

```html
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>Test</title>
  </head>
  <body>
    <ul>
      <li>First
        <ul>
          <li>A</li>
          <li>B</li>
        </ul>
      </li>
      <li>Second</li>
    </ul>
  </body>
</html>
```

```python
from weasyprint import HTML

HTML("test.html").write_pdf("test.pdf", pdf_variant="pdf/ua-1")
```

No custom CSS — the default stylesheet is enough to trigger it.

## Actual structure tree

```
Document
  L
    L              <- the nested list, hoisted out of the parent LI
      LI
        Lbl        "◦"
        LBody      "A"
      LI
        Lbl        "◦"
        LBody      "B"
    LI
      Lbl          "•"
      LBody        "First"
    LI
      Lbl          "•"
      LBody        "Second"
```

Reading order: **A, B, First, Second.**

## Expected structure tree

```
Document
  L
    LI
      Lbl          "•"
      LBody
        "First"
        L          <- nested list inside its parent's LBody
          LI
            Lbl    "◦"
            LBody  "A"
          LI
            Lbl    "◦"
            LBody  "B"
    LI
      Lbl          "•"
      LBody        "Second"
```

Reading order: First, A, B, Second — matching the DOM and the visual rendering.

## Scope

* Affects `<ul>` and `<ol>` identically.
* Affects `pdf/ua-1` and `pdf/ua-2` identically.
* The visual output is correct in every case — only the tag tree is wrong.

## Things that do not help

Wrapping the list item's text in a block element does not change the hoist, and additionally replaces `LBody` with the wrapper's own tag:

```html
<li><p>First</p><ul>…</ul></li>   ->  LI > Lbl, P     (nested L still hoisted before the LI)
<li><div>First</div><ul>…</ul></li> ->  LI > Lbl, Div   (nested L still hoisted before the LI)
```

## Things that do help

Making the nested list a *sibling* of the `<li>` rather than a child — invalid HTML, but it produces the correct order:

```html
<ul><li>First</li><ul><li>A</li><li>B</li></ul><li>Second</li></ul>
```

```
L
  LI  "First"
  L
    LI  "A"
    LI  "B"
  LI  "Second"
```

The rendering is byte-for-byte identical to the valid markup (same bullets, same indentation), so the hoisting logic appears to be sensitive to whether the nested list is a child box of the list item.

## Possibly related

#2470 (`overflow: hidden` breaks order of tags in PDF/UA) looks like the same family of problem — a box construct causing a tag to be emitted at the wrong position in the tree — though a different trigger. That one was fixed; this case survived.

## Environment

* WeasyPrint 69.0
* Python 3.14.6
* macOS 15.7.3
* Validated with veraPDF 1.30.2 (`-f ua1`) — reports the file as having no PDF/UA-1 violations of this kind.

liZe:
Fix #2874.

### DIFF
--- a/tests/test_pdf.py
+++ b/tests/test_pdf.py
@@ -796,6 +796,29 @@ def test_pdf_tags_split_table_with_caption():
 
 
 @assert_no_logs
+def test_pdf_tags_nested_lists_order():
+    # Regression test for #2874.
+    pdf = FakeHTML(string='''
+      <html lang=fr><ul>
+        <li>a</li>
+        <li>
+          ba
+          <ul>
+            <li>b1</li>
+            <li>b2</li>
+          </ul>
+          bb
+        </li>
+        <li>c</li>
+      </ul>
+    ''').write_pdf(pdf_tags=True)
+    list_children = re.findall(b'/S ?/L ?/K.*\\[(.*)\\]', pdf)
+    for children in list_children:
+        numbers = [int(number) for number in re.findall(b'[1-9][0-9]*', children)]
+        assert sorted(numbers) == numbers
+
+
+@assert_no_logs
 def test_pdf_ua_2_namespace_type():
     # Regression test for #2786.
     pdf = FakeHTML(string='<html lang="en"><body>abc').write_pdf(
--- a/weasyprint/pdf/tags.py
+++ b/weasyprint/pdf/tags.py
@@ -269,12 +269,14 @@ def _build_box_tree(box, parent, pdf, page_number, nums, links, tags):
                         child_parent = parent
                     else:
                         child_parent = element
+                    # Keep the current number of parent children, so that we can add the
+                    # new children before the possible grandchildren, for example for
+                    # lists.
+                    child_index = len(child_parent['K'])
                     child_elements = _build_box_tree(
                         child, child_parent, pdf, page_number, nums, links, tags)
-
-                    # Check if it is already been referenced before.
-                    for child_element in child_elements:
-                        child_parent['K'].append(child_element.reference)
+                    child_parent['K'][child_index:child_index] = [
+                        element.reference for element in child_elements]
 
     else:
         # Add replaced box.


jmuheim:
Tested this branch. Confirmed fixed. Thank you!

Two small notes, both minor.

**1. Text after a nested list still reads before it.** With the fixture from this PR, a reader sees `a, ba, b1, b2, bb, c` on the page, but the tag tree gives `a, ba, bb, b1, b2, c` — `bb` sits in a second `LBody` of the same `LI`, which now precedes the nested `L`:

```
L
  LI  "a"
  LI  LBody "ba"
      LBody "bb"      ← read here
  L
    LI "b1"
    LI "b2"
  LI  "c"
```

Before the PR it was `a, b1, b2, ba, bb, c`, so this is better either way — and it may be inherent to the flat `L` model. Mentioning it only because the fixture here has exactly that shape.

**2. The test checks object numbering, not order.** It asserts that the object numbers in each `/S /L /K` are ascending:

```
broken:  /K [12 0 R  9 0 R]  → fails ✓
fixed:   /K [ 9 0 R 12 0 R]  → passes ✓
```

That works because the `LI`'s object happens to be allocated before the nested `L`'s. If numbers were ever assigned while walking the finished tree, they would be ascending by construction and the test could no longer fail. Asserting the child sequence itself (`LI, LI, L, LI`) would be tied to the actual property.

jmuheim:
Following up on my second note — here is a concrete suggestion. I hope I don't overwhelm you again - it's mainly Claude.ai that pointed me to that.

If the gap in the test doesn't bother you - your proposed fix already works for me, so just merge it. Thank you. 🙏

Same fixture, only the assertion changes:

```python
@assert_no_logs
def test_pdf_tags_nested_lists_order():
    # Regression test for #2874.
    pdf = FakeHTML(string='''
      <html lang=fr><ul>
        <li>a</li>
        <li>
          ba
          <ul>
            <li>b1</li>
            <li>b2</li>
          </ul>
          bb
        </li>
        <li>c</li>
      </ul>
    ''').write_pdf(pdf_tags=True)
    # Structure type of every structure element, by object number.
    types = dict(re.findall(
        rb'(\d+) 0 obj\s*<< ?/Type ?/StructElem ?/S ?(/\w+)', pdf))
    # Children of every list, as structure types, in tag tree order.
    lists = sorted(
        [types[number] for number in re.findall(rb'(\d+) 0 R', children)]
        for children in re.findall(rb'/S ?/L ?/K ?\[([^]]*)\]', pdf))
    assert lists == [[b'/LI', b'/LI'], [b'/LI', b'/LI', b'/L', b'/LI']]
```

It asserts the property the fix is about — the nested `L` comes after the `LI` it belongs to and before the following one — rather than the object numbering. Your `/S ?/L ?/K` regex is unchanged (it matches neither `/S /LI/K` nor `/S /LBody/K`), `re` is already imported, no new helper. `sorted()` keeps it independent of the order the two `L` objects happen to be written in.

Verified against three builds:

| build | children of the outer `L` | current assertion | this one |
| --- | --- | --- | --- |
| 68.1 | `LI, L, LI, LI` | fails | fails |
| 05384ab (this PR's parent) | `LI, L, LI, LI` | fails | fails |
| `nested-lists` | `LI, LI, L, LI` | passes | passes |

One caveat: it also pins the flat `L` model, so it would fail if nested lists ever moved into the parent `LBody` — correct, I think, given that the flat form is the intended one.

*(Suggestion drafted and verified with Claude Code; posted and reviewed by me.)*

liZe:
This test is a non-regression test, it avoids other people or myself to remove the added code (thinking that it’s useless),  or to catch problems that may appear if we optimize/simplify/change the current function. It also improves coverage.

Really checking that the tree is correct, in a reliable way, would require a real PDF parser and a real rule checker. Everything else is just giving a false impression that the test checks the structure, when it just heavily relies on the current implementation. One day we’ll write a test suite to be used with real PDF checkers, that will be actually useful, and the other tests like mine will be gone.

No need to spend more time on this.

> I hope I don't overwhelm you again - it's mainly Claude.ai that pointed me to that.

There’s something contributors have to understand. No hard feelings, but it has to be said:

**If we want to use AI, we can do that, we don’t need you. We don’t want users to repeat what AI tells them.**

Capture d’écran du 2026-08-07 12-58-39

Let me summarize this whole table: **my test gives the same results as yours**. Why did you use a 4×4 table to say this? Because you didn’t, Claude did this for you, because it looks nice and interesting. It really looks smart and deep, like it’s serious job. This is full of emptiness, on a useless topic.

The first rule of the [guidelines](https://www.courtbouillon.org/code-of-conduct/#guidelines-for-contributors) is:

>Use your own words, write with your keyboard.

This sentence is easy to understand.

jmuheim:
> If we want to use AI, we can do that, we don’t need you. We don’t want users to repeat what AI tells them.

I see that now. I learned something. Thank you. 🙏 I'll be more conscious next time and use my own keyboard.

PR #2884, opened 2026-08-06
Keep absolutely positioned flex children in document order

jmuheim:
Absolutely positioned children of a flex container are collected at the front of the children list `flex_layout` passes on, whatever their position in the document:

```python
box = box.copy_with_children(
    [child for child in children if child.is_absolutely_positioned()])
```

`pdf/tags.py` builds the PDF structure tree from that list, so a tagged PDF hands assistive technologies a reading order in which, say, a logo positioned into a corner comes before the heading it belongs to. Block containers keep the order. This is what `test_flex_absolute_positioning` is marked `xfail` for ("Order is not kept when out-of-flow and in-flow children are mixed").

The placeholders already record their index (`placeholder.index`), so this puts them back at it once the flex items are laid out. Absolutely positioned boxes are painted from the stacking context, so painting is unaffected — no layout changes, the previously xfailing test now asserts both children's positions.

Tests: `pytest` and `ruff check` on macOS/Python 3.14. Same 7 failures as on unpatched `main` here (`font-variant-caps`, `font-stretch`, `acid2` — missing fonts locally), no new ones, and the `xfail` above turns green.

One related thing I did **not** touch: when a flex container is fragmented, its absolutely positioned children are laid out again for every fragment, so they are painted — and tagged — on every page the container spans. The loop at `flex.py:132` does not honour the `skip_stack`, unlike its counterpart in `block.py`. Repro:

```html
<style>
  @page { size: 300px 200px; margin: 5px }
  section { position: relative; display: flex; flex-direction: column }
  a { position: absolute; right: 0; bottom: 0 }
  p { margin: 0 0 60px }
</style>
<section>
  <div><p>one</p><p>two</p><p>three</p><p>four</p></div>
  <a href="https://example.com/">LOGO</a>
</section>
```

`LOGO` appears on all three pages. Happy to open a separate issue for it.

### DIFF
--- a/tests/layout/test_position.py
+++ b/tests/layout/test_position.py
@@ -494,9 +494,7 @@ def test_grid_relative_positioning():
 
 
 @assert_no_logs
-@pytest.mark.xfail
 def test_flex_absolute_positioning():
-    """TODO: Order is not kept when out-of-flow and in-flow children are mixed."""
     page, = render_pages('''
       <style>
         @page { size: 100px 100px }
@@ -512,9 +510,9 @@ def test_flex_absolute_positioning():
     html, = page.children
     body, = html.children
     article, = body.children
-    # That’s currently div2, div1
     div1, div2 = article.children
 
+    assert (div1.position_x, div1.position_y) == (0, 0)
     assert (div2.position_x, div2.position_y) == (10, 10)
 
 
--- a/tests/test_pdf.py
+++ b/tests/test_pdf.py
@@ -796,6 +796,21 @@ def test_pdf_tags_split_table_with_caption():
 
 
 @assert_no_logs
+def test_pdf_tags_absolute_in_flex():
+    # The structure tree is built from the box tree, so an absolutely
+    # positioned flex child listed first would be read out first too.
+    pdf = FakeHTML(string='''
+      <html lang="en">
+        <div style="display: flex; position: relative">
+          <h1>heading</h1>
+          <ul style="position: absolute"><li>item</li></ul>
+        </div>
+    ''').write_pdf(pdf_tags=True, uncompressed_pdf=True)
+    tags = re.findall(rb'/S (/\w+)', pdf)
+    assert tags.index(b'/H1') < tags.index(b'/L')
+
+
+@assert_no_logs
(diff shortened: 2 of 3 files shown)


jmuheim:
One thing I understated in the description, and I think it deserves saying plainly: this is not a cosmetic ordering detail, it is an accessibility defect.

The order of the structure tree **is** the reading order for assistive technology. Getting it wrong fails WCAG 2.2 [SC 1.3.2 Meaningful Sequence](https://www.w3.org/WAI/WCAG22/Understanding/meaningful-sequence.html) (Level A), and ISO 14289-1 requires the structure hierarchy to reflect the document's logical reading order. So any tagged PDF produced from a flex container with absolutely positioned children — a very ordinary layout — is non-conforming in a way that is invisible to the person producing it.

What makes it worse is that **no validator catches it.** Reading order is not machine-determinable (unless you directly compare the source html to the generated pdf), so it is outside veraPDF's PDF/UA-1 profile. In our case the affected files passed `--pdf-variant=pdf/ua-1` without a warning and passed veraPDF's `ua1` flavour clean, and we ship that check in CI. The defect surfaced only when a screen-reader expert proof-read the generated pdfs.

I say this with genuine appreciation for the work behind the PDF/UA support — it is the reason we chose WeasyPrint over the alternatives, and as far as I know it is still the only free tool that gets tagged, accessible PDF right end to end. That is exactly why this matters more here than it would elsewhere: a silent reading-order bug undercuts the one guarantee the PDF/UA output is chosen for, and it does so without the user ever getting a signal.

The fix itself is nine lines and uses an index the code already records. Happy to adjust anything about the approach.

That said: This is another Claude.ai-driven thing. I'm no Python-coder, so I'm limited in giving any other support than naming accessibility requirements here. And I'm off to holidays for 2-3 weeks now; hopefully we will be able to ship this some time in September. Thank you for your effort! 🙏

grewn0uille:
Please follow [contributors guidelines](https://www.courtbouillon.org/code-of-conduct/#guidelines-for-contributors) for your next contributions.

jmuheim:
Understood — sorry. I'll report and wait next time. But after my holidays.

Issue #1211, opened 2020-09-10
Generating low- and high resolution pdf versions from the same html


arjensuijker:
I want to generate two pdf files from my html: one with high resolution images and one with lower resolutions. I'm using a Wordpress as the back-end, which already generates the necessary image resolutions and includes the correct src-set in image tags. Can Weasyprint be configured to select a defined image resolution from this set?

liZe:
It’s a feature I’ve already tried to implement for @arthru earlier this year. But it’s hard.

What’s possible is to use a custom [`url_fetcher`](https://weasyprint.readthedocs.io/en/stable/tutorial.html#url-fetchers) function. Depending on what you really want, and on how the different URLs are generated (is there a pattern depending on the resolution?), it might be possible to solve your problem just with a custom function.

But. If you really want to get an image based on the output resolution, then it’s logically almost impossible to have something perfect. As you can use the same image at multiple places in a document, possibly with different sizes, you can’t associate only one resolution with one URL. And of course, you don’t want to include an image multiple times in the PDF when it’s used with different resolutions.

And it’s even more complicated. In WeasyPrint, when the image is downloaded, we don’t know its size yet in the layout. Actually, its size (and thus its resolution) in the document can depend on its intrinsic size. With other words, the output size depends on the image size, and so it’s difficult to make the image size depend on the output size.

That’s bad news.

But even if it’s complicated, we may find a solution. `src-set` defines different sources, and its specification may give some hints about how to choose the source wisely. Other browsers can do this, we should be able to find a solution too.

arjensuijker:
Good to know I didn't miss an existing feature :) A well-implemented srcset should give enough information to decide which image is needed where. That indeed assumes that the CSS is interpreted before the images are placed (or downloaded). If this feature would be implemented, it would be amazing if we could use the DPI flag (currently only used for png output) to set the desired resolution, just like browsers choose the correct resolution image for normal or retina screens.

Unfortunately I'm not in a situation where I can offer much help with developing this feature. I'd be happy to help test it though!

I'll also look into the custom url_fetcher, that's a solid lead.

petermolnar:
I'd like to revive this discussion. I've recently change the way I add image to my site as:

```html
<figure property="image" vocab="http://schema.org/" typeof="ImageObject">
    <span property="name" content="pwll-y-wrach-waterfall"></span>
    <a href="pwll-y-wrach-waterfall_huge.jpg" class="u-photo u-featured" property="url">
[...]

PR #2895, opened 2026-08-17
Keep JPEG quality when resizing images for DPI

jurriaan:
I ran into this while trying to use `dpi` together with `jpeg_quality`.

As discussed with @liZe this is a bug in the current implementation where the jpeg is re-encoded twice with the second time ignoring the jpeg_quality setting.

I changed this so that the JPEG is resized before it is encoded with the requested quality.

### DIFF
--- a/tests/test_pdf.py
+++ b/tests/test_pdf.py
@@ -6,6 +6,7 @@ import re
 from codecs import BOM_UTF16_BE
 
 import pytest
+from PIL import Image
 
 from weasyprint import Attachment
 from weasyprint.document import Document, DocumentMetadata
@@ -498,6 +499,21 @@ def test_embed_jpeg():
     assert b'/Filter /DCTDecode' in FakeHTML(
         base_url=resource_path('dummy.html'),
         string='<img src="blue.jpg">').write_pdf()
+
+
+@assert_no_logs
+def test_resize_jpeg_with_quality():
+    image = Image.open(resource_path('not-optimized.jpg'))
+    image.thumbnail((5, 5))
+    expected = io.BytesIO()
+    image.save(expected, format='JPEG', optimize=False, quality=1)
+
+    pdf = FakeHTML(
+        base_url=resource_path('dummy.html'),
+        string='<img src="not-optimized.jpg">',
+    ).write_pdf(dpi=48, jpeg_quality=1)
+
+    assert expected.getvalue() in pdf
 
 
 @assert_no_logs
--- a/weasyprint/images.py
+++ b/weasyprint/images.py
@@ -34,6 +34,8 @@ class ImageLoadingError(ValueError):
 class RasterImage:
     def __init__(self, pillow_image, image_id, image_data, filename=None,
                  cache=None, orientation='none', options=DEFAULT_OPTIONS):
+        source_image_data = image_data
+        source_filename = filename
         # Transpose image
         original_pillow_image = pillow_image
         pillow_image = rotate_pillow_image(pillow_image, orientation)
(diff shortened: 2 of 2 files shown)


liZe:
Hi!

After checking the code more deeply, this problem is a bit more complicated than expected. Here are a few details about how things work before your patch:

1. We download an image when we meet one in HTML/CSS. We need to download it before the layout, because the layout sometimes needs the intrinsic size of the image.
2. This image is cached, on disk or in memory. As we want to optimize memory, optimization and image quality is applied before cache. Possible format transformation (we only embed formats supported by PDF) is done here too.
3. When the image is actually embedded, we have to resize images if a dpi option is set. This step can only be done when the whole layout has been done, because we need to know the physical size of this image for each time it appears on the PDF.

We actually have two solutions to respect all these conditions:

- The current solution is to apply optimizations and format transformations before the layout, and resize the image after the layout. The problem of this solution is that, if there’s both an optimization/conversion/quality and a resize, we need two steps. Optimization is done twice, JPEG quality is only set for the first step (your original problem). Even if we applied the JPEG quality twice, these two steps are not optimal.
- Your solution does the optimization, quality and resize operations in one step, after layout. The problem is that it caches the original (possibly high-quality) image in memory. We could improve the situation by using WeasyPrint’s cache instead (that can handle both in-memory and on-disk cache according to the user’s configuration, and defaults to memory), but the problem remains: it takes more space.

Unfortunately, this problem of memory optimization for images is a real one for many users: we’ve had GitHub issues and real clients complained, that’s why we have this complex workflow now. 😁

As disk access is often fast and temporary disk storage is often cheap, we *could* change the default workflow to something way more simple: cache the original image on disk, and apply all transforms only after the layout. To be honest, the main problem I see for that, other than bugs that will happen because of corner cases we can’t even imagine 😁, is that we’ll have to manage temporary files by default, and deleting temporary files reliably [is a bit complex](https://github.com/Kozea/WeasyPrint/issues/2090). We also have to find a way to handle rotation to get width/height correctly without altering the original image. But, well, that may be a good default setting… if nobody relies on the fact that we cache optimized images!

Any thoughts about this?

jurriaan:
@liZe Hey, sorry for the late response!

For our case another workflow would probably be more ideal. For almost all assets we know the size in advance. I was thinking, what if there would be some kind of possibility to pass that information to weasyprint (using srcset?) to make this phase somewhat simpler. 

Besides that, we run a image proxy (thumbor) in front of our assets buckets which also handles resizing and caching of those variants, so we could make it work more efficiently by making better use of that. 

If weasyprint would be able to use the information we already have about the assets to do the layout passes without actual image data that would be more efficient. 

And, when on top of that there would be hooks to do the resizing and request a resized image from thumbor directly, bypassing the resizing in the python process, it could help a lot performance-wise for our specific use case.

What do you think about this? Is this something that could work for you as well?

liZe:
> what if there would be some kind of possibility to pass that information to weasyprint (using srcset?) to make this phase somewhat simpler.

Supporting srcset would be a good idea. I should really read the spec to see how browsers handle that, see #1211.

> For almost all assets we know the size in advance.

If you know the size, then you can already use the URL of the asset of this size, and avoid the whole resize step in WeasyPrint by not using the `dpi` option, can’t you?

(Maybe we’ll need a real-life example at some point!)

septatrix:
> * We could improve the situation by using WeasyPrint’s cache instead (that can handle both in-memory and on-disk cache according to the user’s configuration, and defaults to memory), but the problem remains: it takes more space.

I think this would be the easiest solution. For most cases the on-disk cache is not required and if the proposed alternative is to always write stuff to disk it seems better to just reuse the existing cache implementation and defer that decision to the user.

Apart from that, supporting srcset would be great anyhow

PR #2904, opened 2026-08-30
Support text-emphasis properties

### LINKED ISSUE #2024: Could you support text-emphasis and writing-mode?
I want these feature:
- [text-emphasis](https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis)
- [text-emphasis-position](https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis-position)
- [writing-mode](https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode)

But these feature is not support on latest version 60.2

Thank you.

i-am-paradox:
This adds support for the `text-emphasis` properties from the Text Decoration Level 4 draft: `text-emphasis-style`, `text-emphasis-color`, `text-emphasis-position`, and the `text-emphasis` shorthand.

A few notes on the implementation:

- Marks are drawn at half the font size, above the ascent or below the descent depending on the position keyword. One mark is drawn per grapheme cluster, centered on it, and whitespace clusters are skipped.
- The mark itself is shaped with a small Pango layout instead of looking it up in the text font directly. That way font fallback can find a glyph for marks the text font doesn't have, which is the common case for sesame or double-circle marks in Latin fonts.
- Only the vertical keywords of `text-emphasis-position` are supported (`over`/`under`). The horizontal keyword has to be `right`, the initial value, since there are no vertical writing modes in WeasyPrint. A warning is logged when another value is used.
- `text-emphasis: none` resets the style, as it's an inherited property.
- Line boxes are not grown to make room for the marks yet, they can overflow above the line, in the same way `text-decoration` overlines can. I can look at reserving ruby space in the layout as a follow-up if that's wanted.
- `writing-mode` support from the same issue is left out, it seems like a much bigger project.

Tests: 38 validation tests for the three longhands, 16 expander tests for the shorthand, and 2 pixel tests drawing filled dots over and under text. All pass. On this machine the full suite gives 2 other failures (`test_font_stretch`, `test_acid2` and the 5 `test_font_variant_caps_*` ones), which fail the same way on a clean main checkout, they're unrelated to this change.

Fixes #2024

### DIFF
--- a/weasyprint/css/properties.py
+++ b/weasyprint/css/properties.py
@@ -268,11 +268,9 @@ INITIAL_VALUES = {
     'text_decoration_style': 'solid',
     'text_decoration_thickness': 'auto',
     'text_underline_offset': 'auto',
-
-    # Text Emphasis Module (WD): https://drafts.csswg.org/css-text-decor-4/
-    'text_emphasis_style': None,
+    'text_emphasis_style': 'none',
     'text_emphasis_color': 'currentcolor',
-    'text_emphasis_position': 'over right',
+    'text_emphasis_position': ('over',),
 
     # Overflow Module 3/4 (WD): https://www.w3.org/TR/css-overflow-4/
     'block_ellipsis': 'none',
--- a/weasyprint/css/properties.py
+++ b/weasyprint/css/properties.py
@@ -269,6 +269,11 @@ INITIAL_VALUES = {
     'text_decoration_thickness': 'auto',
     'text_underline_offset': 'auto',
 
+    # Text Emphasis Module (WD): https://drafts.csswg.org/css-text-decor-4/
+    'text_emphasis_style': None,
+    'text_emphasis_color': 'currentcolor',
+    'text_emphasis_position': 'over right',
+
     # Overflow Module 3/4 (WD): https://www.w3.org/TR/css-overflow-4/
     'block_ellipsis': 'none',
     'continue': 'auto',
@@ -348,6 +353,9 @@ INHERITED = {
     'tab_size',
     'text_align_all',
     'text_align_last',
+    'text_emphasis_color',
+    'text_emphasis_position',
+    'text_emphasis_style',
     'text_indent',
     'text_transform',
     'text_underline_offset',
--- a/weasyprint/css/validation/properties.py
+++ b/weasyprint/css/validation/properties.py
@@ -129,6 +129,7 @@ def background_attachment(keyword):
 @property('border-inline-end-color')
 @property('column-rule-color', unstable=True)
 @property('text-decoration-color')
+@property('text-emphasis-color')
 @single_token
 def other_colors(token):
     if parse_color(token):
@@ -1293,25 +1294,18 @@ def text_indent(token):
 
 
 @property()
-@single_token
-def text_emphasis_color(token):
-    """``text-emphasis-color`` property validation."""
-    if result := parse_color(token):
-        return 'inherit' if result == 'currentcolor' else token
-
-
-@property()
 def text_emphasis_position(tokens):
     """``text-emphasis-position`` property validation."""
     keywords = {get_keyword(token) for token in tokens}
     if len(tokens) != len(keywords):
         return
-    for vertical in ('over', 'under'):
-        for horizontal in ('right', 'left'):
-            if keywords == {vertical, horizontal}:
-                return f'{vertical} {horizontal}'
-    if keywords == {'over'} or keywords == {'under'}:
-        return f'{keywords.copy().pop()} right'
+    if not keywords <= {'over', 'under', 'right', 'left'}:
+        return
+    if len({keyword for keyword in keywords if keyword in ('over', 'under')}) > 1:
+        return
+    if len({keyword for keyword in keywords if keyword in ('right', 'left')}) > 1:
+        return
+    return tuple(keywords)
 
 
 @property()
--- a/weasyprint/css/validation/properties.py
+++ b/weasyprint/css/validation/properties.py
@@ -1293,6 +1293,51 @@ def text_indent(token):
 
 
 @property()
+@single_token
+def text_emphasis_color(token):
+    """``text-emphasis-color`` property validation."""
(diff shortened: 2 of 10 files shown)


weasyprint/css/properties.py
```diff
@@ -269,6 +269,11 @@
     'text_decoration_thickness': 'auto',
     'text_underline_offset': 'auto',
 
+    # Text Emphasis Module (WD): https://drafts.csswg.org/css-text-decor-4/
```
liZe: This is the same specification as the previous one, you can group the properties together.

weasyprint/css/properties.py
```diff
@@ -269,6 +269,11 @@
...
+    'text_emphasis_style': None,
+    'text_emphasis_color': 'currentcolor',
+    'text_emphasis_position': 'over right',
```
liZe: We should have a tuple for the computed value.

weasyprint/draw/text.py
```diff
@@ -302,3 +305,137 @@ def draw_text_decoration(stream, textbox, offset_x, offset_y, thickness, color):
         stream, textbox.position_x, textbox.position_y + offset_y,
         textbox.position_x + textbox.width, textbox.position_y + offset_y,
         thickness, textbox.style['text_decoration_style'], color, offset_x)
+
+
+# Maps (fill, shape) keyword pairs to Unicode codepoints, as defined in
```
liZe: We can put that in `text/contsants.py` instead.

weasyprint/css/validation/properties.py
```diff
@@ -1292,6 +1292,51 @@ def text_indent(token):
...
+@single_token
+def text_emphasis_color(token):
+    """``text-emphasis-color`` property validation."""
+    if result := parse_color(token):
+        return 'inherit' if result == 'currentcolor' else token
+
+
+@property()
+def text_emphasis_position(tokens):
+    """``text-emphasis-position`` property validation."""
```
liZe, line 1298: You should return a tuple.

weasyprint/css/validation/properties.py
```diff
@@ -1292,6 +1292,51 @@ def text_indent(token):
...
+    keywords = {get_keyword(token) for token in tokens}
+    if len(tokens) != len(keywords):
+        return
+    for vertical in ('over', 'under'):
+        for horizontal in ('right', 'left'):
+            if keywords == {vertical, horizontal}:
+                return f'{vertical} {horizontal}'
+    if keywords == {'over'} or keywords == {'under'}:
+        return f'{keywords.copy().pop()} right'
```
liZe: The computed value is the specified keywords, so it means that we should return `tuple(keywords)` with a single value here, and use the default `right` value when actually drawing. It looks like it’s the same, but we've had problems with other properties in the past because of that, so let’s blindly follow the spec here. 😁

weasyprint/css/properties.py
```diff
@@ -269,6 +269,11 @@
     'text_decoration_thickness': 'auto',
     'text_underline_offset': 'auto',
 
+    # Text Emphasis Module (WD): https://drafts.csswg.org/css-text-decor-4/
+    'text_emphasis_style': None,
```
liZe: Let’s store `'none'` here, so that we can avoid the double `'none'`/`None` check everywhere.

weasyprint/css/validation/properties.py
```diff
@@ -1292,6 +1292,51 @@ def text_indent(token):
...
+
+
+@property()
+def text_emphasis_style(tokens):
+    """``text-emphasis-style`` property validation."""
```
liZe, line 1313: I’ll simplify that later. 😁

weasyprint/css/validation/properties.py
```diff
@@ -1292,6 +1292,51 @@ def text_indent(token):
         return length
 
 
+@property()
+@single_token
+def text_emphasis_color(token):
+    """``text-emphasis-color`` property validation."""
```
liZe: You can add `'text-emphasis-color'` to `other_colors` instead.

weasyprint/draw/text.py
```diff
@@ -302,3 +305,137 @@ def draw_text_decoration(stream, textbox, offset_x, offset_y, thickness, color):
...
+    ('open', 'double-circle'): '◎',
+    ('filled', 'triangle'): '▲',
+    ('open', 'triangle'): '△',
+    ('filled', 'sesame'): '﹅',
+    ('open', 'sesame'): '﹆',
+}
+
+
+def draw_text_emphasis(stream, textbox, x, y):
+    """Draw emphasis marks of ``textbox`` to a ``pdf.stream.Stream``.
```
liZe, line 312: Oh, I first thought that that was just decoration, not text, so… that’s not an easy fix after all! 😁

I’ll take some time to check that carefully, and see if we can also take care of the line height consideration.

liZe (changes requested):
Thanks, that’s great!

Here are a few minor remarks, the overall implementation is cool. I thought that it was a simple decoration, not ruby-like text, so that’s pretty hard for a first contribution!

When the requested changes are applied, I’ll see if it’s possible to make the rendering part closer to what ruby text requires (including the line height management) and to filter the characters that are in the right Unicode categories, so that it can be a more solid base for ruby in the future.

It should be ready to merge in version 71. 🤞

i-am-paradox:
Thanks for the detailed review! All remarks are applied:

- Initial values are now grouped with the other Text Decoration properties, and text-emphasis-style starts at 'none' so the None/none double check is gone.
- text-emphasis-position keeps the specified keywords in a tuple, and 'right' is used as the default when drawing.
- text-emphasis-color is validated with the other colors.
- The mark map moved to text/constants.py.
- The fail_json simplification you mentioned is kept as-is for now.

The ruby-like aspects are indeed a rabbit hole, happy to see it become a base for ruby. I can look at reserving space for the marks in the line box if that helps, or leave that to you as discussed, whatever fits best for 71.

liZe:
Thanks for the changes, we’ll get back after 70 is released.

PR #2907, opened 2026-09-05
Align rounding with CSS spec

### LINKED ISSUE #2906: Rounding not aligned with CSS spec
In `css/__init__.py`, the rounding uses the inbuilt Python `round` function, which rounds x.5 to the nearest even number to x. However, the [CSS values spec](https://drafts.csswg.org/css-values-4/#combine-integers) states that x.5 should round towards $+\infty$.

CalculatorAddict:
Aligns rounding in weasyprint.css with [CSS values 4 spec](https://drafts.csswg.org/css-values-4/#combine-integers) by replacing the inbuilt Python `round` function with a `lambda` which algebraically forces rounding upwards rather than the inbuilt's banker rounding.

Edited test coverage to add a case here. I felt it was most natural to add cases to `test_math_functions` rather than add a new test for just the rounding. This required refactoring the Pytest mark to pass in an `expected` argument as well so that I could test that the rounding is correct for both even and odd numbers. This was needed because the old test just checked `isclose(div.width, 100)` which would not have provided sufficient coverage for the new rounding (since it only diverges from Python's inbuilt on odd numbers).

Fixes #2906

### DIFF
--- a/tests/css/test_math.py
+++ b/tests/css/test_math.py
@@ -10,81 +10,85 @@ from ..testing_utils import assert_no_logs, capture_logs, render_pages
 
 
 @assert_no_logs
-@pytest.mark.parametrize('width', [
-    'calc(100px)',
-    'calc(10em)',
-    'calc(50vw)',
-    'calc(20pvh)',
-    'calc(50%)',
-    'calc(10px + 90px)',
-    'calc(5em + 50px)',
-    'calc(2 * 5em)',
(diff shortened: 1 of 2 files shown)


liZe:
I’ll keep your commit as a base and remove the todo, to avoid bots automatically trying to fix that again and again.

I don’t think that we need this huge changes to test something nobody cares about in real life. 😁

PR #2908, opened 2026-09-06
Reserve bottom padding and border when detecting overflow

### LINKED ISSUE #2798: WeasyPrint truncates PDF files when HTML carries style height:100% and encounters unlucky layout

gyanu2507:
With the default `box-decoration-break: slice`, layout ignored bottom padding and border when deciding whether a block overflowed the page. An unbreakable bottom decoration (the `<hr>` in #2798) could then overflow unnoticed, and remaining siblings were dropped.

Bottom padding and border are now always reserved for overflow detection. Margin is still only added when the fragment actually draws the bottom decoration.

Fixes #2798

### DIFF
--- a/tests/layout/test_page.py
+++ b/tests/layout/test_page.py
@@ -1832,3 +1832,33 @@ def test_running_float():
         Hello!
       </footer>
     ''')
+
+
+@assert_no_logs
+def test_fixed_height_root_keeps_overflowing_siblings():
+    """Content after an overflowing bottom decoration must still be laid out.
+
+    See https://github.com/Kozea/WeasyPrint/issues/2798
+    """
+    pages = render_pages('''
+      <html style="height:100%">
+        <body>
+          <h3 style="padding-top: 914.8px;">Header</h3>
+          <hr/>
+          Overflow
+        </body>
+      </html>
+    ''')
+    texts = []
+
+    def collect(box):
+        if isinstance(box, boxes.TextBox) and box.text:
+            texts.append(box.text)
+        children = getattr(box, 'children', None)
+        if children:
+            for child in children:
+                collect(child)
+
+    for page in pages:
+        collect(page)
+    assert 'Overflow' in ''.join(texts)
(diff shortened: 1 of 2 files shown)


liZe:
Hi,

This pull request breaks tests and don’t fix the examples I provided in #2798.

Capture d’écran du 2026-09-06 10-24-53

Please follow the [guidelines](https://www.courtbouillon.org/code-of-conduct/#guidelines-for-contributors), assume the fact that you use a bot to build your CV, and don’t waste our time.

PR #2915, opened 2026-09-08
Restore nested SVG viewport size on the SVG object

### LINKED ISSUE #2914: Percentages resolve against a stale viewport after a nested <svg>
After a nested `<svg>` is drawn, its size is left behind as the percentage reference, so all later percentages resolve against the wrong viewport.

### Reproduce

```html
<style>@page { size: 8px } svg { display: block }</style>
<svg width="8px" height="8px" xmlns="http://www.w3.org/2000/svg">
    <svg width="100%" height="100%" viewBox="0 0 4 4"></svg>
    <rect x="50%" y="50%" width="2" height="2" fill="blue"/>
</svg>
```

The blue rect should start at `x=4` (50% of 8px). It starts at `x=2` (50% of the nested `viewBox`).

### Fix

`weasyprint/svg/__init__.py` `draw_node` restores the size on `svg`, but there's no local `svg` there. Changing to `self` fixes it:

```diff
         # Restore concrete and inner size of root svg tag
         if node.tag == 'svg':
-            self.tree.set_svg_size(svg, concrete_width, concrete_height)
+            self.tree.set_svg_size(self, concrete_width, concrete_height)
```

havelaer:
Fixes [Issue 2914](https://github.com/Kozea/WeasyPrint/issues/2914):

The concrete size of the svg tag is properly restored.

### DIFF
--- a/tests/draw/svg/test_units.py
+++ b/tests/draw/svg/test_units.py
@@ -99,3 +99,81 @@ def test_units_unknown(assert_pixels):
               stroke-width="2px" stroke="red" fill="none" />
       </svg>
     ''')
+
+
+@assert_no_logs
+def test_units_percentage(assert_pixels):
+    assert_pixels('''
+        ________
+        ________
+        ________
+        ________
+        ____BB__
+        ____BB__
+        ________
+        ________
+    ''', '''
+      <style>
+        @page { size: 8px }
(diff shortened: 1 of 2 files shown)


liZe:
Thank you!

PR #2920, opened 2026-09-10
Don’t duplicate entries in PDF/A AF entry

### LINKED ISSUE #2917: /AF array lists every attachment twice in PDF/A-3 output
Hello,

This issue has been written with the help of Claude Code. 
I made my best to write a useful issue report and avoid any slope as much as I can.
I have run myself all the bash command on my terminal (GNU/Linux - TuxedoOS up to date) and study the output.
I am not an expert on the PDF format but the output returned by the bash and Claude code points looks logical to me that there's a duplication in what weasyprint is generating in the PDF.

So now the issue :-)
With the parameter `--pdf-variant=pdf/a-3b`, the catalog's `/AF` array holds each attachment twice, although a single `/Filespec` object exists for it:

``` bash
~/_poc/bug-weasyprint
$ python3 -m venv venv
~/_poc/bug-weasyprint
$ ./venv/bin/pip install --upgrade pip
Requirement already satisfied: pip in ./venv/lib/python3.12/site-packages (24.0)
Collecting pip
  Downloading pip-26.2.1-py3-none-any.whl.metadata (4.6 kB)
Downloading pip-26.2.1-py3-none-any.whl (1.8 MB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.8/1.8 MB 38.6 MB/s eta 0:00:00
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 24.0
    Uninstalling pip-24.0:
      Successfully uninstalled pip-24.0
Successfully installed pip-26.2.1
~/_poc/bug-weasyprint
$ ./venv/bin/pip install weasyprint==70.0
Collecting weasyprint==70.0
  Using cached weasyprint-70.0-py3-none-any.whl.metadata (3.8 kB)
...
...
Installing collected packages: brotli, zopfli, webencodings, Pyphen, pydyf, pycparser, Pillow, fonttools, tinyhtml5, tinycss2, cffi, cssselect2, weasyprint
Successfully installed Pillow-12.3.0 Pyphen-0.18.1 brotli-1.2.0 cffi-2.1.1 cssselect2-0.10.1 fonttools-4.64.0 pycparser-3.0 pydyf-0.12.1 tinycss2-1.5.1 tinyhtml5-2.1.0 weasyprint-70.0 webencodings-0.6.1 zopfli-0.4.3
~/_poc/bug-weasyprint
$ ./venv/bin/weasyprint --version
WeasyPrint version 70.0
~/_poc/bug-weasyprint
$ ./venv/bin/python -c "import sys, weasyprint, pydyf; print(sys.version.split()[0], weasyprint.__version__, pydyf.__version__)"
3.12.3 70.0 0.12.1
~/_poc/bug-weasyprint
$ printf '<!DOCTYPE html>\n<html><body><h1>Invoice 42</h1></body></html>\n' > invoice.html
~/_poc/bug-weasyprint
$ echo '<x/>' > factur-x.xml
~/_poc/bug-weasyprint
$ ./venv/bin/weasyprint invoice.html out.pdf \
  --attachment=factur-x.xml --attachment-relationship=Alternative --pdf-variant=pdf/a-3b
~/_poc/bug-weasyprint
$ ls
factur-x.xml  invoice.html  out.pdf  venv
~/_poc/bug-weasyprint
$ mutool clean -d -a out.pdf decomp.pdf
~/_poc/bug-weasyprint
$ ls
decomp.pdf  factur-x.xml  invoice.html  out.pdf  venv
~/_poc/bug-weasyprint
$ grep -ao '/AF *\[[^]]*\]' decomp.pdf
/AF [ 11 0 R 11 0 R ]
~/_poc/bug-weasyprint
$ grep -c '/Type /Filespec' decomp.pdf
1
~/_poc/bug-weasyprint
$ grep -ao '/Producer *([^)]*)' decomp.pdf
/Producer (WeasyPrint 70.0)
~/_poc/bug-weasyprint
$ 

```

If you look at the bash above, you'll see that I attached only one file but the PDF says they're two: `/AF [ 11 0 R 11 0 R ]`.

This is not a validation failure: veraPDF passes such a file:
``` bash
~/_poc/bug-weasyprint
$ docker run --rm -v "$PWD":/data verapdf/cli:v1.30.2 --flavour 3b --format text out.pdf
Unable to find image 'verapdf/cli:v1.30.2' locally
v1.30.2: Pulling from verapdf/cli
db5b5e075fa3: Pull complete 
780851c4d3fa: Pull complete 
988a3cb80ef0: Pull complete 
6a0ac1617861: Pull complete 
803a70357bd1: Pull complete 
abaf802d09f9: Pull complete 
Digest: sha256:d5ee329657cf9bc4b2400392dd54c7d0a0ce9980ff6fa2da5590eebeec007cdb
Status: Downloaded newer image for verapdf/cli:v1.30.2
PASS /data/out.pdf 3b
```

And duplicates aren't forbidden anywhere Claude Code could find, so the two reasons I am posting here anyway: 
- A reader that lists associated files from `/AF` can show the same attachment twice
- The exemple Factur-X invoices Claude Code downloaded from FNFE-MPE all carry exactly one reference per attachment.

liZe:
Fix #2917.

### DIFF
--- a/tests/test_pdf.py
+++ b/tests/test_pdf.py
@@ -755,6 +755,16 @@ def test_custom_rdf_metadata():
 
 
 @assert_no_logs
+def test_pdf_a_attachments():
+    html = '<title>test</title><link rel="attachment" href="data:,hi"><body>test'
+    pdf_document = FakeHTML(string=html).render()
+    pdf_bytes = pdf_document.write_pdf(
+        pdf_variant='pdf/a-3b', pdf_identifier=b'example-bytes', uncompressed_pdf=True)
+    assert pdf_bytes.count(b'AFRelationship') == 1
+    assert re.search(rb'/AF \[\d+ \d+ R\]', pdf_bytes)
+
+
+@assert_no_logs
 def test_color_spaces_with_icc_output_intent():
     shading_pdf = FakeHTML(string='''
       <style>html { background: linear-gradient(red, blue) }</style>
--- a/weasyprint/pdf/pdfa.py
+++ b/weasyprint/pdf/pdfa.py
@@ -32,12 +32,6 @@ def pdfa(pdf, document, page_streams, attachments, compress, version, variant):
         relationships = {
             f'<{attachment.md5}>': attachment.relationship
             for attachment in attachments if attachment.md5}
-        pdf_attachments = []
-        if 'Names' in pdf.catalog and 'EmbeddedFiles' in pdf.catalog['Names']:
-            reference = int(pdf.catalog['Names']['EmbeddedFiles'].split()[0])
-            names = pdf.objects[reference]
-            for name in names['Names'][1::2]:
-                pdf_attachments.append(name)
         for pdf_object in pdf.objects:
             if not isinstance(pdf_object, dict):
                 continue
@@ -47,11 +41,9 @@ def pdfa(pdf, document, page_streams, attachments, compress, version, variant):
             checksum = pdf.objects[reference].extra['Params']['CheckSum']
             relationship = relationships.get(checksum, 'Unspecified')
             pdf_object['AFRelationship'] = f'/{relationship}'
-            pdf_attachments.append(pdf_object.reference)
-        if pdf_attachments:
             if 'AF' not in pdf.catalog:
                 pdf.catalog['AF'] = pydyf.Array()
-            pdf.catalog['AF'].extend(pdf_attachments)
+            pdf.catalog['AF'].append(pdf_object.reference)
 
     # Print annotations.
     for pdf_object in pdf.objects:


Yanal-Yves:
I learned a bit of PDF syntax, I am a total newbie on PDF :-). To be able to review this PR. I understood the basics of the main structure of a PDF, the `/AF` array  and the `/Names/EmbeddedFiles` name tree. 

I reviewed the code change carefully. I understand that your fix consists of removing the iteration of the `/Names/EmbeddedFiles` name tree that does not contains all the attachments (ex: it does not contains the `<a rel="attachment"`). 

I also reviewed the unit test. It does a minimal coverage of the fix (it fails on `main` and passes with the fix). May I suggest to replace your unit test `test_pdf_a_attachments` a more complex unit test `test_pdf_a_associated_files`:

``` python
@assert_no_logs
def test_pdf_a_associated_files():
    # The three ways to attach a file: the API, a <link> in the head, an <a> in
    # the body. Only the first two are registered in the EmbeddedFiles name tree.
    html = ('<title>test</title><link rel="attachment" href="data:,link-bytes">'
            '<body><a rel="attachment" href="data:,anchor-bytes">link</a>')
    pdf = FakeHTML(string=html).write_pdf(
        attachments=[Attachment('data:,api-bytes')],
        pdf_variant='pdf/a-3b', uncompressed_pdf=True)
    assert pdf.count(b'/Type /Filespec') == 3
    associated_files, = re.findall(rb'/AF \[(.*?)\]', pdf)
    references = re.findall(rb'\d+ 0 R', associated_files)
    assert len(references) == 3, references
    assert len(set(references)) == 3, references
```

What I like with this test over the suggested one : 
- Your test still passes if the `/AF` entries are collected by the first loop instead of the second one. I tried it: keep the first loop, and remove only the `pdf_attachments.append(pdf_object.reference)` line from the second one. In that case, your test is green, because a `<link>` attachment is in the name tree anyway. My test fails in that case.
- Test the three ways I am aware of adding a document (`<link rel="attachment" ...`, `<a rel="attachment"` and via the API `attachments=[Attachment('data:,api-bytes')],`

liZe:
We’re talking about a bug that was only found by a bot, that probably causes no problem in any PDF reader, that has been fixed, that’s been tested with a non-regression test. There’s no need to text extra cases that actually use the same code path, and to be honest there are more important tasks to spend time on. 😁

PR #2930, opened 2026-09-21
Stream deflate HTTP responses like gzip

FallenGodfather:
okay so now like `gzip` streamed via `StreamingGzipFile`, stream `deflate` responses using a streaming decompressor rather than buffering the entire decompressed payload into memory with `response.read()` and `zlib.decompress()`, I have also verified all 115 URL HTTP test cases passes, without any issues, and also code passed ruff checks also, would you like to add something ?

Supports both zlib format (RFC 1950) and raw deflate (RFC 1951) with fallback.


liZe:
I’ve been stupid to think that it would add a few lines like `StreamingGzipFile`.

And… I should have spent some time to do my homework. Zlib’s documentation asks to [use gzip and not deflate](https://zlib.net/zlib_faq.html#faq39) for HTTP. Apache [doesn’t support it](https://httpd.apache.org/docs/2.4/en/mod/mod_deflate.html), and NGINX [doesn’t even mention it](https://docs.nginx.com/nginx/admin-guide/web-server/compression/).

We’ve lost so much time on this.

Note for my future self: never, ever spend some time on a "security" problem or a "bug" found by a bot. They look interesting, but they’re terribly useless after you actually use a brain.

Let’s use #2932 instead.

